PR #1026, opened 2014-04-16
Append `Vary: Cookies` header to response when session is accessed

citruspi:
To prevent caching proxies like Varnish from providing multiple users with the same cache, you can set a `Vary: Cookie` header. 

This pull request appends the header to each response when the session is accessed.

I came across this issue while working on Warehouse, PyPi 2.0. @dstufft wanted something that added the header to the response anytime the session was being used.

&mdash; @citruspi + @abadger

### DIFF
--- a/flask/sessions.py
+++ b/flask/sessions.py
@@ -51,6 +51,13 @@ class SessionMixin(object):
     #: The default mixin implementation just hardcodes `True` in.
     modified = True
 
+    #: the accessed variable indicates whether or not the session object has
+    #: been accessed in that request. This allows flask to append a `Vary:
+    #: Cookie` header to the response if the session is being accessed. This
+    #: allows caching proxy servers, like Varnish, to use both the URL and the
+    #: session cookie as keys when caching pages, preventing multiple users
+    #: from being served the same cache.
+    accessed = True
 
 class TaggedJSONSerializer(object):
     """A customized JSON serializer that supports a few extra types that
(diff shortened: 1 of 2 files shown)


jeffwidman:
As of Nginx 1.7.7, the Nginx proxy cache (commonly used alongside uWSGI) also supports using the `Vary` header to manage what gets cached/returned--see the final FAQ question here: https://www.nginx.com/blog/nginx-caching-guide/

flask/sessions.py
```diff
@@ -334,24 +350,30 @@ def save_session(self, app, session, response):
...
+            response.headers.add('Vary', 'Cookie')
+
+        else:
+
+            # Delete case.  If there is no session we bail early.
+            # If the session was modified to be empty we remove the
+            # whole cookie.
+            if not session:
+                if session.modified:
+                    response.delete_cookie(app.session_cookie_name,
```
davidism: Seems like this should happen regardless of the header.

untitaker:
This appears to be the only way for Flask to do this automatically, but I'm not sure whose responsibility we want caching config to be.

davidism:
Django adds `Vary: Cookie` based on session access: https://github.com/django/django/blob/1ce04bcce0076360623ae164afd3541a5c031af2/django/contrib/sessions/middleware.py#L45. It also checks to delete the cookie before adding the header.

untitaker:
Then I suppose I am fine with it conceptually.

davidism:
Continued in #2288.

PR #1049, opened 2014-05-07
Set DEBUG and TESTING config value to default to None, not False.

smurfix:
Rationale: Some code might want to distinguish between "DEBUG has not been
set, determine correct value from the environment / other means" and "DEBUG
has been turned off explicitly".

One user of this is the Flask-Script extension, which would like to use a
default of True for the DEBUG setting when its built-in script runner is
used. Rationale: Many people use a special gunicorn-or-whatever runner in
production, but start a quick local "manage.py runscript" when they want to
test something / reproduce a bug. In this case it's helpful to not require
a different set of options that the user has to remember, or to have to use
a second config file which could go out of date.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -30,6 +30,10 @@ Version 1.0
   debug server through the click CLI system.  This is recommended over the old
   ``flask.run()`` method as it works faster and more reliable due to a
   different design and also replaces ``Flask-Script``.
+- Configuration settings for the ``DEBUG`` and ``TESTING`` config keys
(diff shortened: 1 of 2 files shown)


davidism:
I don't like the idea of automatically enabling debug mode without user action. Especially since there's a lot of users out there who don't understand that they shouldn't use `app.run()` / `manage.py run` in production. Allowing Flask-Script to figure this out would cause Flask-Script to shoot them in the foot instead of themselves doing it with an explicit `--debug` flag, which is less good.

PR #1056, opened 2014-05-12
Allow the jinja2 environment class to be overridden.

smurfix:
I need this change because I have to override a couple of Jinja2's templating functions.

Better than monkeypatching flask.app.Environment. ;-)


untitaker:
Apparently this has been duplicated by https://github.com/mitsuhiko/flask/pull/1422. I've merged that PR since it also had a testcase. I liked your naming style more though, so I've also renamed the class.

Thanks!

PR #1176, opened 2014-09-14
turn on autoescape for `flask.templating.render_template_string`

alanhamlett:
This seems like a sane default, since Flask turns on autoescape by default for html files when using the more common `render_template` so users will be expecting `render_template_string` to autoescape their template variables.


alanhamlett:
What should we do about this? If this pull request isn't going to be merged I'll remove it...

untitaker:
I think it should be included in 1.0, but mitsuhiko should decide.

It doesn't seem that he is reachable atm, but please reopen anyway.

alanhamlett:
I deleted the fork, so can't reopen. Created a new pull request #1515.

PR #1195, opened 2014-10-03
Link from Debug Mode to Application Errors

frabcus:
As a new user, I found that errors / stack traces were being swallowed. I searched for 'debug' in the documentation and found the Debug Mode section.

What I really wanted to know was how logging worked; it took me ages to find out because other frameworks (e.g. Rails) just calls this debugging.

This pull request adds a reference to the documentation that I think will help other people in this situation.


frabcus:
Thanks! This did confuse me for ages, so hope it helps others!

PR #1209, opened 2014-10-20
Handle a single unusual argument in jsonify

nickretallack:
I ran into this issue today, googled it, and found [someone else having the same issue](http://stackoverflow.com/questions/20668883/flask-jsonify-object-is-not-iterable).

If you're using the JSON Encoder in Overholt, you may notice that passing one of your models to jsonify does not work.  It says "object is not iterable", even though your custom json encoder will turn it into an iterable in a moment.  This is because Flask is trying to make it into a dict before that can happen.

This patch fixes this case.  Also it allows you to return other legitimate singular json values such as integers and strings.  Hopefully this doesn't raise any security issues.  If it does, someone else may wish to handle this more carefully.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -102,6 +102,9 @@ JSON Security
    with the builtin constructor of ``Array`` which closes this particular
    attack vector.
 
+   To disable this security check, set JSONIFY_ARRAY_SECURITY = False in
+   your config file.
+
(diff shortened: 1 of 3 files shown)


dghubble:
I'm sad to see this hasn't merged, though appreciate the responsibly cautious approach in #248

PR #1253, opened 2014-11-21
solution for running context dependent attributes

mastier:
It's not the best solution. It's just a concept of running out from situation when I need to get context dependant data for almost every view.
```
class MyView(BaseView):
    methods = ['GET']
    def dispatch_request(self):
        super(MyView, self).dispatch_request()
```
to get some attributes, like items in navigation menu
```
class BaseView(View):
    """
    BaseView provides parent class for most of the other views
    giving the content for navigation bar etc.
    """
  
    content = {}
    content['current_user'] = current_user
  
    def dispatch_request(self):
        """
        function dispatch_request is needed here, because db require
        app context is provided only in the dispatch_request scope
        """
        self.content['navigation_projects'] = sorted(get_project_mapping(db).keys())
```
`get_project_mapping is` is a reusable function
```
def get_project_mapping(db):
    projects_dict = {}
    projects = db.session.query(Project).filter(Project.name != None)
    for p in projects:
        projects_dict[p.name] = {
            'some_value: p.some_value,
        }
    return projects_dict
```

### DIFF
--- a/flask/views.py
+++ b/flask/views.py
@@ -81,6 +81,12 @@ class View(object):
         """
         def view(*args, **kwargs):
             self = view.view_class(*class_args, **class_kwargs)
+
+            # solution for running context dependent attributes
+            # i.e. db must have app initialized
+            if self.__bases__ != (object,):
+                for base in self.__bases__:
+                    base.dispatch_request(self)
             return self.dispatch_request(*args, **kwargs)
 
         if cls.decorators:


davidism:
I'm having trouble understanding the problem you're trying to solve here.  I have a feeling there's a better way to do what you want.  If you can explain with a more complete example, I'll look into this again.  Closing for now.

PR #1275, opened 2014-12-11
Include backend argument when instantiating Celery

AndrewBMartin:
Not including the backend argument can lead to AttributeError: DisabledBackend object has no attribute for '_get_task_meta_for'. See e.g., http://stackoverflow.com/questions/23215311/celery-with-rabbitmq-attributeerror-disabledbackend-object-has-no-attribute.

At the same time, including the backend argument doesn't seem to harm anything else.


untitaker:
Applied to master.

PR #1291, opened 2014-12-23
Fixed and intuitivized exception handling

flying-sheep:
abandoning pull request #1281 for this.

the rationale for this is to

1. fix some logic flaws
2. make errorhandlers more intuitive by making them rely on exception hierarchy instead of registration order
3. Enable 500 handlers on blueprint level

and all this while without breaking compatibility in not-already-broken usecases (which rely on the logic flaws i fixed)

TODO:

1. bake this into some tests as well
2. one error says

    > It is currently not possible to register a 500 internal server error on a per-blueprint level.

    why is this?

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1205,7 +1205,11 @@ class Flask(_PackageBoundObject):
         :type code_or_exception: int|T<=Exception
         :type f: callable
         """
-        assert not isinstance(code_or_exception, HTTPException)  # old broken behavior
+        if isinstance(code_or_exception, HTTPException):  # old broken behavior
+            raise ValueError(
+                'Tried to register a handler for an exception instance {0!r}. '
+                'Handlers can only be registered for exception classes or HTTP error codes.'
+                .format(code_or_exception))
         
         code = code_or_exception
         is_code = isinstance(code_or_exception, integer_types)
--- a/flask/app.py
+++ b/flask/app.py
@@ -8,14 +8,13 @@
     :copyright: (c) 2015 by Armin Ronacher.
     :license: BSD, see LICENSE for more details.
 """
-from collections import ChainMap
-
 import os
 import sys
 from threading import Lock
 from datetime import timedelta
 from itertools import chain
 from functools import update_wrapper
+from collections import Mapping
 
 from werkzeug.datastructures import ImmutableDict
 from werkzeug.routing import Map, Rule, RequestRedirect, BuildError
@@ -34,7 +33,7 @@ from .templating import DispatchingJinjaLoader, Environment, \
      _default_template_ctx_processor
 from .signals import request_started, request_finished, got_request_exception, \
      request_tearing_down, appcontext_tearing_down
-from ._compat import reraise, string_types, text_type, integer_types
+from ._compat import reraise, string_types, text_type, integer_types, iterkeys
 
 # a lock used for logger initialization
 _logger_lock = Lock()
@@ -76,7 +75,7 @@ def get_http_code(error_class_or_instance):
     return None
 
 
-class ExceptionHandlerDict(ChainMap):
+class ExceptionHandlerDict(Mapping):
     """A dict storing exception handlers or falling back to the default ones
     
     Designed to be app.error_handler_spec[blueprint_or_none]
@@ -86,12 +85,13 @@ class ExceptionHandlerDict(ChainMap):
     Returns None if no handler is defined for blueprint or app
     """
     def __init__(self, app, blueprint):
+        super(ExceptionHandlerDict, self).__init__()
         self.app = app
-        init = super(ExceptionHandlerDict, self).__init__
+        self.data = {}
         if blueprint:  # fall back to app mapping
-            init({}, app.error_handler_spec[None])
+            self.fallback = app.error_handler_spec[None]
         else:
-            init({})
+            self.fallback = {}
     
     def get_class(self, exc_class_or_code):
         if isinstance(exc_class_or_code, integer_types):
@@ -103,14 +103,28 @@ class ExceptionHandlerDict(ChainMap):
         return exc_class
     
     def __contains__(self, e_or_c):
-        return super(ExceptionHandlerDict, self).__contains__(self.get_class(e_or_c))
+        clazz = self.get_class(e_or_c)
+        return clazz in self.data or clazz in self.fallback
     
     def __getitem__(self, e_or_c):
-        return super(ExceptionHandlerDict, self).__getitem__(self.get_class(e_or_c))
+        clazz = self.get_class(e_or_c)
+        item = self.data.get(clazz)
+        if item is not None:
+            return item
+        elif len(self.fallback):
+            return self.fallback[clazz]
+        else:
+            raise KeyError(e_or_c)
     
     def __setitem__(self, e_or_c, handler):
         assert callable(handler)
-        return super(ExceptionHandlerDict, self).__setitem__(self.get_class(e_or_c), handler)
+        self.data[self.get_class(e_or_c)] = handler
+    
+    def __iter__(self):
+        return iterkeys(self.data)
+    
+    def __len__(self):
+        return len(self.data)
     
     def find_handler(self, ex_instance):
         assert isinstance(ex_instance, Exception)
--- a/flask/app.py
+++ b/flask/app.py
@@ -65,16 +65,6 @@ def setupmethod(f):
     return update_wrapper(wrapper_func, f)
 
 
-def get_http_code(error_class_or_instance):
-    if (
-        isinstance(error_class_or_instance, HTTPException) or
-        isinstance(error_class_or_instance, type) and
-        issubclass(error_class_or_instance, HTTPException)
-    ):
-        return error_class_or_instance.code
-    return None
-
-
 class ExceptionHandlerDict(Mapping):
     """A dict storing exception handlers or falling back to the default ones
     
@@ -93,7 +83,8 @@ class ExceptionHandlerDict(Mapping):
         else:
             self.fallback = {}
     
-    def get_class(self, exc_class_or_code):
+    @staticmethod
+    def get_class(exc_class_or_code):
         if isinstance(exc_class_or_code, integer_types):
             # ensure that we register only exceptions as keys
             exc_class = default_exceptions[exc_class_or_code]
--- a/flask/app.py
+++ b/flask/app.py
@@ -1079,14 +1079,12 @@ class Flask(_PackageBoundObject):
         return decorator
 
     @staticmethod
-    def _ensure_exc_class(exc_class_or_code):
+    def _get_exc_class_and_code(exc_class_or_code):
         """ensure that we register only exceptions as handler keys"""
         if isinstance(exc_class_or_code, integer_types):
             exc_class = default_exceptions[exc_class_or_code]
-        elif isinstance(exc_class_or_code, type):
-            exc_class = exc_class_or_code
         else:
-            exc_class = type(exc_class_or_code)
+            exc_class = exc_class_or_code
         
         assert issubclass(exc_class, Exception)
         
@@ -1164,7 +1162,7 @@ class Flask(_PackageBoundObject):
                 'Handlers can only be registered for exception classes or HTTP error codes.'
                 .format(code_or_exception))
         
-        exc_class, code = self._ensure_exc_class(code_or_exception)
+        exc_class, code = self._get_exc_class_and_code(code_or_exception)
         
         handlers = self.error_handler_spec.setdefault(key, {}).setdefault(code, {})
         handlers[exc_class] = f
@@ -1412,7 +1410,7 @@ class Flask(_PackageBoundObject):
         """Finds a registered error handler for the request’s blueprint.
         If neither blueprint nor App has a suitable handler registered, returns None
         """
-        exc_class, code = self._ensure_exc_class(e)
+        exc_class, code = self._get_exc_class_and_code(type(e))
         
         def find_superclass(d):
             if not d:
@@ -1422,6 +1420,7 @@ class Flask(_PackageBoundObject):
                     return None
                 handler = d.get(superclass)
                 if handler is not None:
+                    d[exc_class] = handler  # cache for next time exc_class is raised
                     return handler
             return None
         
--- a/flask/app.py
+++ b/flask/app.py
@@ -1080,7 +1080,7 @@ class Flask(_PackageBoundObject):
 
     @staticmethod
     def _get_exc_class_and_code(exc_class_or_code):
-        """ensure that we register only exceptions as handler keys"""
+        """Ensure that we register only exceptions as handler keys"""
         if isinstance(exc_class_or_code, integer_types):
             exc_class = default_exceptions[exc_class_or_code]
         else:
@@ -1412,15 +1412,15 @@ class Flask(_PackageBoundObject):
         """
         exc_class, code = self._get_exc_class_and_code(type(e))
         
-        def find_superclass(d):
-            if not d:
+        def find_superclass(handler_map):
+            if not handler_map:
                 return None
-            for superclass in exc_class.mro():
+            for superclass in exc_class.__mro__:
                 if superclass is BaseException:
                     return None
-                handler = d.get(superclass)
+                handler = handler_map.get(superclass)
                 if handler is not None:
-                    d[exc_class] = handler  # cache for next time exc_class is raised
+                    handler_map[exc_class] = handler  # cache for next time exc_class is raised
                     return handler
             return None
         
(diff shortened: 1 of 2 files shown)


flask/app.py
```diff
@@ -1133,16 +1150,21 @@ def register_error_handler(self, code_or_exception, f):
...
-            self.error_handler_spec.setdefault(key, {})[code_or_exception] = f
-        else:
-            self.error_handler_spec.setdefault(key, {}).setdefault(None, []) \
-                .append((code_or_exception, f))
+        """
+        :type key: None|str
+        :type code_or_exception: int|T<=Exception
+        :type f: callable
+        """
+        if isinstance(code_or_exception, HTTPException):  # old broken behavior
```
untitaker: You catch this broken behavior here, but try to correct it in _ensure_exc_class.

flask/app.py
```diff
@@ -1370,6 +1392,32 @@ def url_defaults(self, f):
...
+        
+        def find_superclass(d):
+            if not d:
+                return None
+            for superclass in exc_class.mro():
+                if superclass is BaseException:
+                    return None
+                handler = d.get(superclass)
+                if handler is not None:
+                    return handler
```
untitaker: Could we store this handler at ``d[e]`` to avoid traversing the MRO every time?

untitaker:
@flying-sheep This looks really good now BTW.

flask/app.py
flying-sheep: i don’t correct it. `_ensure_exc_class` is called here on a exception class or error code and in `_find_error_handler` on an exception instance.

i could have written two different methods (or inlined the code), but decided not to.

untitaker: >and in _find_error_handler on an exception instance.

Ah I see. But why not do ``_ensure_exc_class(type(e))`` in ``_find_error_handler``? This saves one typecheck and makes for a cleaner API.

flask/app.py
```diff
@@ -1133,16 +1150,21 @@ def register_error_handler(self, code_or_exception, f):
...
+        """
+        if isinstance(code_or_exception, HTTPException):  # old broken behavior
+            raise ValueError(
+                'Tried to register a handler for an exception instance {0!r}. '
+                'Handlers can only be registered for exception classes or HTTP error codes.'
+                .format(code_or_exception))
+        
+        exc_class, code = self._ensure_exc_class(code_or_exception)
+        
+        handlers = self.error_handler_spec.setdefault(key, {}).setdefault(code, {})
```
untitaker: Why special-case HTTP exceptions and have a separate dict for each status code?

Since ``self.error_handler_spec`` is already backwards-incompatible (as in, code that directly modifies it breaks), I think we can remove this distinction. I suppose that also eliminates the need for most of the code in ``_ensure_exc_class``.

flask/app.py
flying-sheep: you mean `d[exc_class]`. `e` is an exception instance.

and that would create an edge case:

```python
app = Flask(__name__)

# hierarchy: Exception → OSError → ConnectionError → ConnectionAbortedError

@app.errorhandler(OSError)
def handle_os_error(e):
    ...

app.run()

somewhere during runtime some connection aborts, a ConnectionAbortedError instance is raised. now, due to your caching: `d[ConnectionAbortedError] is d[OSError]`

also during runtime (after that), some new error handler is registered. (not inconceivable):

```python
@app.errorhandler(ConnectionError)
def handle_connection_error(e):
    ...
```

well fuck, now another ConnectionAbortedError instance being raised won’t trigger the newly registered error handler but the cached one!

---

we can circumvent this by instead having a separate dict holding all cached handlers, and clearing all superclasses of a newly registered exception from that cache handler dict. like so:

```python
# in __init__:
self._error_handler_cache = {}

# in _register_error_handler:
for superclass in exc_class.mro():
    try:
        del self._error_handler_cache[superclass]
    except KeyError:
        pass

# in _find_error_handler → find_superclass (after “if not d: return None”):
handler = self._error_handler_cache.get(exc_class)
if handler is not None:
    return handler

# in _find_error_handler → find_superclass (before returning the handler):
self._error_handler_cache[exc_class] = handler
```

untitaker: >You mean exc_class

Ah, yes, sorry.

>also during runtime (after that), some new error handler is registered. (not inconceivable)

This is explicitly forbidden by Flask. See the ``setupmethod`` decorator.

flask/app.py
flying-sheep: i thought about that as well and thought that it’s clearer this way. all handlers are neatly grouped and accessing them in debug mode using `self.error_handler_spec[None][403]` will list you everything.

but i’m not particularly attached to this.

PS: we can preserve partial backwards compat by using a dict subclass for the `error_handler_spec[foobar]` level by redefining `__setitem__`:

```python
class BackwardsCompatDict(dict):
    def __init__(self, app):
        super().__init__()
        self.app = app
    def __setitem__(self, key, value):
        if callable(value) and not isinstance(value, dict):
            exc_class, code = self.app._ensure_exc_class(key)
            self.setdefault(code, {})[exc_class] = value
        else:
            super().__setitem__(key, handler_or_handlerdict)

app.error_handler_spec[None][403] = forbidden_handler
assert app.error_handler_spec[None][403][Forbidden] is forbidden_handler
```

flying-sheep:
> @flying-sheep This looks really good now BTW.

thanks! :D

flask/app.py
flying-sheep: makes sense!

flask/app.py
untitaker: >we can preserve partial backwards compat

I think the partial backwards-compatibility is not worth the "magic", especially as it's partial.

untitaker:
@flying-sheep Don't invest too much time in this yet, @mitsuhiko must decide this ultimatively.

flask/app.py
flying-sheep: seems kinda arbitrary, but makes things easier for us here.

also i don’t see any problems with your caching idea. i’ll just add `d[exc_class] = handler` and we’re done, since the loop will only do one iteration when hitting the cache. (`exc_class.mro()[0] is exc_class`, so `d.get(superclass)` will always return something from then on)

untitaker: >seems kinda arbitrary, but makes things easier for us here.

I suppose otherwise errorhandler and route registration has to be threadsafe.

>i’ll just add d[exc_class] = handler

Yup, sounds good.

flying-sheep:
well, this is already discussed and implemented, everything that comes now is mere minutes of work, so i might as well.

otherwise i trust mitsuhiko’s judgment except in the matter of Python 2 vs Python 3 ;)

flying-sheep:
and regarding judgment: do you think the cleanness and neatness of grouping exception handlers using their http codes (or lack thereof) is worth the fact that it’s unnecessary for the logic? or not?

untitaker:
I don't think it's worth it, but @mitsuhiko may come up with a reason why it is absolutely necessary (I've encountered such surprises before).

flying-sheep:
can we now start the magic dance that summons him?

untitaker:
I still don't know how that dance works. But I usually spam him on IRC regularly.

untitaker:
@flying-sheep Would it be possible to fix https://github.com/mitsuhiko/flask/issues/503 in the run?

flask/app.py
```diff
@@ -1370,6 +1390,33 @@ def url_defaults(self, f):
...
+    def _find_error_handler(self, e):
+        """Finds a registered error handler for the request’s blueprint.
+        If neither blueprint nor App has a suitable handler registered, returns None
+        """
+        exc_class, code = self._get_exc_class_and_code(type(e))
+        
+        def find_superclass(d):
+            if not d:
+                return None
+            for superclass in exc_class.mro():
```
untitaker: From [SO](https://stackoverflow.com/questions/19713585/whats-the-difference-between-the-mro-method-and-the-mro-attribute-of-a-clas):

>mro() is called on instanciation and stores its result in __mro__. They don't really have the same purpose.

I don't even understand why this method is made available in Python, but using ``__mro__`` seems more fitting. (brought up by @mitsuhiko in IRC)

untitaker:
Requesting help on reviewing this PR. I think it's a great change that definetly should land in 1.0.

flask/app.py
```diff
@@ -1075,6 +1075,21 @@ def decorator(f):
             return f
         return decorator
 
+    @staticmethod
+    def _get_exc_class_and_code(exc_class_or_code):
+        """ensure that we register only exceptions as handler keys"""
```
adambyrtek: Start comment from a capital letter.

flask/app.py
```diff
@@ -1370,6 +1390,33 @@ def url_defaults(self, f):
         self.url_default_functions.setdefault(None, []).append(f)
         return f
 
+    def _find_error_handler(self, e):
+        """Finds a registered error handler for the request’s blueprint.
+        If neither blueprint nor App has a suitable handler registered, returns None
+        """
+        exc_class, code = self._get_exc_class_and_code(type(e))
+        
+        def find_superclass(d):
```
adambyrtek: Would be nice to have a more meaningful argument name, like `handlers` or `handler_dict`?

flask/app.py
```diff
@@ -1370,6 +1390,33 @@ def url_defaults(self, f):
...
+            if not d:
+                return None
+            for superclass in exc_class.mro():
+                if superclass is BaseException:
+                    return None
+                handler = d.get(superclass)
+                if handler is not None:
+                    d[exc_class] = handler  # cache for next time exc_class is raised
```
adambyrtek: Is the caching really worth it? In the context of request handling and exception processing a few iterations down the MRO seem insignificant.

flask/app.py
flying-sheep: hmm, a bit out of the loop after all this time.

i see a `assert issubclass(exc_class, Exception)` in `_get_exc_class_and_code`, which would already catch this.

should i simply replace that assertion with the more meaningful `ValueError`?

flask/app.py
flying-sheep: done

flask/app.py
flying-sheep: [mro() can be overridden by a metaclass that’s why it’s there](https://docs.python.org/3/library/stdtypes.html#class.mro)

so i’ll keep it. and you and mitsuhiko should switch to it.

flask/app.py
flying-sheep: well, it’s just two lines and doesn’t do any harm, since the registered handlers may not be changed after app start (so the cache will never be invalidated)

and the few bytes more memory are even less significant than the cycles saved ;)

flying-sheep:
btw. the regression test fails on master currently so i din’t break anything, i just rebased :)

flask/app.py
untitaker: I think it's fine as-is. You did address this suggestion, it seems:

>But why not do _ensure_exc_class(type(e)) in _find_error_handler? This saves one typecheck and makes for a cleaner API.

flask/app.py
flying-sheep: done

untitaker:
>btw. the regression test fails on master currently so i din’t break anything, i just rebased :)

It doesn't fail for me (I reverted the offending commit a while ago)

flying-sheep:
oh, strange. it doesn’t fail on travis. wat.

locally, `test_memory_consumption` fails.

flask/app.py
flying-sheep: ok. github really isn’t perfect for sub-issues like this. i have to write “done” everywhere i actually did something, or trust on the automated diff collapsing (outdated change)

untitaker:
Hmm, how does https://github.com/mitsuhiko/flask/pull/1407 behave for you then?

flask/app.py
untitaker: >trust on the automated diff collapsing (outdated change)

That usually works pretty well.

flying-sheep:
so checkout unitaker/flask#no-threads and try? will do

flask/app.py
```diff
@@ -1078,6 +1078,21 @@ def decorator(f):
...
+    def _get_exc_class_and_code(exc_class_or_code):
+        """Ensure that we register only exceptions as handler keys"""
+        if isinstance(exc_class_or_code, integer_types):
+            exc_class = default_exceptions[exc_class_or_code]
+        else:
+            exc_class = exc_class_or_code
+        
+        assert issubclass(exc_class, Exception)
+        
+        if issubclass(exc_class, HTTPException):
```
untitaker: I think we missed this discussion: https://github.com/mitsuhiko/flask/pull/1291#discussion_r22427026

>

Why special-case HTTP exceptions and have a separate dict for each status code?

Since self.error_handler_spec is already backwards-incompatible (as in, code that directly modifies it breaks), I think we can remove this distinction. I suppose that also eliminates the need for most of the code in _ensure_exc_class.

untitaker: Hmm, maybe we should ignore that for now and maybe do a second refactoring. The current one is big enough as-is.

flying-sheep: why not. it isn’t as if it would do any harm to have them grouped like this

untitaker:
OK, so the current PR is excellent as-is.

The only thing missing now are docs. From a quick glance at `errorhandling.rst` there doesn't seem to be anything contradictory in there, but we should definetly clarify the current behavior.

We also need something in `upgrading.rst`, and a changelog entry.

flying-sheep:
writing something up right now, but `errorhandling.rst` doesn’t seem to contain anything about error handlers, only logging and debugging AFAICS

untitaker:
We need to document how Flask prioritizes errorhandlers, doesn't matter where. `errorhandling` seemed like the most related page to me.

untitaker:
Thanks @flying-sheep!

flying-sheep:
been a pleasure :smile:

flask/app.py
adambyrtek: This might be confusing during debugging, since the handler dict mutates during execution, but you might argue that when debugging you should be aware of the cache, so I don't feel strongly about this, I just don't like premature optimization.

untitaker: Good point about debugging, with that in mind I'd like to remove that cache as well. @adambyrtek care to setup a PR, since this is already merged?

flying-sheep: i’m curious: what’s the problem during debugging?

untitaker: They might not invalidate themselves correctly when the programmer is testing a very recent change they made, or otherwise warp results. As @adambyrtek said, you could argue that the programmer should simply be *aware* of those caches (as is the case with e.g. browser caches -- I always use icognito mode to test e.g. redirecting behavior), but in our case, it's questionable whether this cache even gives us any performance benefits (which is clearly the case with browser caches).

Also: In my opinion, caches are a tempting (and usually bad) answer to CPU-bound performance problems: They are simply there for delaying the writing of a more performant algorithm.

flying-sheep: hmm, as i said: it’s forbidden to register new routes after app start, so the cache will never be invalidated.

that’s the main reason i implemented it: very simple and no potential for failure

untitaker: >and the more efficient algorithm may be to statically include all exceptions in the mapping upon app start. so basically replacing my JIT cache with an AOT cache :smile:

That's impossible, we can't reach subclasses from parent classes.

Honestly I think instead of proving that additional code doesn't cause problems, it should be proven that the absence of the code doesn't cause problems. IOW, the right for the cache to exist doesn't lie in "it doesn't cause any problems", but rather "it improves performance". And that has to be proven instead of saying "why not".

flying-sheep: well, that’s obvious, but it’s not clear if it does *notably so*

untitaker: Just took a more exact look at the cache impl, it seems half as expensive as I
thought. I somehow assumed we used a second dict for it.

untitaker: Just looked through the old discussions, and it was me who suggested this cache
in the first place >.<

flying-sheep: haha, i also didn’t remember. but yeah :laughing:

adambyrtek: Looks like you have a consensus, so I'm not trying to reopen this, just explaining my arguments:

- When Flask user encounters issues with error handling he is likely to check the contents of `error_handler_spec` during debugging, and be surprised that there are entries there which haven't been explicitly registered.
- On first sight `error_handler_spec` seems to be modified only when a new handler is registered. It's not obvious that we pass one of the inner dicts by reference and modify it inside `_find_error_handler`, especially since the name doesn't suggest that the function has side effects.

Anyway, I think optimizations are worth introducing when there is a real performance problem to solve, and I don't see any problem here... Just my 2 cents.

untitaker: The part about the user expecting no further errorhandlers in that dict is a
pretty good argument... I'm still not really sure whether we should keep that
in or not. 

Please cont. at https://github.com/mitsuhiko/flask/issues/1433, this discussion
is kinda hidden.

PR #1295, opened 2014-12-27
register user errorhandler in the right order

moonshadow:
solved the problem proposed by  #1291 in a lighter way,
thus determining the `exception  hierarchy` at app setup time instead of running time.


untitaker:
#1291 has been merged.

PR #1301, opened 2014-12-31
Allow to know about Blueprints registeration state

noirbizarre:
This pull-request permit to know if a blueprint has already been registered or not (very usefull for extensions).

It also remove the never used (and misleading) ``Blueprint._got_registered_once`` attribute.

Now, you can write:
```python
blueprint = Blueprint('blueprint', __name__)

assert not blueprint.is_registered

app = Flask(__name__)
app.register_blueprint(blueprint)
assert blueprint.is_registered
```

It also allows to specify than a blueprint can't be registered twice with the ``register_once`` constructor kwarg:
```python
blueprint = Blueprint('blueprint', __name__, register_once=True)
app = Flask(__name__)
app.register_blueprint(blueprint)
app.register_blueprint(blueprint) # Raise a warning
```

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -118,6 +118,20 @@ On top of that you can register blueprints multiple times though not every
 blueprint might respond properly to that.  In fact it depends on how the
 blueprint is implemented if it can be mounted more than once.
 
+You can hint other developers on your implementation with the ``register_once``
+parameter::
+
+  from flask import Flask, Blueprint
+
+  blueprint = Blueprint('blueprint', __name__, register_once=True)
+  app = Flask(__name__)
+
+  assert not blueprint.is_registered
+  app.register_blueprint(blueprint)
+  assert blueprint.is_registered
+
+  app.register_blueprint(blueprint) # Raise a warning
+
 Blueprint Resources
 -------------------
 
--- a/flask/blueprints.py
+++ b/flask/blueprints.py
@@ -87,12 +87,11 @@ class Blueprint(_PackageBoundObject):
     """
 
     warn_on_modifications = False
-    _got_registered_once = False
 
     def __init__(self, name, import_name, static_folder=None,
                  static_url_path=None, template_folder=None,
                  url_prefix=None, subdomain=None, url_defaults=None,
-                 root_path=None):
+                 root_path=None, register_once=False):
         _PackageBoundObject.__init__(self, import_name, template_folder,
                                      root_path=root_path)
         self.name = name
@@ -102,6 +101,8 @@ class Blueprint(_PackageBoundObject):
         self.static_url_path = static_url_path
         self.deferred_functions = []
         self.view_functions = {}
+        self.register_once = register_once
+        self.is_registered = False
         if url_defaults is None:
             url_defaults = {}
         self.url_values_defaults = url_defaults
@@ -112,7 +113,7 @@ class Blueprint(_PackageBoundObject):
         state as argument as returned by the :meth:`make_setup_state`
         method.
         """
-        if self._got_registered_once and self.warn_on_modifications:
+        if self.is_registered and self.warn_on_modifications:
             from warnings import warn
             warn(Warning('The blueprint was already registered once '
                          'but is getting modified now.  These changes '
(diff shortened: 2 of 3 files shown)


untitaker:
In which specific situation do you need this information?

noirbizarre:
In my case, I want it for two reasons:

Being able to raise an explicit and detailled error message on multiple blueprint registerations.

Being able to provide an optionnal helper function with assets provided by a blueprint (and needed to be registered only in this case):
```python
blueprint = Blueprint('helper', __name__,
    template_folder='templates',
    static_folder='static'
)
def my_helper(obj):
    '''Render a SwaggerUI for a given API'''
    if not blueprint.is_registered:
        current_app.register_blueprint(blueprint)
    return render_template('something.html', obj=obj)
```

Allowing to write somewhere else:
```python
from mylib import my_helper
bp = Blueprint('test', __name__)

@bp.route('/somewhere')
def my_route():
    obj = build()
    do_something(obj)
    return my_helper(obj)
```

Without, I need to register an extra blueprint.
I don't want the developper to register himself the blueprint as it does not expose any URL, only assets.

untitaker:
Afaik it is a explicitly prohibited (via code in Flask itself) to lazily
register blueprints this way (during execution of the endpoint function), as
the routing system needs to be completely set up before the first request comes
in.

It also seems like a bad idea in any way because the blueprint could override
some other templates, altering the behavior of the whole application when the
user issues a request to a GET endpoint (which are supposed to be idempotent).

I recommend you do the following (as per :

    class MyHelper(object):
        def __init__(self, app=None):
            self.app = app
            if app is not None:
                self.init_app(app)

        def init_app(self, app):
            app.register_blueprint(blueprint)

        @staticmethod
        def get_swagger_ui(obj):
            return render_template('something.html', obj=obj)

You can also detach ``get_swagger_ui`` from MyHelper and make it a global
function. In any case, the usage would be:

    helper = MyHelper(app)

    @app.route('/somewhere')
    def my_route():
        ...
        helper.get_swagger_ui(obj)

untitaker:
Huh, it seems GitHub doesn't correctly format email messages anymore.

noirbizarre:
I already have class like ``MyHelper``.
My problem is that this is class that can be registered many times, not like the blueprint.
So, the problem is the same, I need to be able to write:
```python
class MyHelper(object):
    def __init__(self, app=None):
        self.app = app
        if app is not None:
            self.init_app(app)

    def init_app(self, app):
        if not blueprint.is_registered:
            app.register_blueprint(blueprint)

    @staticmethod
    def get_swagger_ui(obj):
        return render_template('something.html', obj=obj)
```
and then:
```python
helper1 = MyHelper(app)
helper2 = MyHelper(app)

@app.route('/somewhere')
def my_route():
    ...
    helper1.get_swagger_ui(obj)
```

By the way, to avoid collision, my blueprint is declared with an url_prefix:
```python
blueprint = Blueprint('helper', __name__,
    template_folder='templates',
    static_folder='static',
    url_prefix='/libprefix'
)
```

If you look at the code, there was an attribute meant to implement that: ``_got_registered_once``.
The problem: it wasn't working and induces me in error.
This is why I started this pull-request.

untitaker:
Good point. Your approach breaks down when using the same extension object on different apps though (which is a valid usage of most extension objects). The solution would be to store this kind of information on the app object (and not on the blueprint):

    ext_dict = app.extensions.setdefault('myhelper', {})
    if not ext_dict.get('registered_bp', False):
        app.register_blueprint(blueprint)
    ext_dict['registered_bp'] = True

noirbizarre:
I like your approach. I will give it a try.

But, I still think removing ``_got_registered_once`` is good idea if it's not working, not used, and not documented.

untitaker:
I agree.

untitaker:
Actually ``_got_registered_once`` is used inside ``Blueprint.record`` and set inside ``Blueprint.register``. Seems to work fine, but the value should apparently not be used by users.

noirbizarre:
Yes, I know, but it's actually not working ().
I provided a test case with ``is_registered``, but for the original ``_got_registered_once`` you can test with:
```python
bp1 = Blueprint('bp1', __name__)
bp2 = Blueprint('bp2', __name__)

assert not bp1._got_registered_once
assert not bp2._got_registered_once

app = Flask(__name__)
app.register_blueprint(bp1)

assert bp1._got_registered_once
assert not bp2._got_registered_once  # Fails
```

untitaker:
That runs fine for me. Which version of Flask is this?

PR #1308, opened 2015-01-04
Replace hard coded path

aliasm2k:
Use `os.path.dirname` and `os.path.abspath` to dynamically build upload path


untitaker:
Actually this doesn't work, url_for needs an appcontext.

I am not sure if this is out of bounds for a simple example.

aliasm2k:
Infact it does work. :smiley: 
Anyway, if you think it is complicated for simple example, its fine :+1:

ThiefMaster:
Just FYI, you are supposed to have two spaces before and one space after the `#`, e.g. `foo()  # do foo stuff`

PR #1311, opened 2015-01-10
Conditionally install default logging handlers

dtheodor:
This is an attempt to address https://github.com/mitsuhiko/flask/issues/641, in particular the problem that `del logger.handlers[:]` is always invoked in app logger creation, destroying any previously created handlers.

Since the `LOGGER_HANDLER_POLICY` was introduced, I am using it to only conditionally install each of the two default handlers (`'debug'` and `'production'`). If the policy is `'never'`, the handlers do not need to be added and such cleanup `del logger.handlers[:]` is not necessary.

The test I added describes my use case (and I believe everyone's in https://github.com/mitsuhiko/flask/issues/641) quite accurately, and I think it is a reasonable one. 

My solution to not install handlers is different than the previous situation, which would always install them and then would do a check on each invocation of `handler.emit`. That means that if someone relied on this check and was for some reason dynamically modifying the app's configuration to change the logging policy while the app was running, it won't work any more. But, to me this seems more like an implementation artifact than a valid use-case.

### DIFF
--- a/flask/logging.py
+++ b/flask/logging.py
@@ -54,37 +54,41 @@ def create_logger(app):
     function also removes all attached handlers in case there was a
     logger with the log name before.
     """
-    Logger = getLoggerClass()
-
-    class DebugLogger(Logger):
-        def getEffectiveLevel(x):
-            if x.level == 0 and app.debug:
-                return DEBUG
-            return Logger.getEffectiveLevel(x)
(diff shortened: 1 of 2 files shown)


jakul:
@mitsuhiko What do we need to do to move this PR forward?

PR #1313, opened 2015-01-13
Added missing import.

homeworkprod:
As this specific code block lists imports in general and an import from the `flask` package specifically, it should also include `g` for the sake of completeness.


untitaker:
Thanks!

PR #1315, opened 2015-01-13
Replace TextField with StringField in registration page example

brettgerry:
Replaces TextField with StringField for the username and email. Per http://wtforms.readthedocs.org/en/latest/whats_new.html#deprecated-api-s TextField has been deprecated in WTForms.


untitaker:
Thanks!

PR #1325, opened 2015-01-21
Clarify contents of request properties.

bsutherland:
Change URL to IRI and update examples to include extended characters. See also this question on stackoverflow.com: http://stackoverflow.com/questions/28056959/in-python-flask-how-to-access-complete-raw-url-prior-to-un-escaping/28058628#28058628

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -92,24 +92,24 @@ Incoming Request Data
    .. attribute:: base_url
    .. attribute:: url_root
 
-      Provides different ways to look at the current URL.  Imagine your
-      application is listening on the following URL::
+      Provides different ways to look at the current `IRI<http://tools.ietf.org/html/rfc3987>`_.  Imagine your
+      application is listening on the following IRI::
 
           http://www.example.com/myapplication
 
-      And a user requests the following URL::
+      And a user requests the following IRI::
 
-          http://www.example.com/myapplication/page.html?x=y
+          http://www.example.com/myapplication/日本/page.html?x=y
 
       In this case the values of the above mentioned attributes would be
       the following:
 
       ============= ======================================================
-      `path`        ``/page.html``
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Wouldn't it suffice to use e.g. simple German Umlauts as a non ASCII example? When reading your commit I first thought it's well-disguised vandalism ;)

bsutherland:
ThiefMaster - I agree with your point. At first I though a special European character might be more universally recognised, but I went with Japanese for two reasons:

1) I though a couple of small dots or an accent could be easily missed
2) I don't read or write any European languages other than English, but do know some Japanese

Maybe the German eszet would be a better choice then?

untitaker:
I think it'd be nice if the intermediate URI is also shown in the docs... although I am not sure if that is within scope.

bsutherland:
untitaker, do you mean show the URI for each of the examples in the table? I guess that might clarify things, and maybe referencing iri_to_uri would be a good idea too..

untitaker:
I don't think that's necessary, I mean only the encoded URI that is sent by the client. Before the table.

Not sure linking to ``iri_to_uri`` is a good idea, people might try to use it in situations where it isn't required.

docs/api.rst
```diff
@@ -92,25 +92,25 @@ Incoming Request Data
    .. attribute:: base_url
    .. attribute:: url_root
 
-      Provides different ways to look at the current URL.  Imagine your
-      application is listening on the following URL::
+      Provides different ways to look at the current `IRI<http://tools.ietf.org/html/rfc3987>`_.  Imagine your
+      application is listening on the following application root::
```
untitaker: Could you word-wrap this at ~70 chars?

untitaker:
Excellent, thanks!

PR #1331, opened 2015-01-28
Fix #1330 Support capital letters in mimetype in is_json

### LINKED ISSUE #1330: Method is_json does not support capital letters in mimetype
Method is_json (flask/wrappers.py:112) tests whether mimetype is in format 'application/...+json'. According to http://tools.ietf.org/html/rfc2045, mime types are case insensitive.

'The type, subtype, and parameter names are not case sensitive.  For example, TEXT, Text, and TeXt are all equivalent top-level media types.'
 
So mimetype 'ApLiCaTiOn/jSoN' is technically fine (and yes, extremely rare) and as the method tests the mimetype just for lowercase mimetype 'application/json' the previous example will fail, yet it should be ok.

KarelJakubec:


untitaker:
Your bugreport is fine, but I think this should be fixed on a broader scale, i.e. make response.mimetype always lowercase.

PR #1335, opened 2015-02-01
Update api.rst, fixed spelling

phouse512:


untitaker:
Ah, thanks!

untitaker:
I see now that this was against the maintenance branch: You'd think this would make it into the online docs, but it won't... the docs build from tags.

phouse512:
Yeaa I noticed that.  Just curious, I'm still learning, what does it mean
that the docs build from tags?

untitaker:
Master and the maintenance branches are modifyable... they point to a particular commit, their HEAD, but branches are "designed to be updated". Tags are really similar to branches, except that they're immutable (unless you force git to behave otherwise), so you can't just submit a PR against a tag.

We use branches to track minor/major versions (0.10.X, 1.0, etc, see semver), where master is the latest one, and do releases from each of them. Each release has its own tag.

phouse512:
Ahh, I see.  So if I understand correctly, although the branch we just
pushed to is the .10 maintenance one, since the docs build from the the .10
maintenance tag, this won't be visible online?

PR #1337, opened 2015-02-05
Fixed `stream_with_context` decorator not working with generator functions that take in arguments

univerio:
The `stream_with_context` decorator currently just throws away `*args` and `**kwargs`; it would be helpful to pass them along to the generator.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -100,7 +100,7 @@ def stream_with_context(generator_or_function):
         gen = iter(generator_or_function)
     except TypeError:
         def decorator(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


jeffwidman:
I'm afraid I'm just being dumb here and don't understand the impact of this change. 

Can you give an example or two of how passing along arguments to the generator is helpful?

Any potential drawbacks to adding this?

ThiefMaster:
Currently you can only decorate functions that take no arguments with `stream_with_context`

davidism:
I already merged #1747 which addressed the same thing.  Didn't see this older one.

Issue #348, opened 2011-11-08
Static files only work for blueprints registered with url_prefix


deshipu:
The option `static_folder` for blueprints only works correctly when the blueprint has been registered with a `url_prefix`, otherwise the application's default static URL takes precedence.

There is nothing wrong in that behavior, but it is not documented. It would be nice to have that fact mentioned at http://flask.pocoo.org/docs/blueprints/#static-files and/or http://flask.pocoo.org/docs/api/#blueprint-objects

oxbambooxo:
<https://vilimpoc.org/blog/2012/11/21/serving-static-files-from-root-and-not-static-using-flask/>
this essay look like is can work fix that situation can be resolved

juanitogan:
Something very strange is up when using `static_url_path`, and I'm not using blueprints.  When I set:

`app = Flask(__name__, static_url_path="", static_folder="../public")`

then `@app.route("/<path:path>")` doesn't match on anything even though all my other routes work.  When I set:

`app = Flask(__name__, static_folder="../public")`

then `@app.route("/<path:path>")` matches on everything including intercepting all links to resources in index.html, which is annoying.  It seems my only hope is to rebuild :tired_face: to default Flask locations (`<app>/static`) and see if that magically works.  I dislike magic.  Flask 0.10.1.

danielmhair:
Hey guys, thank you @juanitogan, we have been working on this for days and your post really solved it, thanks!

jbejar:
I manually created the rules instead of registering the blueprint
```
    react = Blueprint('react', __name__, static_folder='../react/dist/')
    app.add_url_rule('/', endpoint='root',
                     view_func=lambda: react.send_static_file("index.html"))
    app.add_url_rule('/<path:filename>',
                     endpoint='react.static',
                     view_func=react.send_static_file)
```

DusanMadar:
It works for me when:

1. set `static_folder='static'` when creating the Blueprint
2. use `'.static'` (note the `.`) in `url_for`, e.g.  I have `{{ url_for('.static', filename='js/a-script.js') }}` in `base.html`

And here's my app structure:
```
flask_app
├── flask_app
│   ├── app.py
│   ├── blueprints
│   │   ├── __init__.py
│   │   └── a-blueprint
│   │       ├── blueprint.py
│   │       ├── __init__.py
│   │       ├── static
│   │       │   └── js
│   │       │       └── a-script.js
│   │       ├── templates
│   │       │   └── base.html
[...]

PR #1339, opened 2015-02-05
It resolves problems with blueprints and url_prefix "/"

guilleJB:
Issue #348 

I want to separate an app with various blueprints, frontend, backend, etc.. But when I try to register my frontend blueprint with follow code, my static css, images, js don't load.
``` python
front = Blueprint('front', __name__, static_folder='static', template_folder='./templates', url_prefix='/')
```
After debugging I found the problem. It's when try to register route, it tries to register route like //static (https://github.com/mitsuhiko/flask/blob/master/flask/blueprints.py#L68)

And then when use url_for the werkzeug routing.py (https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/routing.py#L1655) resolve url_for //static/filename it found a problem beacause there are two rules like this: 
```python
[<Rule '/static/<filename>' (HEAD, OPTIONS, GET) -> static>, <Rule '//static/<filename>' (HEAD, OPTIONS, GET) -> front.static>]
```

So, a simple solution like concatenate a `blueprint.name` before url_prefix when is "/" resolve this conflict and now we can  use `url_for('front.static', filename='2.jpg')` routing resolves without problem.

### DIFF
--- a/flask/blueprints.py
+++ b/flask/blueprints.py
@@ -147,7 +147,10 @@ class Blueprint(_PackageBoundObject):
         self._got_registered_once = True
         state = self.make_setup_state(app, options, first_registration)
         if self.has_static_folder:
-            state.add_url_rule(self.static_url_path + '/<path:filename>',
+            rule = self.static_url_path
+            if self.url_prefix == '/':
+                rule = self.name + rule
+            state.add_url_rule(rule + '/<path:filename>',
                                view_func=self.send_static_file,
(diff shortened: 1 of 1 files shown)


untitaker:
Hello @guilleJB, thanks for your contribution! Sorry for making you wait so long, but I don't believe this is a correct fix. It special-cases a particular setting of yours, but I'm sure there's a bug somewhere else.

PR #1341, opened 2015-02-06
Use PEP440-compatible dev version

ThiefMaster:
This fixes the problem of python packaging considering an installed dev version older than 0.0.


jeffwidman:
I noticed this format has since been reverted to the old behavior.

> This fixes the problem of python packaging considering an installed dev version older than 0.0.

Is this no longer a problem?

untitaker:
The release script needs modification, not just the current dev version.

Issue #1135, opened 2014-07-27
Deprecate flask.ext and create migration tool to flask_*


DasIch:

keyanp:
I'd like to try to tackle this. Just to get some clarification, the proposed migration tool would essentially search old sourcecode and replace instances of `import flask.ext.foo` with `import flask_foo`, am I understanding this correctly?

ThiefMaster:
Yes. It would have to cover these imports (unless I missed anything):

```
from flask.ext import foo
from flask.ext.foo import <anything
import flask.ext.foo
```

keyanp:
And the aforementioned imports would be replaced with what?

ThiefMaster:
flask.ext.foo => flask_foo

untitaker:
Exactly what ThiefMaster said. Thanks for giving this a try!

>And the aforementioned imports would be replaced with what?

* ``from flask.ext import foo`` => ``import flask_foo as foo``
* ``from flask.ext.foo import bam`` => ``from flask_foo import bam``
* ``import flask.ext.foo`` => We'd have to rewrite any reference to ``flask.ext.foo`` in the code further below. If that's too hard, just skip those and show a warning.

davidism:
If you can actually evaluate the import, you could try to figure out the actual package name, rather than guessing that it's `flask_<name>`.

untitaker:
True, but then you'd be dependent on whether the migration script is running in the correct virtualenv, and things like that. Also I don't think anybody actually uses ``flaskext`` extensions anymore.

lepture:
Let's correct every answers on <http://stackoverflow.com/search?q=flask.ext>, and encourage users to use `flask_foo` instead of `flask.ext.foo`.

There are many Flask extentions whose documentation contains `from flask.ext.foo import bar`. We should encourage the developers to change the documentation. For example: https://pythonhosted.org/Flask-Testing/

keyanp:
@lepture that's certainly ambitious, I think it would be nice to resolve #1367 and #1366 before doing something like that. The current documentation still suggests using the `.ext` format. It will be easier to encourage a transition once the docs are fixed and use of the migration tool is fully supported and documented.

lepture:
@keyanp in the meanwhile, we should notify Flask extention developers. Their documentations would also confuse users.

keyanp:
@lepture sure, we can certainly try to let some developers know, alerting **all** extension developers is probably unrealistic.

lepture:
[...]

PR #1342, opened 2015-02-06
flask.ext Import Migration Tool #1135

keyanp:
This is an early commit to get some feedback. This script supports single imports in two formats. I am concerned that the naive approach is not what you guys are looking for. I would obvious add support for multiple imports and the third case `import flask.ext.foo` => `import flask_foo`. But the solution is definitely super hacky.


untitaker:
>This script supports single imports in two formats. I am concerned that the naive approach is not what you guys are looking for.

Yeah... There are some forms with which this script would break:

    from flask.ext.foo import \
        a, \
        b

    from flask.ext.foo import (
        a,
        b
    )

Parsing definetly should be delegated to some third-party library. As said in #1135, https://github.com/python-modernize/python-modernize looks great, but I am not sure if one can easily extend it with own fixers.

keyanp:
I looked into using lib2to3 initially but I didn't look at python-modernize. Upon inspection I think writing a custom fixer wouldn't be too bad. I'll work on this over the weekend. Thanks for the feedback.

untitaker:
@keyanp Just found https://github.com/Psycojoker/baron, maybe you'll find it useful.

keyanp:
@untitaker Thanks! That looks promising, I've been fiddling with writing a fixer for lib2to3, but it's pretty confusing. Hopefully this will be easier.

keyanp:
@untitaker So RedBaron ended up being perfect for the job, the new module supports everything mentioned in issue #1135 except it doesn't change imports in the format:

`import flask.ext.foo.[more stuff]`

it only handles:

`import flask.ext.foo`

And even it that case, it doesn't fix every function call in the source. I also added a few tests, but RedBaron needs to be installed on Travis in order to pass. Let me know how you want me to handle the above issue.

scripts/flaskext_migrate.py
```diff
@@ -0,0 +1,94 @@
...
+def fix_from_imports(red):
+    """
+    Converts "from" style imports to not use "flask.ext".
+
+    Handles:
+    Case 1: from flask.ext.foo import bam --> from flask_foo import bam
+    Case 2: from flask.ext import foo --> import flask_foo as foo
+    """
+    from_imports = red.find_all("FromImport")
+    for x in range(len(from_imports)):
```
untitaker: You can use ``enumerate`` here.

untitaker:
Regarding renaming functions, I think finding all ``NameNode`` instances and replacing them will succeed for most cases (although it might refactor code that doesn't need it, but I don't think we should care about that). That would probably also make it easier to handle the other import type properly.

There's an edge case that your tests don't cover yet, although I don't know if your code does:

    from flask.ext.foo import bar as baz

That said, this looks *really* nice! Apart from the inline code comments I gave, please also add a simple usage message if the user hasn't provided ``sys.argv[1]``.

keyanp:
@untitaker So I added a test for that case and fixed the logic to cover it. Also added a message if no argument is provided. I still need to fix another edge case, I should push that soon. Just a couple things I have questions about:
1. How do I pip install RedBaron on the Travis build so I can pass the build?
2. What is the correct way to import the script into my tester file? I'm just using: `sys.path.append('scripts')` right now, which works locally, but I don't think it will work when building on Travis.

untitaker:
>1. How do I pip install RedBaron on the Travis build so I can pass the build?

Enter it as dependency directly after ``greenlet`` under ``deps= ...``

>What is the correct way to import the script into my tester file?

I was about to suggest that you move the tests for this script to the same directory as the script itself. py.test should pick it up automatically.

keyanp:
Great! Thanks!

keyanp:
So I moved the test file, added RedBaron to the dependencies, and also added an extra test for this case (which is now covered):

>from flask.ext.foo import (
>    a,
>  b
>)

untitaker:
Yeah, we still have to fix imports like these:

    import flask.ext.foo

Just warning about it is unfortunately not an option. :(

keyanp:
Yeah no problem, could you refer to the specific case you are talking about? The case you wrote in the last comment already works: import flask.ext.foo --> import flask_foo

untitaker:
I meant rewriting the usages of ``flask.ext.foo`` to the new form inside the code itself. See my comment above about NameNode.

keyanp:
@untitaker So I couldn't figure out a good way to do it using the NameNodes, but the function calls are represented by Atomtrailers and replacing those worked for some test files I ran it against. I added a simple test to cover that case as well. Otherwise I refactored the code a bit.

scripts/flaskext_migrate.py
```diff
@@ -0,0 +1,147 @@
...
+    Modifies function calls in the source to reflect import changes.
+
+    Searches the AST for AtomtrailerNodes and replaces them.
+    """
+    atoms = red.find_all("Atomtrailers")
+    for x, node in enumerate(atoms):
+        try:
+            if (node.value[0].value == 'flask' and
+               node.value[1].value == 'ext'):
+                node.replace("flask_foo%s" % node.value[3])
```
untitaker: I think that's a leftover, "foo" is not supposed to be hardcoded. :)

scripts/test_import_migration.py
```diff
@@ -0,0 +1,63 @@
...
+def test_parens_import():
+    red = RedBaron("from flask.ext.foo import (bar, foo, foobar)")
+    output = migrate.fix_tester(red)
+    assert output == "from flask_foo import (bar, foo, foobar)"
+
+
+def test_function_call_migration():
+    red = RedBaron("flask.ext.foo(var)")
+    output = migrate.fix_tester(red)
+    assert output == "flask_foo(var)"
```
untitaker: This works, but this doesn't:

    def test_function_call_migration():
        red = RedBaron("import flask.ext.foo\n\n"
                       "flask.ext.foo.bar(var)")
        output = migrate.fix_tester(red)
        assert output == ("import flask_foo\n\n"
                          "flask_foo.bar(var)")

scripts/flaskext_migrate.py
keyanp: Woops, good catch. Fixed that.

scripts/test_import_migration.py
keyanp: Okay, there are a few problems with this. Let me work on it. Thanks for looking it over.

keyanp:
@untitaker Refactored to clean it up a bit, added the test you mentioned, and added a function to cover it. Thanks again for your help.

DasIch:
Looks good to me.

DasIch:
@untitaker What's your opinion?

untitaker:
In IRC I expressed the concern that this tool might rewrite things that it is not supposed to rewrite, because it doesn't do a lot of "checks" on the AST.

But since this is just a feeling I have, this is actually pretty great. Thanks @keyanp!

davidism:
This moved to its own repo: [pocoo/flask-ext-migrate](https://github.com/pocoo/flask-ext-migrate).  This should probably be removed from the Flask repo and the documentation should be updated to mention the migration tool.

untitaker:
Agreed, please file a new issue with a blocker label and appropriate milestone.

On 4 April 2016 20:23:10 CEST, David Lord <notifications@github.com> wrote:
>This moved to its own repo:
>[pocoo/flask-ext-migrate](https://github.com/pocoo/flask-ext-migrate). 
>This should probably be removed from the Flask repo and the
>documentation should be updated to mention the migration tool.
>
>---
>You are receiving this because you modified the open/close state.
>https://github.com/pallets/flask/pull/1342#issuecomment-205433643

-- 

untitaker:
For future reference, followup on https://github.com/pallets/flask/issues/1765

PR #1346, opened 2015-02-07
Remove bad security advice about send_file, fixes #1345

### LINKED ISSUE #1345: flask.send_file gives bad security advice
On http://flask.pocoo.org/docs/latest/api/#flask.send_file the following is suggested:

> Please never pass filenames to this function from user sources without checking them first. Something like this is usually sufficient to avoid security problems:

>     if '..' in filename or filename.startswith('/'):
           abort(404)

This is *terrible* advice. It would be much better to tell people that if they want to pass in filenames, they should use http://flask.pocoo.org/docs/latest/api/#flask.send_from_directory instead.

ezyang:
Signed-off-by: Edward Z. Yang <ezyang@cs.stanford.edu>

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -427,12 +427,8 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
     guessing requires a `filename` or an `attachment_filename` to be
     provided.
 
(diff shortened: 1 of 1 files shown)


untitaker:
Thanks!

PR #1358, opened 2015-02-26
Add skip to fix unnoticed bug with import migration tool

keyanp:
See pull request #1342. Fixes logic so that imports that should not be changed are skipped, which was not happening correctly before. This was huge oversight on my part. I added a test for normal imports that shouldn't be touched.


untitaker:
Thanks!

PR #1360, opened 2015-02-28
Add exception details to response

keyanp:
This pull request is in regards to issue #1317. I reproduced the 400 error that kenjones-cisco was talking about using this code: gist.github.com/keyanp/bc4d1a3a1e6c5cb5170a

So I took a look at making a subclass of BadRequest, but indeed this is part of Werkzeug. So instead I was thinking of adding the ValueError content to the shell console or to the response body. By just adding the exception to the response body it at least improves the 400 message and probably will make it easier to debug. The response body (which is also in the gist above) would now look like:

```
<title>400 Bad Request</title>
<h1>Bad Request</h1>
<p>No JSON object could be decoded</p>
```

If this isn't really that helpful perhaps I can just output the exception details to the console? Could someone point me in the direction of where in the codebase Flask handles console logging?

### DIFF
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -175,7 +175,7 @@ class Request(RequestBase):
 
         .. versionadded:: 0.8
         """
-        raise BadRequest()
+        raise BadRequest(e)
 
     def _load_form_data(self):
         RequestBase._load_form_data(self)
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -177,7 +177,7 @@ class Request(RequestBase):
         ctx = _request_ctx_stack.top
         if ctx is not None:
             if ctx.app.config.get('DEBUG', False):
-                raise BadRequest(e)
+                raise BadRequest('No JSON object could be decoded')
         raise BadRequest()
 
     def _load_form_data(self):
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -174,10 +174,11 @@ class Request(RequestBase):
 
         .. versionadded:: 0.8
         """
-        if current_app.config['DEBUG']:
-            raise BadRequest(e)
-        else:
-            raise BadRequest()
+        ctx = _request_ctx_stack.top
+        if ctx is not None:
+            if ctx.app.config.get('DEBUG', False):
+                raise BadRequest(e)
+        raise BadRequest()
 
     def _load_form_data(self):
         RequestBase._load_form_data(self)
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -13,8 +13,7 @@ from werkzeug.wrappers import Request as RequestBase, Response as ResponseBase
 from werkzeug.exceptions import BadRequest
 
 from . import json
-from .globals import _request_ctx_stack
-
+from .globals import _request_ctx_stack, current_app
 
 _missing = object()
 
@@ -175,7 +174,10 @@ class Request(RequestBase):
 
         .. versionadded:: 0.8
         """
-        raise BadRequest(e)
+        if current_app.config['DEBUG']:
+            raise BadRequest(e)
+        else:
+            raise BadRequest()
 
     def _load_form_data(self):
         RequestBase._load_form_data(self)
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -177,7 +177,7 @@ class Request(RequestBase):
         ctx = _request_ctx_stack.top
         if ctx is not None:
             if ctx.app.config.get('DEBUG', False):
-                raise BadRequest('No JSON object could be decoded')
+                raise BadRequest('Failed to decode JSON object: {}'.format(e))
         raise BadRequest()
 
     def _load_form_data(self):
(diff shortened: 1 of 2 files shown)


flask/wrappers.py
```diff
@@ -175,7 +174,10 @@ def on_json_loading_failed(self, e):
 
         .. versionadded:: 0.8
         """
-        raise BadRequest()
+        if current_app.config['DEBUG']:
```
untitaker: DEBUG is technically around all the time but I feel that we shouldn't fail if the user has somehow managed to corrupt their config => use `config.get('DEBUG', False)`.

untitaker: Also, what if current_app is not around?

untitaker:
Looks pretty good, if you can manage, please also add tests (I can do that too if you don't have time)

Thanks!

flask/wrappers.py
ThiefMaster: How would you ever end up loading json from a `Request` outside an app context?

untitaker: I mean that some user might e.g. want to use Flask's extended request objects without the rest of Flask... take a look at `max_content_length` where similar precautions are made.

keyanp:
Sure I'll add tests, doesn't seem to be a consensus on checking for DEBUG setting though. Do you suggest using `_app_ctx_stack.top` similar to its use for `max_content_length`?

untitaker:
>Do you suggest using _app_ctx_stack.top similar to its use for max_content_length?

I was thinking about using `current_app` and catching RuntimeError, but using the stack directly as you say seems more consistent with the rest of the code.

keyanp:
@untitaker Thanks for looking over the previous commits. I added a couple tests and changed the logic to use `_app_ctx_stack` similar to `max_content_length`. Let me know if you see anything that still needs work.

tests/test_helpers.py
```diff
@@ -49,6 +49,30 @@ def test_jsonify_date_types(self):
...
+        app = flask.Flask(__name__)
+        app.config['DEBUG'] = False
+        @app.route('/json', methods=['POST'])
+        def post_json():
+            flask.request.get_json()
+            return None
+        c = app.test_client()
+        rv = c.post('/json', data=None, content_type='application/json')
+        assert rv.status_code == 400
+        assert 'No JSON object could be decoded' not in rv.data
```
untitaker: You need a `b'...'` string here.

tests/test_helpers.py
```diff
@@ -49,6 +49,30 @@ def test_jsonify_date_types(self):
...
+        @app.route('/json', methods=['POST'])
+        def post_json():
+            flask.request.get_json()
+            return None
+        c = app.test_client()
+        rv = c.post('/json', data=None, content_type='application/json')
+        assert rv.status_code == 400
+        assert 'No JSON object could be decoded' in rv.data
+
+    def test_post_empty_json_doesnt_add_exception_to_reponse_if_no_debug(self):
```
untitaker: Typo in function name.

tests/test_helpers.py
```diff
@@ -49,6 +49,30 @@ def test_jsonify_date_types(self):
             assert rv.mimetype == 'application/json'
             assert flask.json.loads(rv.data)['x'] == http_date(d.timetuple())
 
+    def test_post_empty_json_adds_exception_to_reponse_content_in_debug(self):
```
untitaker: Typo in function name.

keyanp:
@untitaker Thanks for looking over the previous commits. I ended up added a couple tests and changed the logic to use `_app_ctx_stack` similar to `max_content_length`.

The issue I was running into with the tests was that Python 3.4 throws a different exception message for trying to load data from an empty json, and so I ended up hardcoding the exception details. I'm not sure how you feel about that, but I think since the function `on_json_loading_failed` is only being called in this specific instance it should be safe to pass a string manually. Let me know your thoughts.

untitaker:
It's true, it's easier for assertion for the error message to be consistent. This can be achieved with `BadRequest('Failed to decode JSON object: {}'.format(e))` too, though.

keyanp:
Changed so that the response content for Python 3.3 and below (in DEBUG mode) is now: 

`Failed to decode JSON object: No JSON object could be decoded`

While for Python 3.4 it is:

`No JSON object could be decoded: Expecting value: line 1 column 1 (char 0)`

flask/wrappers.py
```diff
@@ -175,6 +174,10 @@ def on_json_loading_failed(self, e):
 
         .. versionadded:: 0.8
         """
+        ctx = _request_ctx_stack.top
+        if ctx is not None:
```
untitaker: Use `and` instead of two nested ifs.

flask/wrappers.py
```diff
@@ -13,8 +13,7 @@
 from werkzeug.exceptions import BadRequest
 
 from . import json
-from .globals import _request_ctx_stack
-
+from .globals import _request_ctx_stack, current_app
```
untitaker: Unused import.

untitaker:
For some miraculous reason we now leak the request context... will look into this later.

keyanp:
Yeah I'm not sure why the request context was being leaked, but the error I was getting was specific to formatting strings in Python 2.6, where you are required to provide an index for arguments. Build looks good now though, and I incorporated the changes you commented on.

untitaker:
Excellent, thanks!

PR #1363, opened 2015-03-04
Added closing paragraph tags

pjhampton:
There was no trailing paragraph tags on lines 77 & 78 for some reason.


jacquerie:
A closing paragraph tag is not always needed, according to the HTML spec: http://www.w3.org/TR/html-markup/p.html.

In particular, this should be the case: the first paragraph is followed by another paragraph, and the second paragraph is not followed by any other content.

pjhampton:
Fair enough, thanks for bringing this to my attention. I was confused but I now see.

PR #1365, opened 2015-03-06
Changed implementation of flash messages to use a dict instead of list

mayankgupta173:
The underlying implementation of flash'ed messages currently uses a list of category, message tuples. This commit changes that to the more natural dictionary from category -> list of messages implementation. This still allows duplicate messages per category, but makes adding flash messages use fewer object allocations, and filtering flash messages by category faster by completely ignoring messages not in the desired set of categories.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -365,8 +365,11 @@ def flash(message, category='message'):
     # This assumed that changes made to mutable structures in the session are
     # are always in sync with the session object, which is not true for session
     # implementations that use external storage for keeping their keys/values.
-    flashes = session.get('_flashes', [])
-    flashes.append((category, message))
+    flashes = session.get('_flashes', {})
+    if category not in flashes:
+        flashes[category] = []
+    category_flashes = flashes[category]
+    category_flashes.append(message)
     session['_flashes'] = flashes
     message_flashed.send(current_app._get_current_object(),
                          message=message, category=category)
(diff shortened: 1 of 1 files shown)


davidism:
Of course, now you've lost the order that messages were flashed in, if they were flashed in more than one category.

ThiefMaster:
Indeed, I think it makes sense to preserve the order.

flask/helpers.py
```diff
@@ -365,8 +365,11 @@ def flash(message, category='message'):
     # This assumed that changes made to mutable structures in the session are
     # are always in sync with the session object, which is not true for session
     # implementations that use external storage for keeping their keys/values.
-    flashes = session.get('_flashes', [])
-    flashes.append((category, message))
+    flashes = session.get('_flashes', {})
+    if category not in flashes:
```
ThiefMaster: category_flashes = flashes.setdefault('category', [])

flask/helpers.py
```diff
@@ -401,14 +404,27 @@ def get_flashed_messages(with_categories=False, category_filter=[]):
...
+    
+    results = []
     if not with_categories:
-        return [x[1] for x in flashes]
-    return flashes
+        for flash_values in filtered_flashes.itervalues():
+            results += flash_values
+    else:
+        for flash_category, flash_values in filtered_flashes.iteritems():
+            results += map(lambda x: (flash_category, x), flash_values)
```
ThiefMaster: NO! You should **never** use `map` with a lambda. It's inefficient and unpythonic. The proper construct here is a list comprehension:

    results += [(flash_category, x) for x in flash_values]

mayankgupta173:
I not consider this use case. I agree.

keyanp:
@mayankgupta173  are you still working on this?

RonnyPfannschmidt:
it should be noted that if the number of flash messages is so high that this optimization yields notable performance gains, the user already is in hell

keyanp:
@RonnyPfannschmidt can we close this then? Looks like the consensus is that this is a bad change.

untitaker:
Yes, ok.

PR #1368, opened 2015-03-08
Configure pytest's norecursedirs

ThiefMaster:
Skip directories which don't contain unittests and especially virtualenvs (in folders named `env*`).
Otherwise you get ugly failures when running tests because pytest descends into the virtualenv.


untitaker:
Thanks!

PR #1369, opened 2015-03-08
Add support for blueprint-level cli commands

### LINKED ISSUE #1357: Support for @blueprint.cli.command
It would be great to have access to the functionality of `app.cli.command` via blueprints - i.e. something like:

``` python
blueprint = Blueprint('myblueprint', __name__,)

@blueprint.cli.command()
def doblueprintstuff():
    """Initialize the database."""
    #....
```

It would help a lot when building modular applications.

ThiefMaster:
Still WIP, but that way it can already be reviewed and commented on.

Tests are still missing and so is a `Blueprint.app_cli` which registers app-level commands from a blueprint (which I think makes sense depending on how one uses blueprints).

When finished, this will close #1357

### DIFF
--- a/flask/blueprints.py
+++ b/flask/blueprints.py
@@ -9,9 +9,65 @@
     :copyright: (c) 2015 by Armin Ronacher.
     :license: BSD, see LICENSE for more details.
 """
+from collections import defaultdict
 from functools import update_wrapper
 
+import click
+import click.decorators
+
+from . import cli
+from ._compat import iteritems
 from .helpers import _PackageBoundObject, _endpoint_from_view_func
+
+
+class BlueprintGroup(object):
+    def __init__(self):
+        self._default_group = object()
+        self.commands = defaultdict(dict)
+
+    def add_command(self, cmd, name=None, group=None):
+        """Registers another :class:`Command`.  If the name is not provided,
+        the name of the command is used.
+        """
+        name = name or cmd.name
+        if name is None:
+            raise TypeError('Command has no name.')
+        self.commands[group][name] = cmd
+
+    def command(self, *args, **kwargs):
+        """This works exactly like the method of the same name on a regular
+        :class:`click.Group` but it wraps callbacks in :func:`with_appcontext`
+        unless it's disabled by passing ``with_appcontext=False``.
+        """
+        wrap_for_ctx = kwargs.pop('with_appcontext', True)
+        group = kwargs.pop('group', self._default_group)
+        def decorator(f):
+            if wrap_for_ctx:
+                f = cli.with_appcontext(f)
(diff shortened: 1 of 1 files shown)


flask/blueprints.py
```diff
@@ -105,6 +106,19 @@ def __init__(self, name, import_name, static_folder=None,
         if url_defaults is None:
             url_defaults = {}
         self.url_values_defaults = url_defaults
+        if cli_group_name is None:
+            cli_group_name = name
+        self.cli_group_name = cli_group_name
+        if cli_group_desc is None:
+            cli_group_desc = 'Commands provided by {0}.'.format(self.name)
```
untitaker: I think given the context, it should be clarified that this is a blueprint.

ThiefMaster: I initially had a longer description (*"Commands provided by the ... blueprint"*), but click ended up truncating it...

untitaker: "Commands by the {} blueprint" maybe?

ThiefMaster: good idea; that way only the "blueprint" is truncated if the name is too long, and at least with my test blueprint ("yellowprint") nothing was truncated at all

untitaker:
Maybe registering global commands should be done like with routes: During registration, provide a prefix, otherwise it's global by default.

ThiefMaster:
Sounds fine, but that might get a bit ugly when you end up using the same group/prefix in two blueprints. Each of them *must* create the group immediately (since you can nest them and recording all the nested calls would be somewhat of an overkill), so they'd have to be merged during registration.

After looking into `click.Group` that might not be such a big problem though, since I'd just have to merge the `commands` dict from the new group with the existing group's command dict...

untitaker:
Sounds tricky, but having `app_cli` before registering to the app sounds equally tricky.

ThiefMaster:
Managed to get it working. What I still don't like that much is the fact that you can only specify additional arguments such as the help text for groups created using the default cli_group of the the blueprint. Adding another kwarg to the decorators is not really a good idea since the group description is not really related to a command.

Maybe the better solution would be getting rid of the `group` kwarg and simply creating the group using the decorator and creating its commands using the decorator provided by click.Group.command...

ThiefMaster:
Simplified it a lot in the second commit. I think it's more straightforward and especially closer to the regular `click` API like that.

ThiefMaster:
I guess I still need to merge groups if two blueprints define the same one. Going to have a look at that after work...

untitaker:
Hey @ThiefMaster what's the status of this?

ThiefMaster:
It's not really clear what's the best way to implement this so it probably makes more sense to close this for now...

Issue #532, opened 2012-06-25
Mention Extensions in the Docs more


kennethreitz:
I had a friend that was going through the quickstart, and he had absolutely no idea that there were any extensions available for flask.

He was using SQLAlchemy directly and was becoming quite frustrated quickly.

Since an ecosystem is clearly part of a framework, I think we should add some hints about extensions throughout the documentation. Not necessarily usage examples, just "hey, this extension might be useful to you".

KWinston:
Hi, I'm a new contributor and would like to work on this issue. Where in the documentation are you referring to within the quickstart?

untitaker:
There's already some work done in #1376, but I'll leave this open because there are surely more places where we can include extension hints.

davidism:
Looks like we link to them more now, so closing this.

kennethreitz:
✨🍰✨

PR #1376, opened 2015-03-12
Update quickstart.rst doc for awareness of Flask extensions

KWinston:
Addresses concerns in issue #532 regarding not knowing about flask extensions as it is not mentioned the quick start guide. Maybe this can help increase the awareness of flask extensions to new users?


untitaker:
That seems sufficient, thanks!

PR #1380, opened 2015-03-13
Add line of code required to run examples for Other Testing Tricks section

KWinston:
As it stated, documentation states full example that demonstrates this approach, with the way it is implemented. It requires this line of code to perform the test functions on the three examples under Other Testing Tricks.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -207,6 +207,8 @@ temporarily.  With this you can access the :class:`~flask.request`,
 :class:`~flask.g` and :class:`~flask.session` objects like in view
 functions.  Here is a full example that demonstrates this approach::
 
+    import flask
+    
     app = flask.Flask(__name__)
 
     with app.test_request_context('/?name=Peter'):


keyanp:
Hi @KWinston, in the future it would be helpful make your PR title more specific and clear. The commit message you have is much less ambiguous and would be a good PR title as well.

To address the suggested change, it is meant to be implied that the code provided in that subsection is part of the same module defined above in the "The Testing Skeleton" subsection. Therefore an import statement is not necessary. Notice how the sample code in the "Logging In and Out" and "The First Test" sections similarly does not include import statements.

KWinston:
Okay, I will keep note of that. In this case, import flask is not in "The Testing Skeleton" either meaning this section of code will not be able to run without import flask.
Should it be in "The Testing Skeleton"?

Other lines such as in Faking Resources and Context has
from flask import appcontext_pushed
which allow it to use appcontext_pushed.function()
instead of flask.appcontext_pushed.function()

There is a differences of expressions in the snippets.

untitaker:
I think this is fine as-is. The testing skeleton covers a specific example, the code blocks there are completely distinct.

untitaker:
Thanks!

PR #1383, opened 2015-03-16
Fix errorhandler HTTPException normalization

F30:
This is barely noticeable in the development code since #839, but in the latest released version. The current code to normalize the error condition misses its apparent purpose: code_or_exception will never be an instance of HTTPException, as one might only supply an exception class and not an object.
The bug still appears when you declare an errorhandler for a numeric code after setting one for the respective exception: Then, the exception's handler will always take precedence.

The HTTPException base class has no code and therefore must be dealt with like any other exception.


davidism:
The test case passes without changes in master.

PR #1390, opened 2015-03-23
before_render_template signal

### LINKED ISSUE #1370: before_render_template signal
This signal would receive the name of the template and the context and would be able to modify the context before the template is actually rendered. It could even go one step further and use the return value from the signal instead of rendering the template if it's not None.

alexpantyukhin:
fix #1370


flask/signals.py
```diff
@@ -52,4 +53,4 @@ def _fail(self, *args, **kwargs):
 appcontext_tearing_down = _signals.signal('appcontext-tearing-down')
 appcontext_pushed = _signals.signal('appcontext-pushed')
 appcontext_popped = _signals.signal('appcontext-popped')
-message_flashed = _signals.signal('message-flashed')
+message_flashed = _signals.signal('message-flashed')
```
ThiefMaster: You removed the trailing newline in that file. Can you please revert that (simply amend the commit and force-push)? Python files should end with a linebreak.

tests/test_signals.py
```diff
@@ -46,6 +46,30 @@ def record(sender, template, context):
...
+    def index():
+        return flask.render_template('simple_template.html', whiskey=42)
+
+    recorded = []
+
+    def record(sender, template, context):
+        context['whiskey'] = 43
+        recorded.append((template, context))
+
+    flask.before_render_template.connect(record, app)
```
ThiefMaster: You can use the [context manager](https://pythonhosted.org/blinker/#blinker.base.Signal.connected_to) to connect the signal temporarily

ThiefMaster:
Can you add a `closes #1370` somewhere in the commit message so it closes the related issue #1370 once it's merged?

ThiefMaster:
What about checking if there's a not-None return value from that signal? If there is exactly one, you could return that instead of rendering the original template. This would allow one to override the template rendering process if necessary. Could be useful in unit tests if all you care about is the context but don't want the actual template to be rendered for some reason.

untitaker:
>Can you add a closes #1370 somewhere in the commit message so it closes the related issue #1370 once it's merged?

I added it to the PR body, I hope that helps. Doesn't really matter though.

alexpantyukhin:
Hi, all!
Adrian, thank you for your review!

>What about checking if there's a not-None return value from that signal? If there is exactly one, you could return that instead of rendering the original template. This would allow one to override the template rendering process if necessary. Could be useful in unit tests if all you care about is the context but don't want the actual template to be rendered for some reason.

Actually it is not so clear for me. Do you want to use this feature only for unit tests? Because I think this will look a little bit strange in the code.

ThiefMaster:
It would be one case. Another use-case could be adding extra functionality in a application that supports plugins.

For example, in the project I'm working on we have a plugin system where we let plugins override templates in some cases. While I'd mostly use the `before_render_template` signal to inject extra data in the context when overriding a template, I could imagine cases where it'd make more sense to completely intercept the rendering of a certain template.

alexpantyukhin:
>You can use the context manager to connect the signal temporarily

It seems that test env uses blinker v 1.0. This version doesn't have connected_to method. I think that old approach in test is quite ok.

ThiefMaster:
I would merge the "fix" commits into the related commit (i.e. do an interactive git rebase, and mark those commits as "fixup").

flask/templating.py
```diff
@@ -101,7 +101,19 @@ def list_templates(self):
...
 def _render(template, context, app):
-    """Renders the template and fires the signal"""
+    """Renders the template and fires signals.
+
+    It is possible to override render template in the before_render_template signal.
+    It can be done only if exactly one receiver and it return not None result."""
+
+    brt_resp = before_render_template.send(app, template=template, context=context)
+    
+    if len(brt_resp) == 1:
```
ThiefMaster: I think this can be simplified a bit:

```python
if len(brt_resp) == 1 and brt_resp[0][1] is not None:
    return brt_resp[0][1]
```

Signals always return `func, func_return_value` tuples, so you don't need to check that you have two elements.

Another possibility would be this:

```python
overrides = [rv for _, rv in brt_resp if rv is not None]
if len(overrides) == 1:
    return overrides[0]
elif len(overrides) > 1:
    raise RuntimeError('More than one before_render_template signal '
                       'returned data')
```

alexpantyukhin:
Yes, I agree, thx!
It's done.

flask/templating.py
```diff
@@ -101,7 +101,20 @@ def list_templates(self):
...
 def _render(template, context, app):
-    """Renders the template and fires the signal"""
+    """Renders the template and fires signals.
+
+    It is possible to override render template in the before_render_template signal.
+    It can be done only if exactly one receiver and it return not None result."""
+
+    brt_resp = before_render_template.send(app, template=template, context=context)
+    
+    overrides = [rv for _, rv in brt_resp if rv is not None]
```
untitaker: What happens if a signal subscriber itself calls `render_template` to override the normal `render_template` call?

ThiefMaster: Hm... I guess it would trigger the signal again, and if it's not written properly it would result in an endless loop.

However, I'd assume any function subscribing to this signal would not override *any* template, so if it's called again for a template it renders to override one, it'd be expected not to intercept this call, too.

untitaker: Similar to how handling errors from errorhandlers works, I think Flask should do more to avoid these kind of situations, for example by somehow not firing the signal a second time (no idea how), or by allowing the signal user to only modify the context instead of overriding the return value (as you mentioned in the original issue).

alexpantyukhin: Is there some way to intercept function call with decorator? If there is then it's possible to override render functions.
I tried to find some "depth" parameter in the blinker lib but unfortunately it doesn't exist. I think it would not bad feature.

ThiefMaster: Nope, all a decorator can do is this:

```
def foo():
    ...
foo = bar(foo)
```

So without monkeypatching a decorator won't help.

ThiefMaster: One way to avoid all kinds of trouble would be moving the `_render` util into the `Flask` class. That would allow one to keep using the `flask.render_template` function (which is incredibly useful, since you have *one* place to import it from) but still customize its behavior on the application level.

I'd keep the signal nonetheless to modify the context but ignore the return value.

untitaker:
@ThiefMaster @alexpantyukhin What's up with this PR?

ThiefMaster:
I guess the simple way of only sending the signal and not caring about the return value is the most straightforward way... the extra functionality I had initially suggested probably fits much better in a separate PR (that moves `_render` into the `Flask` class to allow overriding it if necessary).

alexpantyukhin:
I agree with @ThiefMaster .

untitaker:
Ok, then this implementation seems fine to me. If you agree @alexanderpantyukhin please add a changelog and proper docs for this signal. Thanks!

alexpantyukhin:
It seems that only person who has 'write access' can resolve conflicts in merge.

untitaker:
Thanks @alexanderpantyukhin!

merge conflicts can be resolved by [rebasing](https://help.github.com/articles/about-git-rebase/) your branch against master.

ThiefMaster:
The history of that PR could have been cleaned up a bit :/ it's rather messy with all the intermediate "fix ..." commits...

Issue #1392, opened 2015-03-23
AppContext.pop() should use a different sentinel for the exception value.


mjpieters:
Consider the following:

```python
from flask import Flask
app = Flask(__name__)

@app.teardown_appcontext
def teardown(exception):
    print exception

with app.app_context():
    try:
        raise ValueError()
    except ValueError:
        pass
```

Within the `app.app_context()` context (managed by `with`) an exception is raised and handled. However, the `teardown_appcontext()` handler is still passed the exception value.

That's because the exception was raised and handled within the same frame, and thus `sys.exc_info()` is still available. The `AppContext.__exit__()` method did pass in `None` for the exception value:

```python
def __exit__(self, exc_type, exc_value, tb):
    self.pop(exc_value)
```

but the `AppContext.pop()` method cannot distinguish this from not passing in a value at all:

```python
def pop(self, exc=None):
    """Pops the app context."""
    self._refcnt -= 1
    if self._refcnt <= 0:
        if exc is None:
            exc = sys.exc_info()[1]
        self.app.do_teardown_appcontext(exc)
```

This can easily be remedied by using a different sentinel default value:

```python

_sentinel = object()

def pop(self, exc=_sentinel):
    """Pops the app context."""
    self._refcnt -= 1
    if self._refcnt <= 0:
        if exc is _sentinel:
            exc = sys.exc_info()[1]
        self.app.do_teardown_appcontext(exc)
```

mjpieters:
This applies to the request context as well, plus the `app.do_teardown_request()` and `app.do_teardown_appcontext()` methods. I'll create a pull request.

untitaker:
Fixed by #1393

PR #1393, opened 2015-03-23
Ignore handled exception when popping the context

mjpieters:
This is a fix for #1392; the app and request contexts should not report an exception when tearing down after an exception has already been handled in a `with` statement.


untitaker:
Excellent, thanks!

PR #1401, opened 2015-03-27
Fix Possible Typo

bsandrow:
Looks like that was meant to be `config_key`. It works by accident because the function is defined in the same scope as the loop that passes `config_key` to `apprunner`.


untitaker:
Thanks for bringing this to attention! The whole testcase was a bit smelly, I rewrote it.

untitaker:
Actually I had to revert this, relevant PR: https://github.com/mitsuhiko/flask/pull/1407

PR #1403, opened 2015-03-27
fix bad autodoc signature when functions are decorated

ldgeo:
It's very annoying when you read the API documentation, all decorated functions haven't the correct signature.

Example : 

`errorhandler(*args, **kwargs)`

should be 

`errorhandler(code_or_exception)`

this PR fixes this issue with an elegant solution found here : https://github.com/praw-dev/praw/blob/master/praw/decorators.py

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,6 +19,8 @@ import sys, os
 sys.path.append(os.path.abspath('_themes'))
 sys.path.append(os.path.abspath('.'))
 
+os.environ['SPHINX_BUILD'] = '1'
+
 # -- General configuration -----------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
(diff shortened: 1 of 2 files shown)


untitaker:
Yeah, this is about as elegant as it gets, but it kinda freaks me out that an environment variable could change fundamental behavior of Flask. Will need to think more about this.

ThiefMaster:
https://pypi.python.org/pypi/decorator looks like a cleaner solution, even though it does require decorating all the decorators...

untitaker:
[Holy moly](https://github.com/micheles/decorator/blob/master/src/decorator.py), not sure what I prefer...

untitaker:
I would suggest https://github.com/mitsuhiko/flask/pull/1483 instead of this one. I know it patches the stdlib, but it seems like a much broader solution to me while it avoids making Flask itself "vulnerable" to weird environments.

PR #1405, opened 2015-03-28
Test_import_migration.py updated

amaneight:
From Issue #1135 "Test_import_migration.py" file updated.


untitaker:
This code specifically tests flask.ext.

PR #1408, opened 2015-03-29
JSON support for the Flask test client

adambyrtek:
This change adds support for JSON requests to the Flask test client, which makes testing JSON APIs much more convenient, and makes the arguments consistent with `request.data` and `request.json`.

Before:
```response = client.post('/url', data=json.dumps(json_data), content_type='application/json')```

After:
```response = client.post('/url', json=json_data})```

If this gets merged then I have another change to support `response.json`, which would make it easier to make assertions on JSON responses.

### DIFF
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -37,7 +37,7 @@ def make_test_environ_builder(app, path='/', base_url=None, json=None, *args, **
 
     if json is not None:
         if 'data' in kwargs:
-            raise RuntimeError('Client cannot provide both `json` and `data`')
+            raise ValueError('Client cannot provide both `json` and `data`')
         kwargs['data'] = json_dumps(json)
 
         # Only set Content-Type when not explicitly provided
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -35,16 +35,14 @@ def make_test_environ_builder(app, path='/', base_url=None, json=None, *args, **
             if url.query:
                 path += '?' + url.query
 
-    if json:
+    if json is not None:
         if 'data' in kwargs:
             raise RuntimeError('Client cannot provide both `json` and `data`')
         kwargs['data'] = json_dumps(json)
 
         # Only set Content-Type when not explicitly provided
-        if 'Content-Type' not in kwargs.get('headers', {}):
-            new_headers = kwargs.get('headers', {}).copy()
-            new_headers['Content-Type'] = 'application/json'
-            kwargs['headers'] = new_headers
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'application/json'
 
     return EnvironBuilder(path, base_url, *args, **kwargs)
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: flask/testing.py [hunk 1]
```diff
@@ -33,6 +34,18 @@ def make_test_environ_builder(app, path='/', base_url=None, *args, **kwargs):
             path = url.path
             if url.query:
                 path += '?' + url.query
+
+    if json:
```
ThiefMaster: You really want `if json is not None` here. Otherwise `json={}` won't result in a JSON request.

adambyrtek:
I'm looking into the CI failures.

flask/testing.py
adambyrtek: Fixed, thanks.

flask/testing.py
```diff
@@ -33,6 +34,16 @@ def make_test_environ_builder(app, path='/', base_url=None, *args, **kwargs):
             path = url.path
             if url.query:
                 path += '?' + url.query
+
+    if json is not None:
+        if 'data' in kwargs:
+            raise RuntimeError('Client cannot provide both `json` and `data`')
```
untitaker: This should be ValueError.

untitaker:
LGTM, please add a changelog entry and proper documentation.

flask/testing.py
adambyrtek: Done

adambyrtek:
@untitaker Please take a second look. I've updated the changelog, but couldn't find a good place for documentation. There's no API reference for `FlaskClient` besides the Werkzeug one. The user guide for testing doesn't cover JSON requests, and I didn't want to completely change it.

davidism:
Maybe it would be good to extend the json support to Werkzeug's client/response as well.  Then the documentation could go there along with what already exists.

adambyrtek:
@davidism Yeah, that's my next patch, but since this is my first contribution to Flask I decided to submit this piece by piece.

untitaker:
@adambyrtek I think it only makes sense to do this on one of them -- Flask's test client is a subclass of Werkzeug's.

adambyrtek:
@untitaker Sorry, I think I misread the comment. Just to clarify, I don't intend to patch Werkzeug (since I believe JSON support is too high-level for that), but I'd like to extend the Flask response object in a separate pull request to allow:

```
response = client.post('/url', json=json_data})
assert response.json == json_data
```

However, I'm open to any suggestions on how to improve the documentation for this.

untitaker:
Ahh, I see.

untitaker:
Ok, currently this is missing documentation -- AFAICT there's no way to discover this functionality from the online docs.

untitaker:
Otherwise, :+1:

adambyrtek:
@untitaker Sure, I'll try to incorporate this in the testing section.

adambyrtek:
Incorporated into #1416.

PR #1413, opened 2015-03-31
Added note that @errorhandler(500) isn't triggered in debug mode

skybert:
Hi,

I was scratching my head on this one and thought it would be good to mention this in the documentation.

Thanks for an excellent lightweight web framework,

-Torstein

### DIFF
--- a/docs/patterns/errorpages.rst
+++ b/docs/patterns/errorpages.rst
@@ -77,3 +77,6 @@ An example template might be this:
      <p>What you were looking for is just not there.
      <p><a href="{{ url_for('index') }}">go somewhere nice</a>
    {% endblock %}
+
+Please note that if you add an error handler for "500 Internal Server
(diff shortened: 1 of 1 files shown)


keyan:
Hi @skybert, thanks for the clarification! I think this would flow better if you put it before the two example code blocks, what do you think?

skybert:
Hi @keyan, sure, I don't mind. The reason for putting it after the code examples, was that the text at that point has shown the reader how to set up an error handler for 404 errors and I thought the tip about status=500 fitted in after that. But that's not a big deal, the most important thing is that the note is in there, it should save others some head scratching reading through various bug reports/discussion threads through Google.

As for other status codes: I haven't encountered others so far. Don't know the flask source code, but from the look of ```handle_exception(self, e)```, it looks like ```500``` is the only thing that receives special attention.

keyan:
@skybert Thanks!

PR #1414, opened 2015-03-31
Added OpenShift as a free hosting provider alternative

charleswhchan:


keyan:
@charleswhchan Thanks!

PR #1416, opened 2015-04-01
JSON support for test client and response object

adambyrtek:
Common code has been refactored into a mixin shared between request and response classes. This is a continuation of #1408.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,9 @@ Version 1.0
 
 (release date to be announced, codename to be selected)
 
+- Added `json` keyword argument to :meth:`flask.testing.FlaskClient.open`
+  (and related ``get``, ``post``, etc.), which makes it more convenient to
+  send JSON requests from the test client.
 - Added before_render_template signal.
 - Added `**kwargs` to :meth:`flask.Test.test_client` to support passing
   additional keyword arguments to the constructor of
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,8 @@ Version 1.0
 
 (release date to be announced, codename to be selected)
 
+- Added ``is_json`` and ``get_json`` to :class:``flask.wrappers.Response``
+  in order to make it easier to build assertions when testing JSON responses.
 - Added `json` keyword argument to :meth:`flask.testing.FlaskClient.open`
   (and related ``get``, ``post``, etc.), which makes it more convenient to
   send JSON requests from the test client.
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -368,15 +368,17 @@ convenient:
 
     @app.route('/api/auth')
     def auth():
-        email = request.get_json()['email']
-        password = request.get_json()['password']
+        json_data = request.get_json()
+        email = json_data['email']
+        password = json_data['password']
         return jsonify(token=generate_token(email, password))
 
     with app.test_client() as c:
         email = 'john@example.com'
         password = 'secret'
         resp = c.post('/api/auth', json={'login': email, 'password': password})
-        assert verify_token(email, resp.get_json()['token'])
+        json_data = resp.get_json()
+        assert verify_token(email, json_data['token'])
 
 Note that if the ``json`` argument is provided then the test client will put
 JSON-serialized data in the request body, and also set the
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -362,7 +362,7 @@ Testing JSON APIs
 
 Flask has great support for JSON, and is a popular choice for building REST
 APIs. Testing both JSON requests and responses using the test client is very
-convenient:
+convenient::
 
     from flask import jsonify
 
@@ -371,12 +371,14 @@ convenient:
         json_data = request.get_json()
         email = json_data['email']
         password = json_data['password']
+
         return jsonify(token=generate_token(email, password))
 
     with app.test_client() as c:
         email = 'john@example.com'
         password = 'secret'
         resp = c.post('/api/auth', json={'login': email, 'password': password})
+
         json_data = resp.get_json()
         assert verify_token(email, json_data['token'])
 
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -360,24 +360,24 @@ Testing JSON APIs
 
 .. versionadded:: 1.0
 
-Flask has a great support for JSON, and is a popular choice for building REST
+Flask has great support for JSON, and is a popular choice for building REST
 APIs. Testing both JSON requests and responses using the test client is very
 convenient:
 
-    from flask import json, jsonify
+    from flask import jsonify
 
     @app.route('/api/auth')
     def auth():
-        email = request.json['email']
-        password = request.json['password']
+        email = request.get_json()['email']
+        password = request.get_json()['password']
         return jsonify(token=generate_token(email, password))
 
     with app.test_client() as c:
         email = 'john@example.com'
         password = 'secret'
-        resp = c.post('/api/auth', json={'email': email, 'password': password})
-        assert verify_token(email, resp.json['token'])
+        resp = c.post('/api/auth', json={'login': email, 'password': password})
+        assert verify_token(email, resp.get_json()['token'])
 
 Note that if the ``json`` argument is provided then the test client will put
-the JSON-serialized object in the request body, and also set the
-``Content-Type: application/json`` header.
+JSON-serialized data in the request body, and also set the
+``Content-Type: application/json`` HTTP header.
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -353,3 +353,11 @@ independently of the session backend used::
 Note that in this case you have to use the ``sess`` object instead of the
 :data:`flask.session` proxy.  The object however itself will provide the
 same interface.
+
+
+Testing JSON APIs
+-----------------
+
+.. versionadded:: 1.0
+
+Flask has comprehensive
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -360,4 +360,24 @@ Testing JSON APIs
 
 .. versionadded:: 1.0
 
-Flask has comprehensive
+Flask has a great support for JSON, and is a popular choice for building REST
+APIs. Testing both JSON requests and responses using the test client is very
+convenient:
+
+    from flask import json, jsonify
+
+    @app.route('/api/auth')
+    def auth():
+        email = request.json['email']
+        password = request.json['password']
+        return jsonify(token=generate_token(email, password))
+
+    with app.test_client() as c:
+        email = 'john@example.com'
+        password = 'secret'
+        resp = c.post('/api/auth', json={'email': email, 'password': password})
+        assert verify_token(email, resp.json['token'])
+
+Note that if the ``json`` argument is provided then the test client will put
+the JSON-serialized object in the request body, and also set the
+``Content-Type: application/json`` header.
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -37,7 +37,7 @@ def make_test_environ_builder(app, path='/', base_url=None, json=None, *args, **
 
     if json is not None:
         if 'data' in kwargs:
-            raise RuntimeError('Client cannot provide both `json` and `data`')
+            raise ValueError('Client cannot provide both `json` and `data`')
         kwargs['data'] = json_dumps(json)
 
         # Only set Content-Type when not explicitly provided
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -35,16 +35,14 @@ def make_test_environ_builder(app, path='/', base_url=None, json=None, *args, **
             if url.query:
                 path += '?' + url.query
 
-    if json:
+    if json is not None:
         if 'data' in kwargs:
             raise RuntimeError('Client cannot provide both `json` and `data`')
         kwargs['data'] = json_dumps(json)
 
         # Only set Content-Type when not explicitly provided
-        if 'Content-Type' not in kwargs.get('headers', {}):
-            new_headers = kwargs.get('headers', {}).copy()
-            new_headers['Content-Type'] = 'application/json'
-            kwargs['headers'] = new_headers
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'application/json'
 
     return EnvironBuilder(path, base_url, *args, **kwargs)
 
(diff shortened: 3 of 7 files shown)


adambyrtek:
Not ready for review yet.

flask/wrappers.py
```diff
@@ -18,14 +18,94 @@
 _missing = object()
 
 
-def _get_data(req, cache):
+def _get_data(req):
     getter = getattr(req, 'get_data', None)
     if getter is not None:
-        return getter(cache=cache)
```
adambyrtek: The flag was unnecessary here for two reasons:
1. Werkzeug's `request.get_data()` has `cache=True` as a default.
2. `JSONMixin.get_json()` with caching enabled shouldn't call `_get_data()` more than once anyway.

Removed since the response object doesn't support explicit caching.

untitaker: `get_json(cache=False)` shouldn't store any data on the response object and consume any underlying buffers IIRC

adambyrtek: Makes sense... please take a look at the alternative approach I've just pushed.

adambyrtek:
@untitaker I suggest to close #1408 and move the conversation to this pull request, since it incorporates both changes including documentation.

untitaker:
This is quite a bunch to review. I'll take a look at it sometime, but I don't have the time ATM and I'd apprechiate if somebody else looked over it too.

docs/testing.rst
```diff
@@ -351,3 +351,31 @@ independently of the session backend used::
...
+    def auth():
+        email = request.json['email']
+        password = request.json['password']
+        return jsonify(token=generate_token(email, password))
+
+    with app.test_client() as c:
+        email = 'john@example.com'
+        password = 'secret'
+        resp = c.post('/api/auth', json={'email': email, 'password': password})
+        assert verify_token(email, resp.json['token'])
```
ThiefMaster: I don't think example code in the docs should be using a deprecated property.

flask/wrappers.py
```diff
@@ -18,14 +18,93 @@
...
+        if cache:
+            self._cached_json = rv
+        return rv
+
+    def on_json_loading_failed(self, e):
+        """Called if decoding of the JSON data failed.  The return value of
+        this method is used by :meth:`get_json` when an error occurred.  The
+        default implementation just raises a :class:`BadRequest` exception.
+        """
+        raise BadRequest()
```
ThiefMaster: What happened to the more verbose error in debug mode?

```
         ctx = _request_ctx_stack.top
         if ctx is not None and ctx.app.config.get('DEBUG', False):
             raise BadRequest('Failed to decode JSON object: {0}'.format(e))
         raise BadRequest()
```

untitaker:
Thanks @ThiefMaster

flask/wrappers.py
```diff
@@ -95,73 +174,6 @@ def blueprint(self):
...
-                rv = None
-            else:
-                rv = self.on_json_loading_failed(e)
-        if cache:
-            self._cached_json = rv
-        return rv
-
     def on_json_loading_failed(self, e):
         """Called if decoding of the JSON data failed.  The return value of
         this method is used by :meth:`get_json` when an error occurred.  The
```
untitaker: Can this method be moved into the mixin?

flask/wrappers.py
```diff
@@ -18,14 +18,93 @@
...
+        get invoked.
+
+        This is deprecated, the :meth:`get_json` method should be used instead.
+        """
+        from warnings import warn
+        warn(DeprecationWarning('json is deprecated.  '
+                                'Use get_json() instead.'), stacklevel=2)
+        return self.get_json()
+
+    def _get_data_for_json(req, cache):
```
ThiefMaster: First argument should be named `self`

flask/wrappers.py
```diff
@@ -201,3 +213,10 @@ class Response(ResponseBase):
     set :attr:`~flask.Flask.response_class` to your subclass.
     """
     default_mimetype = 'text/html'
+
+    def _get_data_for_json(req, cache):
```
ThiefMaster: First argument should be named `self`

CHANGES
```diff
@@ -8,6 +8,9 @@ Version 1.0
 
 (release date to be announced, codename to be selected)
 
+- Added `json` keyword argument to :meth:`flask.testing.FlaskClient.open`
+  (and related ``get``, ``post``, etc.), which makes it more convenient to
+  send JSON requests from the test client.
```
untitaker: Something's missing here, see PR title.

docs/testing.rst
adambyrtek: Done

flask/wrappers.py
adambyrtek: This is a default mixin implementation that works for both request/response. The more verbose error relies on request context, so it was moved to `Request.on_json_loading_falied` which overloads this one.

The diff can be a bit misleading so feel free to look at the full source.

flask/wrappers.py
adambyrtek: Missed this during refactoring. Fixed.

flask/wrappers.py
adambyrtek: Done

CHANGES
adambyrtek: Added a second entry.

flask/wrappers.py
untitaker: One could use a single method that first checks whether a context is available before trying to access current_app.

flask/wrappers.py
untitaker: I've also noticed that it checks whether the request context is available, but then only accesses the current app. It should only check for the app context.

adambyrtek: You're really quick, actually I just had the same idea, see the last commit :)

adambyrtek:
@untitaker @ThiefMaster Thanks a lot for the review, I've gone through all your comments, so please take one more look.

One thing that's not clear to me is what is the reason for deprecating the `json` property (the changelog doesn't help much)? For me it's much more elegant to write `request.json['foo']` instead of `request.get_json()[`foo`]`, and properties are well-established in the Python community.

I understand that there is some computation and caching hidden under the hood, but most people working with JSON APIs won't have to worry about that, and the rest can use the getters directly if they want to.

ThiefMaster:
I think you are supposed to use something like `data = request.get_json()` and then use `data` in your code. But I'm not a big fan of the deprecation either. It feels nicer to use `request.json` just like `request.form` etc.

untitaker:
I think the general idea was to make resource management more explicit.

IIRC there was a Werkzeug or Flask issue where somebody expected to be able to read from the `data` attribute after already having accessed `json`, but doing so would have implicitly copied the request payload all over the place. Now you have to do it explicitly.

ThiefMaster:
But it's the same for e.g. `request.form` - so I'd prefer not to have write uglier code for the same of someone who can't read the documentation ;)

untitaker:
I'm not sure where I stand either, but I think `form` being still there isn't a deliberate inconsistency.

adambyrtek:
I use `request.json` all over the place in my API, and this is a backwards-incompatible change that would make the life of my team harder, so I'm not sure whether the benefit justifies the pain. I believe it should be the person having unusual use case having to go an extra mile (maybe with a more meaningful error message).

untitaker:
Regarding API breakage in general: To be fair, you have a lot of time to upgrade as Flask releases are really rare.

ThiefMaster:
I think changing `request.form` to a method would be the kind of API breakage that pisses off people, even if it's in a major version bump ;)

untitaker:
Hmm, I think we should move this discussion somewhere else, perhaps a new issue.

I'll keep this PR unmerged and will properly review it tomorrow. I'd like more eyes on it too.

adambyrtek:
Sure, let's track this separately, and I'm happy to stay involved. You know where I stand, even when this becomes deprecated I will still use `request.json` on my project because I believe it leads to cleaner code, and I cannot justify the refactoring.

adambyrtek:
Just to clarify, this should be ready for the final review regardless of the result of the discussion.

adambyrtek:
@untitaker Gentle ping.

untitaker:
@adambyrtek Could you check whether the new attrs for response show up correctly in the docs?

adambyrtek:
@untitaker You mean the API docs? Now is does, it never showed up for `Request` by the way.

adambyrtek:
@untitaker What do you think?

untitaker:
Sorry for the late response: I want @mitsuhiko to look over this, one last time. The implementation looks good, I don't know about the added API though.

untitaker:
Nvm I'll just take a proper look later.

adambyrtek:
@untitaker I'm leaving on vacation tomorrow, so I won't be able to reply to any comments until I come back in two weeks.

untitaker:
@mitsuhiko Please just take a quick look over this if it makes sense conceptually -- implementation-wise it's fine IMO.

adambyrtek:
Any news when it comes to this pull request?

untitaker:
@adambyrtek Please ping @mitsuhiko in IRC, and if he's fine with it, let's merge it.

Just now I've tried to reach him w/o success.

sfermigier:
Trying again to ping @mitsuhiko about this pull request.

adambyrtek:
@sfermigier @untitaker @mitsuhiko I mailed him some time ago as well... I'm at EuroPython now so if any of you is there we could meet and talk :)

untitaker:
@adambyrtek Will merge. Could you rebase?

adambyrtek:
@untitaker Rebased and ready to merge. Just take a look at the final cleanup commit https://github.com/mitsuhiko/flask/commit/26216cf627634d6bbf93bb025c1bd6c0349ffcaf

ThiefMaster:
Hm, what about squashing the commits? There seem to be quite a few intermediate steps right now which are not really relevant in the end. Considering that it's one feature and the diff if not that big I'd even go for one single commit.

adambyrtek:
@untitaker @ThiefMaster What's the project policy? I prefer to keep the history on my projects as it doesn't cost anything and might be useful when debugging later, but I can squash if you prefer.

ThiefMaster:
I don't think there's really a policy (also I'm not a maintainer, just an active user who contributes from time to time). I guess it's mostly a matter of preference. The only important thing is not to clutter history with merge commits that could have been simple rebases (but that's not the case here anyway).

untitaker:
Yeah, there's no policy. You may modify history however you want until it hits master.

adambyrtek:
@untitaker I prefer to keep the history for the record (unless it contains something obviously unnecessary or erroneous) so feel free to merge when you're ready to do so.

untitaker:
I pinged @mitsuhiko about this again in IRC.

adambyrtek:
Just to be sure you remember, you said "Will merge. Could you rebase?" some time ago (see https://github.com/mitsuhiko/flask/pull/1416#issuecomment-123285187). Anyway, I'm happy to wait for the blessing.

While we're at this, it would be great to get @mitsuhiko opinion on https://github.com/mitsuhiko/flask/issues/1421

untitaker:
Argh. Sorry. I don't know what to do.

I can't reach him.

adambyrtek:
@untitaker Maybe you should just make a judgement call? I don't think this PR introduces anything revolutionary. It simply adds JSON capabilities to the response object, mostly for testing purposes.

untitaker:
Yes, and @mitsuhiko has been known to oppose decisions that seemed straightforward to anybody else.

Apparently that means that no real progress can be made, but it's not my decision to judge whether that's a good outcome -- it's still Armin's project after all.

adambyrtek:
Fair enough, I didn't know that he is the sole maintainer.

flask/testing.py
```diff
@@ -20,7 +21,7 @@
     from urlparse import urlsplit as url_parse
 
 
-def make_test_environ_builder(app, path='/', base_url=None, *args, **kwargs):
+def make_test_environ_builder(app, path='/', base_url=None, json=None, *args, **kwargs):
```
johnnyg: I realise this comment is a bit late but wouldn't it be better to not make json a keyword argument and instead check for it inside kwargs? This would mean that passing `json=None` would send a valid json request with data as 'null' instead of not doing anything. It also has the added benefit that you don't need to rename your import of `json`.

adambyrtek:
@mitsuhiko I've seen your recent comment on another JSON-related issue. Would you mind taking a look at this one?

flask/testing.py
gregorynicholas: i agree, but the method signature becomes less explicit, and relies pretty heavily on implicit knowledge of the workings .

besides  look up at url_parse

flask/wrappers.py
```diff
@@ -13,19 +13,108 @@
...
+
+        The :meth:`get_json` method should be used instead.
+        """
+        from warnings import warn
+        warn(DeprecationWarning(
+            'json is deprecated.  Use get_json() instead.'), stacklevel=2)
+        return self.get_json()
+
+    def _get_data_for_json(self, cache):
+        getter = getattr(self, 'get_data', None)
```
gregorynicholas: just use hasattr() ..?

flask/wrappers.py
```diff
@@ -201,3 +206,10 @@ class Response(ResponseBase):
     set :attr:`~flask.Flask.response_class` to your subclass.
     """
     default_mimetype = 'text/html'
+
+    def _get_data_for_json(self, cache):
```
untitaker: This method is already in JSONMixin.

flask/wrappers.py
untitaker: @gregorynicholas Would you really want to try to call `get_data` if it's `None`?

flask/testing.py
untitaker: I agree with @johnnyg.

adambyrtek:
Thanks for the comments. I'm happy to do the changes but only if this has a chance of moving forward.

untitaker:
That is totally understandable, and thank you for all your work. I personally
want this to land in 0.11.

mitsuhiko:
I think it generally looks reasonable. I don't have any arguments for not merging it.

untitaker:
Excellent. @adambyrtek do you have time?

untitaker:
@adambyrtek So this didn't make it in time. Are you still interested in updating this?

DoWhileGeek:
pallets/werkzeug#948 addresses the Response side of this issue. It needs a review if anyone wants to jump on it.

untitaker:
I'm picking this up at https://github.com/pallets/flask/pull/1984

PR #1422, opened 2015-04-06
Allow custom jinja environments

ThiefMaster:
This is useful e.g. when using the new Jinja Environment
attributes added in mitsuhiko/jinja2#404

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -157,6 +157,11 @@ class Flask(_PackageBoundObject):
     #: :class:`~flask.Response` for more information.
     response_class = Response
 
+    #: The class that is used for the Jinja environment.
+    #:
+    #: .. versionadded:: 0.11
+    jinja_env_class = Environment
+
     #: The class that is used for the :data:`~flask.g` instance.
     #:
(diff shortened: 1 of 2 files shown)


flask/app.py
```diff
@@ -157,6 +157,11 @@ class Flask(_PackageBoundObject):
     #: :class:`~flask.Response` for more information.
     response_class = Response
 
+    #: The class that is used for the Jinja environment.
+    #:
+    #: .. versionadded:: 0.11
```
ThiefMaster: Not sure if this should be 0.11 or 1.0 - currently both versions are used, with 1.0 being more common but the current package version in `flask/__init__.py` being 0.11

untitaker: I always assumed it is 1.0 but I'm not sure whether we should release an intermediate version.

ThiefMaster: Hm I guess for now the different versions don't really matter as long as it's cleaned up before whatever the next release's version number will be ;)

untitaker:
Thanks @ThiefMaster!

PR #1429, opened 2015-04-11
Added docs for the error handler rework

flying-sheep:
Documenting the changes in #1291

### DIFF
--- a/docs/errorhandling.rst
+++ b/docs/errorhandling.rst
@@ -1,7 +1,7 @@
 .. _application-errors:
 
-Logging Application Errors
-==========================
+Application Errors
+==================
 
 .. versionadded:: 0.3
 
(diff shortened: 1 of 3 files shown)


docs/upgrading.rst
```diff
@@ -29,6 +29,21 @@ applications.  Instead the new ``LOGGER_HANDLER_POLICY`` configuration can
...
+:meth:`~flask.Flask.register_error_handler`, respectively.
+Now the inheritance hierarchy takes precedence and handlers for more
+specific exception classes are executed instead of more general ones.
+See :ref:`error-handlers` for specifics.
+
+.. note::
+
+    There used to be a logic error allowing you to register handlers
+    only for exception *instances*. This was unintended and plain wrong,
+    and therefore was replaced with the intended behavior of registering
```
untitaker: We should make it clear that Flask raises an error (the current text reads like registering an exception instance is a noop).

docs/errorhandling.rst
```diff
@@ -28,6 +28,59 @@ exception to the :attr:`~flask.Flask.logger`.
...
 with errors.
 
+.. _error-handlers:
+
+Error handlers
+--------------
+
+.. versionadded:: 1.0
+
+Some errors are unavoidable because not only you make them.  Users might
```
untitaker: We now have two general introductions: "Applications fail, servers fail..." and this one.

flying-sheep: i thought that coming here directly from the quickstart would be good.

also this is the introduction for the “i want to show an error page” use case. the other one generally for errors.

which parts do you think should be moved or removed?

docs/upgrading.rst
flying-sheep: ok

docs/errorhandling.rst
untitaker: I'm not sure. I get your point about being directly linked from the quickstart, but if you read the whole page from top to bottom, the train of thought is pretty confusing.

docs/upgrading.rst
```diff
@@ -29,6 +29,23 @@ applications.  Instead the new ``LOGGER_HANDLER_POLICY`` configuration can
...
+    handlers only using exception classes and HTTP error codes.
+    
+    Trying to register a handler on an instance now raises :exc:`ValueError`
```
untitaker: Missing dot at the end of the sentence :)

docs/errorhandling.rst
flying-sheep: hmm, my intention was:

1. general introduction: things go wrong, this is what can happen, now some ways to :sunglasses: deal with it
2. in some cases you might want to show error pages, this is when and how: …
3. if the default logging isn’t enough, do this: …
4. …

untitaker: Right, but I find it unnecessary to introduce the reader to the idea that apps may throw errors *twice*. I find it barely necessary the first time, and regardless whether it's a user or server error, you probably want to register errorhandler for either, so there's no real distinction regarding usecases.

docs/upgrading.rst
flying-sheep: :ballot_box_with_check:

docs/errorhandling.rst
flying-sheep: true, especially the point about wanting to install them anyway (especially since that’s why i put it as first section)

untitaker: Maybe we should just link to the whole document instead of a particular subsection?

docs/errorhandling.rst
```diff
@@ -28,6 +28,54 @@ exception to the :attr:`~flask.Flask.logger`.
...
+Error handlers
+--------------
+
+.. versionadded:: 1.0
+
+You might want to show error pages to the user when an error occurs.
+This can be done by registering error handlers.
+
+Error handlers are normal :ref:`views` but instead of being registered for
+routes they are registered exceptions that are rised while trying to do
```
untitaker: "registered for"

docs/errorhandling.rst
```diff
@@ -28,6 +28,54 @@ exception to the :attr:`~flask.Flask.logger`.
...
 with errors.
 
+.. _error-handlers:
+
+Error handlers
+--------------
+
+.. versionadded:: 1.0
+
+You might want to show error pages to the user when an error occurs.
```
untitaker: "show custom error pages", since some sort of error page is always shown

docs/errorhandling.rst
```diff
@@ -28,6 +28,54 @@ exception to the :attr:`~flask.Flask.logger`.
...
+This can be done by registering error handlers.
+
+Error handlers are normal :ref:`views` but instead of being registered for
+routes they are registered exceptions that are rised while trying to do
+something else.
+
+Registering
+```````````
+
+Register error handlers using :meth:`~flask.Flask.errorhandler` or
```
untitaker: Do these render correctly? I think you need `:py:meth:`, `:py:exc:` etc.

flying-sheep: they do, i tested. the whole thing is rendered in python context apparently, and i see plain `:meth:` and `:exc:` all over the place

untitaker:
Awesome!

flying-sheep:
:blush:

Issue #1440, opened 2015-04-18
Change first argument in a Class methoc to 'self'


andyxning:
I recently run into the **app.logger** usage. However, when i read the source code in flask **v0.10.1** , i find that it uses `x` instead of `self` as the first method argument in class [**DebugLogger**](https://github.com/mitsuhiko/flask/blob/0.10.1/flask/logging.py#L26) and [**DebugHandler**](https://github.com/mitsuhiko/flask/blob/0.10.1/flask/logging.py#L32).

Can we just update this calling convention.

grahamlyons:
Sounds perfectly reasonable. Submit a pull request?

andyxning:
Ok, i will make it later.



++++++
Ning Xie


2015-04-19 2:06 GMT+08:00 Graham Lyons <notifications@github.com>:

PR #1441, opened 2015-04-19
Fix pep8 with logging

andyxning:
This should make the logging module more pep8 compatible. This PR fix problem issued in #1440 .

I made this PR to update first argument of class method in logging module from `x` to `self` in **master** branch. However, i can not make sure if i should at the same time make a PR to also clean the `logging` module in **0.10-maintenance**.

BTW, is the maintenance branch read-only(i checked out that last commit in **0.10-maintenance** is about years ago)? If so, what is the difference between a tag and this maintenance branch.

### DIFF
--- a/flask/logging.py
+++ b/flask/logging.py
@@ -57,10 +57,10 @@ def create_logger(app):
     Logger = getLoggerClass()
 
     class DebugLogger(Logger):
-        def getEffectiveLevel(x):
-            if x.level == 0 and app.debug:
+        def getEffectiveLevel(self):
+            if self.level == 0 and app.debug:
                 return DEBUG
-            return Logger.getEffectiveLevel(x)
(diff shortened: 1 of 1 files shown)


untitaker:
Please remove that merge commit with `git rebasew -i master`.

andyxning:
Sorry for not skilled git knowledge. I think i make things more bad than beginning. I think i should close this PR and issue another one with correct commit.

ThiefMaster:
No, simply force-push to the same branch. Pull requests are always between branches.

untitaker:
Ok, run that command again and force-push to the same branch... don't pull to resolve conflicts.

andyxning:
@ThiefMaster  Thanks very much for the advice and it finally worked correctly. A new skill has been leaned. :)

untitaker:
Great, thanks!

PR #1446, opened 2015-04-26
Built in routes command

kissgyorgy:
I don't want to spam the builtin command namespace (anyway, why not? Django have a lot of builtin commands and works pretty well) but this is the first command I always implement when I start using flask, and my collegues were doing this also.
It shows all the endpoints registered for the app. Routes can be ordered by rules, endpoints and methods.
This is a WIP, because I did not wrote tests for it yet and did not tried with 2.6 and 2.7.
I would like to have some feedback on this and I can work out the details.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -476,8 +476,11 @@ def routes_command(order_by):
     sorted_rules = sorted(app.url_map.iter_rules(), key=order_key)
 
     max_rule = max(len(rule.rule) for rule in sorted_rules)
+    max_rule = max(max_rule, len('Route'))
     max_ep = max(len(rule.endpoint) for rule in sorted_rules)
+    max_ep = max(max_ep, len('Endpoint'))
     max_meth = max(len(', '.join(rule.methods)) for rule in sorted_rules)
+    max_meth = max(max_meth, len('Methods'))
 
     columnformat = '{:<%s}  {:<%s}  {:<%s}' % (max_rule, max_ep, max_meth)
     click.echo(columnformat.format('Route', 'Endpoint', 'Methods'))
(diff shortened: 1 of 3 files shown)


untitaker:
>anyway, why not? Django have a lot of builtin commands and works pretty well

Django has a lot of stuff that works surprisingly well, but that's a bad argument.

I like the command, but what exactly are you using it for?

kissgyorgy:
- If you register a bunch of blueprints on the app, it's harder to track down the exact URLs.
- You don't always know all the routes, e.g. third party blueprints, extensions.
- When I worked in a team, it was really handy to see what new views the others guys were writing.
- I subclassed `Flask`, so I don't need to write `app.route()` all the time, (e.g. I just can give it a simple html files, and it automatically makes a view), this commands still shows everything.

mattupstate:
```
print(app.url_map)
```

:smirk_cat:

wshayes:
+1 I agree - one of the first things I add as well - just nice to get an overview since the routes are spread all over

kissgyorgy:
I rebased it to master

kissgyorgy:
I rebased it to master

jirikuncar:
What about creating a package (e.g. Flask-Routes) that would register such command and you can just add an requirement to your project?

```
entry_points={
        'flask.commands': [
            'routes = flask_routes:routes',
        ],
    },
```

kissgyorgy:
> What about creating a package (e.g. Flask-Routes) that would register such command and you can just add an requirement to your project?

You would rather "left-paddify" this simple command?

davidism:
Thanks for starting this. I liked the general idea but decided to rework it in #2259.

PR #1456, opened 2015-05-10
Add X-Forwarded-Proto header to proxy setup example docs

jonafato:
The `ProxyFix` middleware provided by Werkzeug uses this header for
returning accurate values from request.is_secure and request.scheme.
Without adding this header, Flask won't properly detect when it is being
served over HTTPS and will fail to generate proper external links and
cause certain extensions (e.g. Flask-OAuthlib) to function improperly.
Adding this header to the example setup should reduce issues encountered
by developers when following this guide.

Credit where due: I tend to find this information when I need it in a [blog post](http://cramer.io/2013/06/27/serving-python-web-applications/) by David Cramer.


untitaker:
Excellent, thank you.

PR #1462, opened 2015-05-19
fix doc string

moonshadow:
the doc string for  `_find_error_handler` is incomplete.


untitaker:
I think this phrasing is kind of repetitive, although I know that the current one is not explicitly saying that it falls back. A fix for this could be to replace the last sentence with "Otherwise falls back to ...".

Issue #1466, opened 2015-05-21
try_trigger_before_first_request_functions in test cases


nitori:
Hi,

I'm currently testing a feature (knitori@f2ae742d163300a22b7b03ca2ac60810ed1de59b) that required to move the registration of the static url from `__init__` to a method, and register that method using `before_first_request` (knitori@6bc77363ed9ee1a9ba00d3536f0479ab0967a8bd).

It works pretty well with a normal application, but it fails one test (`tests/test_basic.py::test_static_files FAILED`). And I realized, that the registered `before_first_request` functions are never actually called with the test client `testing.FlaskClient`. I fixed that locally by adding `self.application.try_trigger_before_first_request_functions()` right before the return statement of `testing.FlaskClient.open` (or should it even go to `Client.open`?).

I'm wondering if there is a rationale behind not calling the registered functions? Or has there just never been a case where it's actually required?

**Edit**

I realized this doesn't work. If the first request is to a static URL, it will return 404, because the RequestContext was created before the static route was registered. If anything, I probably need another even like `before_first_request_context` or something like that. Maybe called inside of `Flask.request_context`.

I'm sorry, it seems this turned into a completely different issue.

davidism:
As soon as the application is handling any part of a request, it should no longer be modifying itself.

PR #1468, opened 2015-05-22
Add before_first_request_context signal.

knitori:
I essentially cloned what `before_first_request` does.

Intention is to get the registration of the static route out of `__init__`, and register it instead right before the first request. (for a feature I mentioned in issue #1466)

~~Another possibility would be to pass `environ` along, to make it possible to modify it (for whatever reasons) before it get's passed to `RequestContext` - if desired.~~ well, not actually reasonable for before_*first*_request_context.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -403,6 +403,11 @@ class Flask(_PackageBoundObject):
         #: .. versionadded:: 0.8
         self.before_first_request_funcs = []
 
+        #: A list of functions that should be called before the first request
+        #: context is created.  To register a function here, use the
+        #: :meth:`before_first_request_context` decorator.
+        self.before_first_request_context_funcs = []
+
         #: A dictionary with lists of functions that should be called after
         #: each request.  The key of the dictionary is the name of the blueprint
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Couldn't this cause problems in a multithreaded environment?

knitori:
If so, then wouldn't `before_first_request` cause them as well?

I basically just duplicated what `before_first_request` does, including [the lock](https://github.com/mitsuhiko/flask/pull/1468/files#diff-c3c1b62367ae818e58e148e2e1840b39R1940), that's being used.

untitaker:
Couldn't you subclass the Flask app and override the `request_context` method for your purpose? It seems easy enough without introducing more API.

BTW I think you can reuse `_got_first_request` instead of having your own variable.

knitori:
Okay, I guess I'll do that instead. Haven't even thought of just subclassing.

untitaker:
Happy that this covers your usecase, I personally think the API of Flask needs to rather shrink than grow. :)

PR #1472, opened 2015-05-25
Add a method to register multiple blueprints at once.

sinwoobang:
Hi, I added a method to register multiple blueprints.

#### Example
```python
app.register_blueprints([a_blueprint, b_blueprint, c_blueprint])
```

Thanks.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -931,6 +931,10 @@ class Flask(_PackageBoundObject):
             first_registration = True
         blueprint.register(self, options, first_registration)
 
+    def register_blueprints(self, blueprints):
+        for blueprint in blueprints:
+            self.register_blueprint(blueprint)
+
     def iter_blueprints(self):
         """Iterates over all blueprints by the order they were registered.
 


ThiefMaster:
What's the benefit of having this method? It saves you exactly one line of code in your application...

sinwoobang:
@ThiefMaster 
Yes, you're right.
It just reduce the small amount of code.

But, I wanted to reduce the repetition of register_blueprint method even a little.
The larger the project, I had to be often used to duplicate the function.

And, What is the option you are talking about?
If you have said the url_prefix option, It can be possible to be replace with Blueprint class option.

Sorry, I'm not fluent in English.
I tried to politely as possible.
Please comment me more if you have a question.

Thanks.

untitaker:
Sorry, I don't think this change should be included in Flask. It might seem "convenient" at first, but all it really does is enlargen the API and increase the amount of things to learn when looking at Flask code without knowing Flask itself.

>Sorry, I'm not fluent in English.
>I tried to politely as possible.

You did not come across as impolite in any way!

PR #1476, opened 2015-05-30
Change documentation entries

igncp:
Specially misleading the entry about the `static` directory, as it can have another name and still be valid.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -813,8 +813,9 @@ class _PackageBoundObject(object):
 
     @property
     def has_static_folder(self):
-        """This is ``True`` if the package bound object's container has a
-        folder named ``'static'``.
+        """This is ``True`` if the package bound object's container instance has
+        its attribute ``static_folder`` (defaults to ``'static'``) different
+        than ``None``.
 
         .. versionadded:: 0.5
         """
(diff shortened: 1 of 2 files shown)


untitaker:
In addition to being more correct, I think this version is also more roundabout than the previous version, the property name is more to the point. Also I don't know why you replaced container with container instance?

igncp:
Well, about adding `instance`, I probably missunderstood it, I was refering to the inherinting class instance (e.g. app, instance of Flask, or an instance from Blueprint). Maybe replacing "package bound object's container" by "_PackageBoundObject class instance", but it seems a more abstract explanation than before. Any suggestions? Should I change / close this?

untitaker:
I thought of "container" as containing directory. You can change this PR by committing to the same branch.

davidism:
`_PackageBoundObject.static_folder` does not default to `'static'`, it defaults to `None`.  The `Flask` class changes this, and it's already mentioned in the [docs there][1].  I think this sounds more natural:

~~~
Returns ``True`` if :attr:`static_folder` is not ``None``.
~~~

[1]: https://github.com/mitsuhiko/flask/blob/0.10-maintenance/flask/app.py#L125-L127

igncp:
@davidism You are right, this default only applies for Flask, thanks for the catch.

Well, I would update the PR with @davidism suggestion, but it seems that the code is self-explanatory. Would you put @davidism comment to be readable in the docs? Remove it?

untitaker:
I think @davidism's phrasing is just repeating the function's body in words. I've pushed a commit that probably is a good compromise.

PR #1479, opened 2015-06-02
Implement config.from_envjson (with testing)

tonyseek:
I added a new method `from_envjson` into `Config` class. It allows people to configure their applications with environment variables in JSON syntax, without any config file.

Storing config in environment variables is a recommended practice of [12-Factor App](http://12factor.net/config). It is supported by many development and deployment utilities, including systemd(`EnvironmentFile` field in service file), [honcho](https://github.com/nickstenning/honcho), [autoenv](https://github.com/kennethreitz/autoenv) and more.

Please review it. Thanks.

### DIFF
--- a/flask/config.py
+++ b/flask/config.py
@@ -213,6 +213,56 @@ class Config(dict):
                     self[key] = value
         return True
 
+    def from_envjson(self, prefix, silent=False):
+        """Updates the config from JSON-parsed environment variables.
+
+        For example::
+
+            $ export FLASKR_DEBUG='true'
+            $ export FLASKR_SECRET_KEY='"ba92ffd4d9582827cb8f93c4b48f60e3"'
+            $ export FLASKR_MAX_CONTENT_LENGTH='4096'
+            $ cat <<EOF > foo.py
+            > from flask import Flask
+            >
+            > app = Flask(__name__)
+            > app.config.from_envjson('FLASKR')
+            >
+            > print('DEBUG:%r' % app.debug)
+            > print('SECRET_KEY:%r' app.secret_key)
+            > print('MAX_CONTENT_LENGTH:%r' % app.config['MAX_CONTENT_LENGTH'])
+            > EOF
+            $ python foo.py
+            DEBUG:True
+            SECRET_KEY:'ba92ffd4d9582827cb8f93c4b48f60e3'
+            MAX_CONTENT_LENGTH:4096
+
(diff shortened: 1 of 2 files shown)


untitaker:
This is interesting and definetly worth considering for 1.0. However, at the same time I think Flask has too many `from_` methods already -- maybe it makes sense nuking some others too?

cc @DasIch @mitsuhiko

Bekt:
What's wrong with just getting these values in the config file?

```py
DEBUG = os.environ.get('FLASKR_SECRET_KEY', False)

...

app.config.from_object('configs')
```

untitaker:
@Bekt Of course it's a convenience method (like all `from_*` methods). We add the ones we actually want to encourage.

Bekt:
@untitaker I agree with you. I can think of few other ways too that are convenient sometimes (yaml), but I think having too many `from_*` methods is not Pythonic.

chaoflow:
While I do see the use for configuration via environment variables I have doubts about the json serialization.

I wonder whether typing (part of) app.config similar to click options would be beneficial and then support:
- http://click.pocoo.org/5/options/#values-from-environment-variables
- http://click.pocoo.org/5/options/#multiple-values-from-environment-values
- http://click.pocoo.org/5/commands/#overriding-defaults

An app's config would have default values, could be overridden by config file, provided by envionment variables and via cli parameters.

This might be in line with the new [flask cli](http://flask.pocoo.org/docs/dev/cli/#command-line-interface)

davidism:
While I understand that this can be useful to some projects, I'm not convinced it should be included with Flask.

The function name is confusing.  We already have `from_envvar`, which reads a Python file pointed to by an environment variable.  `from_envjson` first suggests that this is one var that will point at a JSON file, or perhaps one very long env var that contains the full JSON document.  Why is a lot of env vars a better use case than either of these other possibilities?

Before this and the `from_json` methods was added, there was only *one* official way (PEP 20!) to handle configuration: Python, which can be more expressive than any other source.  Either an object, an import, or a file from the root, the instance folder, or an env var, but all Python.  The more methods we add beyond that, the less clear it is to new developers what to choose.  It should be left to the developer to read in their config if they have a reason to use another source.

Ping @DasIch, I'd like to hear your rationale for the `from_json` and `from_mapping` methods, I couldn't find an issue discussing them.

davidism:
Closing due to inactivity and reasons above.

PR #1480, opened 2015-06-02
Handle empty deque on errorhandler lookup.

jmcarp:
After registering a custom errorhandler by exception class, raising any
unhandled exception in a view function swallows the error and instead
throws an `IndexError` on trying to look up the appropriate handler.
This patch avoids the uninformative `IndexError` and preserves the
original exception by looping until the deque of classes is empty, not
forever.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1421,7 +1421,7 @@ class Flask(_PackageBoundObject):
             # __mro__
             done = set()
 
-            while True:
+            while queue:
                 cls = queue.popleft()
                 if cls in done:
                     continue
--- a/tests/test_user_error_handler.py
+++ b/tests/test_user_error_handler.py
@@ -3,6 +3,35 @@ from werkzeug.exceptions import Forbidden, InternalServerError
 import flask
 
 
+def test_error_handler_no_match():
+    app = flask.Flask(__name__)
+
+    class CustomException(Exception):
+        pass
+
+    @app.errorhandler(CustomException)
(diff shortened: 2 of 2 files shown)


jmcarp:
Code example:

```python

from flask import Flask

app = Flask(__name__)

@app.errorhandler(TypeError)
def handle_typeerror(e):
    return 'got a typeerror!'

@app.route('/keyerror')
def keyerror():
    raise KeyError()

c = app.test_client()
c.get('/keyerror')
```

Since we've registered a handler for `TypeError` and not for `KeyError`, I would expect the `KeyError` to be thrown here. But instead we get an `IndexError` with no hint of the original error:

```
  File "/Users/jmcarp/code/flask/flask/app.py", line 1951, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/jmcarp/code/flask/flask/app.py", line 1605, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/jmcarp/code/flask/flask/app.py", line 1505, in handle_user_exception
    handler = self._find_error_handler(e)
  File "/Users/jmcarp/code/flask/flask/app.py", line 1440, in _find_error_handler
    .get(code))
  File "/Users/jmcarp/code/flask/flask/app.py", line 1425, in find_handler
    cls = queue.popleft()
IndexError: pop from an empty deque
```

With the proposed patch, the expected `KeyError` is thrown instead.

untitaker:
Excellent, thanks!

PR #1486, opened 2015-06-07
DRY globals definition

igncp:
Hi there,

Changed a little the globals.py file to avoid some code repetition, and also changed the order. The main change in my opinion is that now the `current_app` proxy would use `return getattr(top, 'app')` instead of `return top.app`. Maybe this has some side effects which I am missing.

### DIFF
--- a/flask/globals.py
+++ b/flask/globals.py
@@ -45,17 +45,14 @@ def _lookup_app_object(name):
     return getattr(top, name)
 
 
-def _find_app():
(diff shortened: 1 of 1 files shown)


untitaker:
Sorry, but I don't think this code is any more readable, and only adds another level of indirection.

ThiefMaster:
My first thought, too, when seeing the diff..

igncp:
Thanks for the explanation, I appreciate the advice.

PR #1489, opened 2015-06-08
Add kwarg to disable auto OPTIONS on add_url_rule

### LINKED ISSUE #1488: If you aren't using classes for `view_func`s, you can't tell Flask not to serve OPTIONS
We're using flask to serve mostly REST endpoints, and doing some setup at startup to create all the URL rules. Effectively, we're calling `app.add_url_rule()` for each method for each endpoint. This means that flask autogenerates an OPTIONS method and function for each ( https://github.com/mitsuhiko/flask/blob/master/flask/app.py#L1019 ), with no way to override.

Can `provide_automatic_options` be added as a kwarg to `add_url_rule`, in case nothing is specified on the `view_func`?

jrmccarthy:
Fixes #1488 : Adds support for a kwarg `provide_automatic_options` on `add_url_rule`, which
lets you turn off the automatic OPTIONS response on a per-URL basis even if
your view functions are functions, not classes (so you can't provide attrs
on them).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -70,9 +70,9 @@ Version 1.0
 - ``flask.g`` now has ``pop()`` and ``setdefault`` methods.
 - Turn on autoescape for ``flask.templating.render_template_string`` by default
   (pull request ``#1515``).
-- Added support for `provide_automatic_options` in `**kwargs` on
-  :meth:`add_url_rule` to turn off automatic OPTIONS when the `view_func`
-  argument is not a class (pull request ``#1489``).
+- Added support for `provide_automatic_options` in :meth:`add_url_rule` to
+  turn off automatic OPTIONS when the `view_func` argument is not a class
+  (pull request ``#1489``).
 
 Version 0.10.2
 --------------
--- a/CHANGES
+++ b/CHANGES
@@ -70,6 +70,9 @@ Version 1.0
 - ``flask.g`` now has ``pop()`` and ``setdefault`` methods.
 - Turn on autoescape for ``flask.templating.render_template_string`` by default
   (pull request ``#1515``).
+- Added support for `provide_automatic_options` in `**kwargs` on
+  :meth:`add_url_rule` to turn off automatic OPTIONS when the `view_func`
+  argument is not a class (pull request ``#1489``).
 
 Version 0.10.2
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Wouldn't it make more sense to give the kwarg priority over the attribute?

Oh, and please check why tests are failing in some environments.

jrmccarthy:
Sure, I can make the kwarg top priority. Doesn't matter to me.

I didn't realize older Werkzeug didn't have an `options` method; I've fixed that in the tests. Thanks for the feedback.

untitaker:
@jrmccarthy Thanks for this PR, could you add a changelog?

I'm not sure where this could be documented, but I think it should be, somehow.

jrmccarthy:
I've added a note in the changelog for this.

flask/app.py
```diff
@@ -1013,8 +1013,10 @@ def index():
 
         # starting with Flask 0.8 the view_func object can disable and
         # force-enable the automatic options handling.
-        provide_automatic_options = getattr(view_func,
-            'provide_automatic_options', None)
+        provide_automatic_options = options.pop(
+            'provide_automatic_options', getattr(view_func,
```
untitaker: I think it's better to add this as normal param to `add_url_rule` instead of popping from `options`, such that it can be documented in its docstring normally.

jrmccarthy:
Moved some things around, so now `provide_automatic_options` is an explicit kwarg, and is documented in the docstring, as requested.

CHANGES
```diff
@@ -70,6 +70,9 @@ Version 1.0
 - ``flask.g`` now has ``pop()`` and ``setdefault`` methods.
 - Turn on autoescape for ``flask.templating.render_template_string`` by default
   (pull request ``#1515``).
+- Added support for `provide_automatic_options` in :meth:`add_url_rule` to
+  turn off automatic OPTIONS when the `view_func` argument is not a class
```
jeffwidman: Because this is restructured text and not markdown, you'll need to use two backticks around `provide_automatic_options` and `view_func`

jeffwidman:
Please rebase and squash.

davidism:
Continued in #2249

PR #1491, opened 2015-06-10
Remove the word `trivially` here

nvie:
A lot of things are trivial, or debatably trivial, but this is not one of them :)


untitaker:
Trivial or not, that word doesn't help anything. Thanks!

PR #1502, opened 2015-06-17
Don't pass version to tox explicitly

untitaker:


ThiefMaster:
When doing that you probably want to remove the travis matrix. Otherwise you end up combining the travis matrix with the tox matrix.

untitaker:
Oh I still want a separate travis job for each tox build, but I'm currently figuring out if tox couldn't infer the version for us instead of having to parse `TRAVIS_PYTHON_VERSION`.

untitaker:
Seems fine to me?

PR #1509, opened 2015-06-20
Add pop and setdefault to AppCtxGlobals

ThiefMaster:
While easy to implement with a custom `app_ctx_globals_class` class, I believe these two methods are useful enough to add them to the default class.

Especially setdefault makes lots of sense if you e.g. cache something on the app context globals and thus want to do something like this:

```python
cache = g.setdefault('whatever_cache', {})
try:
    return cache['foo']
except KeyError:
    rv = cache['foo'] = get_foo()
    return rv
```


untitaker:
Right, thanks!

PR #1533, opened 2015-07-14
fix UnboundLocalError in handle_url_build_error

lobeck:
- caused by changes in the execution model of python 3 where the alias of an except clause is cleared on exit of the except

- tested with python 2.7.10 and 3.4.3

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1769,8 +1769,9 @@ class Flask(_PackageBoundObject):
                 rv = handler(error, endpoint, values)
                 if rv is not None:
                     return rv
-            except BuildError as error:
-                pass
+            except BuildError as e:
+                # make error available outside except block (py3)
+                error = e
 
         # At this point we want to reraise the exception.  If the error is
         # still the same one we can reraise it with the original traceback,
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1067,6 +1067,16 @@ def test_build_error_handler():
     app.url_build_error_handlers.append(handler)
     with app.test_request_context():
         assert flask.url_for('spam') == '/test_handler/'
+    # remove handler after test
+    app.url_build_error_handlers.pop()
+
+    # Test a custom handler which reraises the BuildError
+    def handler_raises_build_error(error, endpoint, values):
+        raise error
+    app.url_build_error_handlers.append(handler_raises_build_error)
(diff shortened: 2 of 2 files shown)


flask/app.py
```diff
@@ -1769,8 +1769,13 @@ def handle_url_build_error(self, error, endpoint, values):
                 rv = handler(error, endpoint, values)
                 if rv is not None:
                     return rv
-            except BuildError as error:
-                pass
+            except BuildError as e:
+                """We can't directly save the BuildError in error because the python3 execution model changed.
```
untitaker: Use a normal comment here.

flask/app.py
```diff
@@ -1769,8 +1769,13 @@ def handle_url_build_error(self, error, endpoint, values):
...
+                The `as` target is cleared after exiting the except clause and this causes an UnboundLocalError
+                see  https://docs.python.org/2/reference/compound_stmts.html#except
+                vs.  https://docs.python.org/3.3/reference/compound_stmts.html#except
+                """
+                error = e
```
ThiefMaster: You should set it to `None` outside. Right now it will be undefined if there was no exception and `handler` returned None for some reason.

flask/app.py
ThiefMaster: Does such a verbose comment even make sense here? It's kind of explaining the language itself which is not an appropriate use of comments. So I'd rather go for a short `# make error available outside except block (py3)` or something like that.

flask/app.py
untitaker: @ThiefMaster see function signature.

ThiefMaster: ah, didn't spot that.. was just looking for assignments

flask/app.py
untitaker: Tend to agree, but don't really care.

lobeck: you're right, simplified it by taking your comment

untitaker:
Thanks, could you squash these two commits?

    git rebase -i master

Then replace the second `pick` with `f`.

lobeck:
done

untitaker:
Sorry, I forgot, could you also add a testcase?

lobeck:
sure, i'll take a look and add one.

lobeck:
I added the testcase, can you please verify? It fails without my fix and passes with the fix.

I'm currently not sure if it's correct, that i have to set `app.config['SERVER_NAME']`, am i missing something?

PR #1534, opened 2015-07-16
make external_url_handler example py3 compliant

lobeck:
- a raises statement with multiple values is no longer allowed in python 3
- It's just a documentation change, no functionality changed.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -219,7 +219,7 @@ def url_for(endpoint, **values):
                 # Re-raise the BuildError, in context of original traceback.
                 exc_type, exc_value, tb = sys.exc_info()
                 if exc_value is error:
-                    raise exc_type, exc_value, tb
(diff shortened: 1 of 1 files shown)


untitaker:
We currently keep the examples polyglot, falling back to Python 2 if that's not reasonably possible. I'm open to change this, but it would have to be a very concentrated effort that changes all examples at once. I'm going to close this as it would make the docs inconsistent.

lobeck:
I'm not sure if i got you right, but this change is polyglot (if that's your concern) or are you rejecting this, because there may be just inconsistencies in the syntax used in the documentation.

I really would love to have this documentation changed because it helped me very much when first implementing the BuildError handling.

untitaker:
I didn't know that `with_traceback` was available in Python 2.7. I suppose this covers the most important Python 2 version.

untitaker:
Thanks!

PR #1535, opened 2015-07-16
Revert "make external_url_handler example py3 compliant"

lobeck:
Sorry, i should have tested this first and not rely on my IDE.

Actually in python 2.7 this causes `AttributeError: 'BuildError' object has no attribute 'with_traceback'`

Therefore this has to be reverted.

Reverts mitsuhiko/flask#1534


untitaker:
Argh, I tested on Python 3. Curse you, Arch!

lobeck:
for reference: https://bugs.python.org/issue11484

PR #1537, opened 2015-07-17
use blueprint specified template folder

alanhamlett:
Fixes issue #1361.

### DIFF
--- a/flask/templating.py
+++ b/flask/templating.py
@@ -126,9 +126,9 @@ def render_template(template_name_or_list, **context):
     template = None
 
     if _request_ctx_stack.top is not None and \
-            _request_ctx_stack.top.blueprint is not None and \
+            _request_ctx_stack.top.request.blueprint is not None and \
             isinstance(template_name_or_list, string_types):
-        bp = ctx.app.blueprints[_request_ctx_stack.top.blueprint]
+        bp = ctx.app.blueprints[_request_ctx_stack.top.request.blueprint]
         try:
             template = bp.jinja_loader.load(ctx.app.jinja_env,
                                             template_name_or_list,
--- a/flask/templating.py
+++ b/flask/templating.py
@@ -11,7 +11,8 @@
 from jinja2 import BaseLoader, Environment as BaseEnvironment, \
      TemplateNotFound
 
-from .globals import _request_ctx_stack, _app_ctx_stack
+from ._compat import string_types
+from .globals import _request_ctx_stack, _app_ctx_stack, current_app, request
 from .signals import template_rendered, before_render_template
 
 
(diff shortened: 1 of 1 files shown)


flask/templating.py
```diff
@@ -121,8 +121,24 @@ def render_template(template_name_or_list, **context):
...
     ctx = _app_ctx_stack.top
     ctx.app.update_template_context(context)
-    return _render(ctx.app.jinja_env.get_or_select_template(template_name_or_list),
-                   context, ctx.app)
+
+    template = None
+
+    if request.blueprint is not None and \
+            isinstance(template_name_or_list, string_types):
+        bp = current_app.blueprints[request.blueprint]
```
untitaker: Use `ctx.app`, it's already in scope.

untitaker: Also `_request_ctx_stack` contains the request, therefore no need to import `current_app` and `request`.

flask/templating.py
```diff
@@ -112,7 +113,6 @@ def _render(template, context, app):
 def render_template(template_name_or_list, **context):
     """Renders a template from the template folder with the given
     context.
-
```
untitaker: Why remove this line?

alanhamlett: My editor automatically trims whitespace from end of lines.

untitaker: Please undo this, I think this actually breaks the rst.

svieira:
Unless I am greatly mistaken this will break anyone out there who is using extensions and overriding (not inheriting from) the extension's blueprints.  For example, given the following extension setup:

```text
extension/
    __init__.py
    some.py
    extension_details.py
    templates/
        extension/
            base.html
            some-page.html
```

before this patch, overriding `base.html` would simply have been a matter of:

```text
the-app/
    __init__.py
    templates/
        the-layout.html
        extension/
            base.html
```

And making sure that `base.html@the-app` had the necessary `block` and `import` declarations.  This would shadow `base.html@extension`, allowing the app developer to override what the extension was doing without the extension developer having to provide a hook for such an overload

With this patch, I don't think there is a way to override `base.html@extension` in `extension` without patching Flask to bring back the old behavior (which will result in a lot of issues if new extensions start depending on this behavior).

davidism:
Agree with @svieira, this is the point I was making in [my comment on the original issue](https://github.com/mitsuhiko/flask/issues/1361#issuecomment-121746515).

> The point of searching the app template folder first is so that an app developer can override templates supplied by a blueprint from an extension.

ThiefMaster:
Those commits should be squashed into a single one eventually.

Also, this would be a breaking change depending on the application structure, so I'd rather make it configurable, with the current behavior being the default.

svieira:
If configurable it would have to be done at the `register_blueprint` method (since you could have one extension looking for the old behavior and one needed the new behavior).

I created #1548 to talk about some another possibility that might solve several of these issues in one go.

AzureusNation:
Is there any update on when this fix could be merged or implemented in a different manner?

davidism:
The referenced issue was closed as wontfix.

PR #1538, opened 2015-07-22
Moved open_session call from RequestContext.push to wsgi_app

### LINKED ISSUE #1528: Application/request stack corruption if unhandled exception is raised in session interface
RequestContext push is called in [wsgi_app method](https://github.com/mitsuhiko/flask/blob/master/flask/app.py#L1952), outside try..finally block. In [push() method](https://github.com/mitsuhiko/flask/blob/master/flask/ctx.py#L332) application's open_session call takes place, which can possibly lead to exception. But, since original call is not in try..finally block, removal of stack top never takes place. This leads to following problems:
### 1. Application context stack never updates again

RequestContext.push() only [checks](https://github.com/mitsuhiko/flask/blob/master/flask/ctx.py#L316) that application context exists or belongs to current application, but same rules apply to previous context.
### 2. Request context stack grows

Again, request context stack is pushed, but never removed. Actually just a memory leak :)
### 3. Misleading code comment about open_session usage

Inspired by [this comment](https://github.com/mitsuhiko/flask/blob/master/flask/ctx.py#L328), i started to develop multi-tenant application, and found nothing but disappointment. **You can't actually access request_context before open_session interface call.** This means, you can't configure which database to use before opening a session.
## My proposition.

Move open_session call to full_dispatch_request, and add interfaces and methods, similar to before_request - before_open_session, or call before_request before opening session (may break code that depends on session interface)

WeirdCarrotMonster:
Fixes #1528

New approach has only one drawback: in any custom tests that require sessions, you need to call `ctx_open_session` directly. I fixed existing tests in flask, but any user test may fail.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -91,6 +91,8 @@ Version 0.10.2
 - Changed logic of before first request handlers to flip the flag after
   invoking.  This will allow some uses that are potentially dangerous but
   should probably be permitted.
+- Fixed Python 3 bug when a handler from `app.url_build_error_handlers`
+  reraises the `BuildError`.
 
 Version 0.10.1
 --------------
--- a/docs/deploying/wsgi-standalone.rst
+++ b/docs/deploying/wsgi-standalone.rst
@@ -31,7 +31,7 @@ Gevent
 -------
 
 `Gevent`_ is a coroutine-based Python networking library that uses
-`greenlet`_ to provide a high-level synchronous API on top of `libevent`_
+`greenlet`_ to provide a high-level synchronous API on top of `libev`_
 event loop::
 
     from gevent.wsgi import WSGIServer
@@ -42,7 +42,7 @@ event loop::
 
 .. _Gevent: http://www.gevent.org/
 .. _greenlet: http://greenlet.readthedocs.org/en/latest/
-.. _libevent: http://libevent.org/
+.. _libev: http://software.schmorp.de/pkg/libev.html
 
 Twisted Web
 -----------
--- a/docs/patterns/wtforms.rst
+++ b/docs/patterns/wtforms.rst
@@ -1,7 +1,7 @@
 Form Validation with WTForms
 ============================
 
-When you have to work with form data submitted by a browser view code
+When you have to work with form data submitted by a browser view, code
 quickly becomes very hard to read.  There are libraries out there designed
 to make this process easier to manage.  One of them is `WTForms`_ which we
 will handle here.  If you find yourself in the situation of having many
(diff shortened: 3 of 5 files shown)


keyan:
@WeirdCarrotMonster I'm concerned about breaking backwards compatibility. Is there anyway to avoid this? Did you update the docs to reflect your proposed change?

WeirdCarrotMonster:
It's not as compatible as i'd like it to be. The problem is: current approach on managing RequestContext and sessions is wrong. I'm not sure if it can be fixed without breaking compatibility for some people, especially if they use `test_request_context` and `session` in their code.

I haven't updated the docs yet, since i hope there is better solution.

voxmens:
would it be possible to reimplement those functions using the newer apis?

ThiefMaster:
Also, please clean up the PR a bit. There shouldn't be merge commits in a PR at all (you always rebase feature branches to sync them with upstream) and the various test-related commits could be squashed into a single one.

tests/test_appctx.py
```diff
@@ -43,6 +43,34 @@ def test_request_context_means_app_context():
...
+    app = flask.Flask(__name__)
+
+    class BadSessionInterface(object):
+
+        """Fails once."""
+        fail = True
+
+        def open_session(self, *args, **kwargs):
+            if self.fail:
+                setattr(flask.g, "test_g_attr", 1)
```
ThiefMaster: Why setattr and not simply `flask.g.test_g_attr = 1`?

tests/test_appctx.py
```diff
@@ -43,6 +43,34 @@ def test_request_context_means_app_context():
...
+                raise Exception()
+        def make_null_session(self, *args, **kwargs):
+            return None
+
+    app.session_interface = BadSessionInterface()
+
+    try:
+        with app.test_request_context():
+            app.ctx_open_session()
+    except:
```
ThiefMaster: even in a test case, `except: pass` is a **terrible** idea.

tests/test_appctx.py
```diff
@@ -43,6 +43,34 @@ def test_request_context_means_app_context():
...
+        pass
+
+    with app.test_request_context():
+        app.ctx_open_session()
+        assert getattr(flask.g, "test_g_attr", None) is None
```
ThiefMaster: `flaks.g.get('test_g_attr')`

tests/test_appctx.py
untitaker: Use pytest.raises

On 3 August 2015 12:18:47 CEST, Adrian <notifications@github.com> wrote:
>> +        fail = True
>> +
>> +        def open_session(self, *args, **kwargs):
>> +            if self.fail:
>> +                setattr(flask.g, "test_g_attr", 1)
>> +                self.fail = False
>> +                raise Exception()
>> +        def make_null_session(self, *args, **kwargs):
>> +            return None
>> +
>> +    app.session_interface = BadSessionInterface()
>> +
>> +    try:
>> +        with app.test_request_context():
>> +            app.ctx_open_session()
>> +    except:
>
>even in a test case, `except: pass` is a **terrible** idea.
>
>---
>https://github.com/mitsuhiko/flask/pull/1538/files#r36072407

-- 
Sent from my phone. Please excuse my brevity.

PR #1539, opened 2015-07-22
Improve function with_metaclass()

mjhennig:
The idea of a metaclass *"that replaces itself with the actual metaclass"* is really clever. Yet I cannot help but suspect that the current implementation of function `with_metaclass` is just a bit too much.

Why does the dummy metaclass extend the one to apply?

It never utilizes any of the base class' resources, except for the base class itself. Since this is done by direct invocation (not via `super()` or something), there should be no need to inherit from the caller-provided type in the first place.

Furthermore, given that the above obvservation is correct, there should be no need to explicitly utilize `type.__call__` and `type.__init__` any more. Not only because one can directly inherit from `type` here, but also because both seem to be required due to the aforementioned inheritance only: If the given `meta` class implements custom magic, the creation of the `temporary_class` could fail. Or even worse; succeed but silently misinterprete arguments.

Speaking of the `temporary_class` creation (and still assuming all of the above assumptions to be correct): There's no need to do that via the `metaclass` constructor at all, which actually saves not only the additional function call but also the condition within `__new__`.

Given I didn't miss something (and there is indeed no actual reason for the metaclass to extend the one to apply), then I believe the patch suggested with this pull request to significantly improve the code without changing any of the expected behavior.

### DIFF
--- a/flask/_compat.py
+++ b/flask/_compat.py
@@ -54,15 +54,14 @@ else:
 
 
 def with_metaclass(meta, *bases):
+    """Create a base class with a metaclass."""
     # This requires a bit of explanation: the basic idea is to make a
     # dummy metaclass for one level of class instantiation that replaces
     # itself with the actual metaclass.
-    #
-    # This has the advantage over six.with_metaclass in that it does not
-    # introduce dummy classes into the final MRO.
-    constructor = lambda c, name, b, dct: meta(name, bases, dct)
-    metaclass = type('with_metaclass', (type,), {'__new__': constructor})
-    return type.__new__(metaclass, 'temporary_class', (object,), {})
+    class metaclass(type):
+        def __new__(cls, name, this_bases, d):
+            return meta(name, bases, d)
+    return type.__new__(metaclass, 'temporary_class', (), {})
 
 
 # Certain versions of pypy have a bug where clearing the exception stack
--- a/flask/_compat.py
+++ b/flask/_compat.py
@@ -56,21 +56,13 @@ else:
 def with_metaclass(meta, *bases):
     # This requires a bit of explanation: the basic idea is to make a
     # dummy metaclass for one level of class instantiation that replaces
-    # itself with the actual metaclass.  Because of internal type checks
-    # we also need to make sure that we downgrade the custom metaclass
-    # for one level to something closer to type (that's why __call__ and
-    # __init__ comes back from type etc.).
+    # itself with the actual metaclass.
     #
     # This has the advantage over six.with_metaclass in that it does not
     # introduce dummy classes into the final MRO.
-    class metaclass(meta):
-        __call__ = type.__call__
-        __init__ = type.__init__
-        def __new__(cls, name, this_bases, d):
-            if this_bases is None:
-                return type.__new__(cls, name, (), d)
-            return meta(name, bases, d)
-    return metaclass('temporary_class', None, {})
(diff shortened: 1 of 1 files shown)


mjhennig:
It turns out that [the *Six* version contains basically the same optimizations](https://bitbucket.org/gutworth/six/diff/six.py?diff2=eb096fc7ae5b&at=default). They have been applied just a few months after [migrating to use the *Flask* implementation](https://bitbucket.org/gutworth/six/issues/66/replace-the-implementation-of).

untitaker:
Could you copy over the exact one of six? Perhaps there's some significant difference we're missing.

On 23 July 2015 10:29:17 CEST, Matze <notifications@github.com> wrote:
>It turns out that [the *Six* version contains basically the same
>optimizations](https://bitbucket.org/gutworth/six/diff/six.py?diff2=eb096fc7ae5b&at=default).
>They have been applied just a few months after [migrating to use the
>*Flask*
>implementation](https://bitbucket.org/gutworth/six/issues/66/replace-the-implementation-of).
>
>---
>https://github.com/mitsuhiko/flask/pull/1539#issuecomment-124020806

-- 
Sent from my phone. Please excuse my brevity.

snoack:
@mjhennig asked me for my opinion on this one. FWIW, I think that using proper class syntax (as done by six and the current version of Flask) is more readable than creating classes with the `type()` built-in function.

However, there is also one logical difference between the implementation in six and this patch. That is that the implementation in six (as well as the current version of Flask) is subclassing the provided metaclass, rather than simply wrapping it by a metaclass subclassed from `type`. This could potentially lead to side-effects when the provided subclass is constructed with yet another subclass.

mjhennig:
While I see that unifying the implementations is favorable in this case, there are actually two points left I like to address:

First, there is indeed an *unintentional* difference in behavior in the proposed patch: It's now a metaclass called `with_metaclass`, not `metaclass` any more. That of course would be quick and easy to align, and it's not easy to find a scenario where it matters anyway.

Sencond, there is also an *intentional* difference between the behavior of the proposed version and the one from Six: The latter still extends the wrapper `metaclass` from the caller-given `meta`. In my 
undestanding, this is not only overhead but a bug:

    >>> class MetaMeta(type):
    ...     def __call__(cls, *args, **kwargs):
    ...       print("CALL", cls, args, kwargs)
    ...       return super(MetaMeta, cls).__call__(*args, **kwargs)
    ... 
    >>> class Meta(six.with_metaclass(MetaMeta, type)): pass
    ... 
    >>> class Test(six.with_metaclass(Meta, object)): pass
    ... 
    ('CALL', <class 'six.metaclass'>, ('Test', (<class 'six.temporary_class'>,), {'__module__': '__main__'}), {})
    ('CALL', <class '__main__.Meta'>, ('Test', (<type 'object'>,), {'__module__': '__main__'}), {})

When using `six.with_metaclass` for a metaclass that has a metaclass for itself, overwriting the `__call__` method, this very method is invoked for the now not-so-invisible-anymore wrapper.

Any feedback on this?

mjhennig:
I've updated the pull request to address the points discussed above. It's now basically the same as Six' version ([400:8a545f4e906f](https://bitbucket.org/gutworth/six/src/8a545f4)), except for using `type` as anchestor of the `metaclass`.

@snoack Thank you. Seems like we posted at the same time, eh?

snoack:
The new patch looks good to me.

untitaker:
Thanks!

Issue #1395, opened 2015-03-23
Support for mixins on MethodViews


taion:
It looks like @menudoproblema already has PR open for this: #1276 

I'm filing this issue to register my interest in having support for implementing HTTP methods via mixins on `MethodView` classes as well.

taion:
On second thought, maybe it's clearer not to have method handlers come in from mixins.

PR #1560, opened 2015-08-23
Fix static route when using host_matching = True

### LINKED ISSUE #1559: url_for('static', _external=True) gives bogus url with app.url_map.host_matching = True
``` python
app = Flask(__name__)
app.url_map.host_matching = True
app.url_map._rules_by_endpoint['static'][0].host = 'mydomain.com'  # XXX better way to do this?
    print url_for('static', filename='foo', _external=True)
```

=>
`'http:///static/foo'`

Expected `'http://mydomain.com/static/foo'`.

Am I missing something, or is this a bug?

Thanks for looking and for the great work on Flask.

jab:
Change `Flask.__init__` to accept two new keyword arguments, `host_matching` and `static_host`.

This enables host_matching to be set properly by the time the constructor adds
the static route, and enables the static route to be properly associated with
the required host.

Previously, you could only enable host_matching once your app was already
instantiated (e.g. `app.url_map.host_matching = True`), but at that point
the constructor would have already added the static route without host matching
and an associated host, leaving the static route in a broken state.

Fixes #1559.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -123,6 +123,9 @@ class Flask(_PackageBoundObject):
     .. versionadded:: 0.11
        The `root_path` parameter was added.
 
+    .. versionadded:: 0.13
+       The `host_matching` and `static_host` parameters were added.
+
     :param import_name: the name of the application package
     :param static_url_path: can be used to specify a different path for the
                             static files on the web.  Defaults to the name
@@ -130,6 +133,13 @@ class Flask(_PackageBoundObject):
     :param static_folder: the folder with static files that should be served
                           at `static_url_path`.  Defaults to the ``'static'``
                           folder in the root path of the application.
+                          folder in the root path of the application. Defaults
+                          to None.
+    :param host_matching: sets the app's ``url_map.host_matching`` to the given
+                          given value. Defaults to False.
+    :param static_host: the host to use when adding the static route. Defaults
+                        to None. Required when using ``host_matching=True``
+                        with a ``static_folder`` configured.
     :param template_folder: the folder that contains the templates that should
                             be used by the application.  Defaults to
                             ``'templates'`` folder in the root path of the
@@ -337,7 +347,8 @@ class Flask(_PackageBoundObject):
     session_interface = SecureCookieSessionInterface()
 
     def __init__(self, import_name, static_path=None, static_url_path=None,
-                 static_folder='static', template_folder='templates',
+                 static_folder='static', static_host=None,
+                 host_matching=False, template_folder='templates',
                  instance_path=None, instance_relative_config=False,
                  root_path=None):
         _PackageBoundObject.__init__(self, import_name,
@@ -525,19 +536,22 @@ class Flask(_PackageBoundObject):
         #:    app.url_map.converters['list'] = ListConverter
         self.url_map = Map()
 
+        self.url_map.host_matching = host_matching
+
         # tracks internally if the application already handled at least one
         # request.
         self._got_first_request = False
         self._before_request_lock = Lock()
 
-        # register the static folder for the application.  Do that even
-        # if the folder does not exist.  First of all it might be created
-        # while the server is running (usually happens during development)
-        # but also because google appengine stores static files somewhere
-        # else when mapped with the .yml file.
+        # Add a static route using the provided static_url_path, static_host,
+        # and static_folder iff there is a configured static_folder.
+        # Note we do this without checking if static_folder exists.
+        # For one, it might be created while the server is running (e.g. during
+        # development). Also, Google App Engine stores static files somewhere
         if self.has_static_folder:
+            assert bool(static_host) == host_matching, 'Invalid static_host/host_matching combination'
             self.add_url_rule(self.static_url_path + '/<path:filename>',
-                              endpoint='static',
+                              endpoint='static', host=static_host,
                               view_func=self.send_static_file)
 
         #: The click command line context for this application.  Commands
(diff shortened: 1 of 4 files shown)


flask/app.py
```diff
@@ -525,10 +531,14 @@ def __init__(self, import_name, static_path=None, static_url_path=None,
         # while the server is running (usually happens during development)
```
jab: (Not related to this PR, but in the line above in this comment it says "Do that even if the folder does not exist", yet the the code that follows actually only runs `if self.has_static_folder`. Is the comment wrong or just confusing?)

tests/test_basic.py
```diff
@@ -1616,3 +1616,13 @@ def run_simple_mock(hostname, port, application, *args, **kwargs):
     hostname, port = 'localhost', 8000
     app.run(hostname, port, debug=True)
     assert rv['result'] == 'running on %s:%s ...' % (hostname, port)
+    
+
+def test_static_route_with_host_matching():
+    app = flask.Flask(__name__, static_host='example.com', host_matching=True)
+    app.url_map.host_matching = True
+    c = app.test_client()
+    assert c.get('http://example.com/static/index.html').status_code == 200
```
jab: This would give 404 without my fix

tests/test_basic.py
```diff
@@ -1616,3 +1616,13 @@ def run_simple_mock(hostname, port, application, *args, **kwargs):
...
+    with app.test_request_context():
+        rv = flask.url_for('static', filename='index.html', _external=True)
+        assert rv == 'http://example.com/static/index.html'
```
jab: This would give `'http:///static/foo'` without my fix

flask/app.py
untitaker: `has_static_folder` doesn't indicate whether the folder exists or not, only if one has been configured.

untitaker:
I think a better fix would be if we delay registration of the static route until the first request. This sort of thing may sound radical, but is already done for a lot of things that depend on configuration changes done past the instantiation of `Flask()`.

jab:
@untitaker Thanks for taking a look at this and sorry for the delay in responding! I've been swamped with urgent work, but I'd still love to contribute a fix for this issue, and make any required changes to get it landed, ideally before I forget what the problem is again and have to start over.

> I think a better fix would be if we delay registration of the static route until the first request.

The user still needs a way to pass a host for the static route though, right? In that case, what is the advantage of delaying registration of the static route until the first request, when we can do it correctly at app instantiation time provided a `static_host` param?

Even still, I tried delaying registration of the static route until the first request anyway just to see if it would work, but it didn't. Ultimately there is a chicken-and-egg problem. By the time we get the first request and go to add the static route, if the first request was for the static endpoint, we will already get an incorrect 404 for it before we've had a chance to add the static route.

I worked up a new PR for this in #2211 which applies cleanly against latest master. In that version, rather than add the two new constructor parameters `host_matching` and `static_host`, where `host_matching` *should* be True if `static_host` is passed, I made it so that there's only the single new `static_host` parameter, which if passed causes `host_matching` to be set to True implicitly. Maybe that's preferable.

untitaker:
Because your fix 1.) introduces another way to do the same thing 2.) will not work for existing code that modifies `url_map` in arbitrary ways.

I see how my suggestion falls apart, but I definitely think that we can initialize the static URL path before the request gets routed, so static requests should work for the first request as well.

I suppose `static_host` requires `host_matching` to be set? In that case we should raise an exception for the case `static_host is not None and not host_matching`. This is doable in both possible implementations.

jab:
Thanks for the continued review, @untitaker. I addressed your comments in 7ea1ed6 as part of #2211. Specifically:
- Defer adding the static route until right before the first request.
- No longer set url_map.host_matching to True implicitly if static_host is
  provided. Instead raise an error at the time the static route is added if
  static_host is truthy and host_matching is falsy.

Please take a look.

I'll close this PR since #2211 is the one with the desired changes. Happy to squash and rebase that one as desired.

jab:
Thanks for reopening, @untitaker. Just pushed 919e5dc. I think that should do the trick to get this all resolved. Please let me know if any further changes are needed.

jab:
Done in bbd88d3.

jab:
This has been approved by @untitaker and @davidism (thanks for reviewing!). Can this be merged?

untitaker:
There is a ResourceWarning when running the tests on Python 3 that needs to be fixed. I think you need to close your response explicitly to make this error go away.

jab:
Oh, I thought that was due to an unrelated issue (#513). Added an explicit `close()` in 5511315 and now the tests are passing for me locally on Python 3, thanks @untitaker! Ready to merge now?

(Note: some Python 2 targets are now failing for REQUIREMENTS=devel and REQUIREMENTS=devel-simplejson, but [that was the case for the most recent build of master](https://travis-ci.org/pallets/flask/builds/218619063) too (revision ae1ac2053bcf0c77de32fc7915e36f8ca2f5c961), so that seems unrelated to this PR and shouldn't block merging.)

untitaker:
I now kicked over the broken builds. We can't test against those devel versions if they fail to install.

untitaker:
Thanks!

jab:
My pleasure, and thank you for reviewing and merging!

bmjjr:
Thanks for this fix all. It quickly helped solve the issue of finding a way to easily allow a reverse proxy to multiple domains. But, breaks debug toolbar, similar to flask-admin mentioned in #1395

PR #1568, opened 2015-09-14
Url conflict assert

rawrgulmuffins:
This PR is an attempt to reconcile the fact that you can't have two routes with the same function name but you can create two routes with the same URL.

```
from flask import Flask
app = Flask("the_flask_module")

@app.route("/hello")
def hello_page():
    return "I'm a hello page"

@app.route("/different_route")
def hello_page():
    return "Different route"
```

will throw an `AssertionError`

but 

```
@app.route("/hello")
def end_point1():
    return "I'm a hello page"

@app.route("/hello")
def end_point2():
    return "Different route"
```

wont. In fact the second route is unreachable. 

This PR checks to see if three conditions are true before raising an assertion. The new rule has to have a `subdomain`, `url_path`, and one `method` that overlap with a current rule in the `self.url_map object`.

There's a couple of wonky pieces in this pull request that were required to get all of the unit tests to pass.

```
if prior_rule.rule == "/static/<path:filename>":
    continue
```

was required to get blueprints to not throw an assertion since each blueprint registers a `"/static/<path:filename>"` rule for each blueprint. I tried to find a way to not have to use a static string but alas I failed. 

Secondly 

```
 methods_minus_required = ((rule.methods & prior_rule.methods) 
     - required_methods)
```

is used to check for all explicitly registered methods for a route since there were a bunch of cases where the `OPTIONS` method was implicitly added to rules that shared a url_path.

---

This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/1568)

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -515,6 +515,10 @@ class Flask(_PackageBoundObject):
         #:    app.url_map.converters['list'] = ListConverter
         self.url_map = Map()
 
+        # tracks internally if the application is adding a new rule that has
+        # already been added previously.
+        self.prior_rules = {}
+
         # tracks internally if the application already handled at least one
         # request.
         self._got_first_request = False
@@ -1029,6 +1033,32 @@ class Flask(_PackageBoundObject):
         rule = self.url_rule_class(rule, methods=methods, **options)
         rule.provide_automatic_options = provide_automatic_options
 
+        # a bound subdomain's null value is '' and a unbound is None.
+        current_subdomain = rule.subdomain
+        if current_subdomain is None:
+            # Normalizing.
+            current_subdomain = ''
+        prior_rule_check = (rule.rule, current_subdomain)
+        if prior_rule_check in self.prior_rules:
+            # static/<path:filename> is added once per blueprint. Ignoring for
+            # this check.
+            if rule.rule != "/static/<path:filename>":
+                prior_rule_methods = self.prior_rules.get(
+                    prior_rule_check,
+                    None)
+                # Flask adds non-specified methods to routes. Removing the
+                # required methods so we don't get false positives.
+                intersection_minus_required = bool(
+                    (rule.methods & prior_rule_methods) - required_methods)
+                if intersection_minus_required:
+                    error = ('View function mapping is overwriting an existing '
+                            'URL Rule: path: %s,  method: %s' % (
+                                rule,
+                                intersection_minus_required))
+                    raise AssertionError(error)
+        else:
+            self.prior_rules[prior_rule_check] = rule.methods
+
         self.url_map.add(rule)
         if view_func is not None:
             old_func = self.view_functions.get(endpoint)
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I think all these commits could be easily squashed into a single one.

rawrgulmuffins:
Done, everything was squashed down to that one commit.

ThiefMaster:
I think you have some leftovers from failed merges with upstream in there. Check the diff and remove anything that's not actually related to your PR (and maybe try to switch to a <=51-char first line in the commit message when committing it again)

flask/app.py
```diff
@@ -1029,6 +1029,27 @@ def index():
         rule = self.url_rule_class(rule, methods=methods, **options)
         rule.provide_automatic_options = provide_automatic_options
 
+        for prior_rule in self.url_map.iter_rules():
```
ThiefMaster: This is not very efficient - you iterate over all existing elements whenever you add a new one. So you end up with `sum(range(num_rules))` total iterations.

rawrgulmuffins: I would love to not have to iterate over all rules but `WerkZeug.Map` only has `iter_rules` as a search mechanism. Another potential is that we add the tuple that we care about `(subdomain, method, url_path)` to a set and see if there's existence in the set. I wasn't sure if it would be better to add a new class attribute or to only use existing structures.

ThiefMaster: Maybe the whole thing would be a better fit in Werkzeug than Flask?

rawrgulmuffins: Part of why I added this change to `add_url_rule` is because this function holds the assert that makes sure no function name is used twice.

If you still think the modification should be moved to `WerkZeug.Map` should I follow the same tuple check?

rawrgulmuffins:
I rebased against master and readded my changes. Looks like I missed a couple of wonky space changes, in `app.py`. I also changed to a hashing check rather than a iteration check.

rawrgulmuffins:
The last commit was to add some in-line comments that got dropped when I re-added the changes.

flask/app.py
```diff
@@ -1029,6 +1033,32 @@ def index():
...
+            current_subdomain = ''
+        prior_rule_check = (rule.rule, current_subdomain)
+        if prior_rule_check in self.prior_rules:
+            # static/<path:filename> is added once per blueprint. Ignoring for
+            # this check.
+            if rule.rule != "/static/<path:filename>":
+                prior_rule_methods = self.prior_rules.get(
+                    prior_rule_check,
+                    None)
+                # Flask adds non-specified methods to routes. Removing the
```
joequery: This is the only part of the PR I don't quite follow, which is due to my ignorance of Flask internals. What do you mean by "Flask adds non-specified methods to routes"?

rawrgulmuffins: Flask adds a `OPTIONS` method to every URL (that's the `required_methods` variable). I might need to add a unit test to make sure that options methods are still routable.

luke0922:
why not this pr merge into master?

untitaker:
In principle I support a mechanism that detects and prevents shadowing of other
URL rules, but this implementation isn't acceptable due to a few reasons:

- The static path is hardcoded for some special casing
- `prior_rules` will only be updated if the app registers its routes the
  "normal" way. It is possible for those two datastructures to get out of sync
  when the app dev modifies them directly, which is still practiced sometimes.
- Unrelated code reformatting, not a minimal diff

untitaker:
I will just close this since it's also unclear to me how a better implementation could look like.

ThiefMaster:
Probably a "better implementation" should be in Werkzeug, not in Flask

PR #1577, opened 2015-10-05
Warn when using development server

sprin:
Also updates docs to emphasize that run() should not be used
in production.

Refs #1102

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -770,8 +770,13 @@ class Flask(_PackageBoundObject):
         return rv
 
     def run(self, host=None, port=None, debug=None, **options):
-        """Runs the application on a local development server.  If the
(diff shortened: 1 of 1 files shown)


jeffwidman:
@sprin thanks for putting this together. Looking #1102, there's split opinions on whether a runtime warning is helpful vs quickly annoying. (I'm undecided myself.) 

If you split this into separate PRs for the code comment change and the warning, I'll happily merge the code comment change.

ThiefMaster:
I really don't like an unconditional warning for this. For anyone besides someone who uses Flask for the first time and doesn't know that the server is only for development this is going to be incredibly annoying.

jeffwidman:
Thanks. Merging, as this is now only a code-comment improvement, not a warning.

PR #1585, opened 2015-10-19
Added flake8 for checking Python syntax

### LINKED ISSUE #1561: add automatic pep8 test in travis
It's very useful and easy to modify travis.yml。

Akasurde:
Fixes #1561 
Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,10 @@ matrix:
 
 install:
     - pip install tox
+    - pip install flake8
+
+before_script:
+    - flake8 .
 
 script:
     - tox -e py-$REQUIREMENTS


ThiefMaster:
I don't think this PR is worth merging unless it also includes a commit that:

- configures flake8 in a sane way (e.g. don't complain about unused imports in `__init__.py` and ignore anything outside the `flask` package)
- fixes all problems

ThiefMaster:
Using `before_script` doesn't even run the tests if flake8 fails. I don't think this is a good idea. Especially for pull requests it's not very helpful if failing style checks skip the unit tests.

ThiefMaster:
Would it make sense to run flake8 from tox.ini?

[pytest-pep8](https://pypi.python.org/pypi/pytest-pep8) could also be an option

Akasurde:
OK. Let me try flake8 from tox.ini

Akasurde:
@ThiefMaster How we are going to handle existing pep8 errors?

zoidbergwill:
Fixing them all might be the easiest option, with `# noqa`s where they make sense.

@ThiefMaster recommending fixing them all [here](https://github.com/mitsuhiko/flask/pull/1585#issuecomment-149191309)

adrianmoisey:
Another option is to pass `--exit-zero` to `flake8` for now, until all the issues can get cleaned up

davidism:
continued in #1702

PR #1598, opened 2015-10-29
Add info about working with VE and Python 3

inirudebwoy:
Python 3 does not have execfile function.


davidism:
This should use `with open` so that the file gets closed after.

---

Unrelated to the specific change:

I'm not too thrilled that we're encouraging users to activate a virtualenv in code or modify it's path at runtime.  That's just bad practice.  Is this really the only way to use mod_wsgi with a virtualenv?

inirudebwoy:
Changed to use ```with open```

davidism:
squashed and merged in 99bc0df

Issue #1600, opened 2015-10-30
Add session methods to handle destruction and regeneration


twolfson:
We have been using `flask-session` for our sessions and noticed that there were no easy methods on `flask` or `flask-session` to handle session destruction/regeneration. These are practical as they aid in preventing session fixation attacks (e.g. when someone knows your session id and it doesn't get changed on login so they can reuse your session):

https://www.owasp.org/index.php/Session_fixation

We saw that `flask-kvsession` has some logic for this:

https://github.com/mbr/flask-kvsession/blob/0.6.2/flask_kvsession/__init__.py#L89-L124

but we were already invested into `flask-session` (added `sha256` signing support) that we decided to implement similar logic there as well:

https://github.com/fengsp/flask-session/pull/27

However, I'm not sure that's the ideal interface. I think the final product would look something more like

``` python
flask.session.destroy()
flask.session.regenerate()
```

which in turn call `flask.session_interface.destroy_session(self, app, session)` and `flask.session_interface.regenerate_session(self, app, session)`.

What are your thoughts on this?

davidism:
This seems to be specific to session interfaces that need a session id, and those two examples you gave seem to be handling it already.  What do we gain by adding this to Flask?  The default session doesn't use a session id, and can be wiped with `session.clear()`.

twolfson:
Ah, good point. I think the benefit of implementing it in Flask (even as `NotImplementedError`) would lead to more consistent APIs for sessions (e.g. everyone's already using `session_interface` and `session.clear()`). It feels like `session.destroy` and `session.regenerate` would be a natural extension.

On a side note, `session.clear` and `session.destroy` are similar but not the same. `clear` erases the contents of the session but still persists a cookie. `destroy` empties session contents, remove it from the store (if there is a non-cookie one), and delete the cookie entirely.

davidism:
Removing the cookie doesn't actually seem to be part of the security, only clearing the data or changing the id matter.  The cookie *is* removed if the dict is empty, so the subclass could just override `clear` to also clear the server side storage.  Regenerate just becomes "change the value of the `id` key", or "remove the id, and let it be generated when `save_session` doesn't find it".  I guess that could use a standard name.  I'm not sure about encouraging this though, it seems more secure to clear the session when logging in, and make sure anything that needs to persist for a user is persisted to the database, not the cookie.

twolfson:
In a session id context, regenerate does more than change the value of the `id` key. It also corresponding session data in the store.

[...]

PR #1603, opened 2015-11-02
Added session regenerate/destroy methods

twolfson:
In #1600, we have been discussing adding session regeneration/destroy methods to make preventing session fixation part of Flask. This is a rough draft based on that discussion. In this PR:
- Added `app.session_regenerate` and `app.session_destroy` methods to `app.py`
- Added `session_interface.regenerate_session` and `session_interface.destroy_session` to session interfaces
- Added tests against cookie interface verifying the methods work as expected
- Updated quickstart documentation to include notes on when to use `session_regenerate` and `session_destroy`

**Missing:**
- The documentation probably needs more thought/work (e.g. add to API, maybe other documents).

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -919,6 +919,35 @@ class Flask(_PackageBoundObject):
         """
         return self.session_interface.save_session(self, session, response)
 
+    def regenerate_session(self, session):
+        """Regenerate the session if it is session id based.  For the default
+        implementation, this is a noop.  Instead of overriding this
+        method we recommend replacing the :class:`session_interface`.
+
+        This should be used upon login to prevent session fixation.
+
+        https://www.owasp.org/index.php/Session_fixation
+
+        :param session: the session to be saved (a
+                        :class:`~werkzeug.contrib.securecookie.SecureCookie`
+                        object)
+        """
+        return self.session_interface.regenerate_session(self, session)
+
+    def destroy_session(self, session):
+        """Destroy the session.  Instead of overriding this
+        method we recommend replacing the :class:`session_interface`.
+
+        This should be used upon logout to prevent session fixation.
+
+        https://www.owasp.org/index.php/Session_fixation
+
+        :param session: the session to be saved (a
+                        :class:`~werkzeug.contrib.securecookie.SecureCookie`
+                        object)
+        """
+        return self.session_interface.destroy_session(self, session)
+
     def make_null_session(self):
         """Creates a new instance of a missing session.  Instead of overriding
         this method we recommend replacing the :class:`session_interface`.
--- a/flask/sessions.py
+++ b/flask/sessions.py
@@ -278,6 +278,18 @@ class SessionInterface(object):
         """
         raise NotImplementedError()
 
+    def regenerate_session(self, app, session):
+        """This is called for regenerating sessions to prevent session fixation
+        attacks.
+        """
+        raise NotImplementedError()
+
+    def destroy_session(self, app, session):
+        """This is called for destroying a session's contents and underlying store
+        information. This is used to prevent session fixation attacks.
+        """
+        raise NotImplementedError()
+
     def save_session(self, app, session, response):
         """This is called for actual sessions returned by :meth:`open_session`
         at the end of the request.  This is still called during a request
@@ -329,6 +341,16 @@ class SecureCookieSessionInterface(SessionInterface):
             return self.session_class(data)
         except BadSignature:
             return self.session_class()
+
+    def regenerate_session(self, app, session):
+        # There isn't anything related to a session id so we do nothing
+        pass
+
+    def destroy_session(self, app, session):
+        # Empty store contents
+        session.clear()
+
+        # `save_session` will take care of deleting our cookie
 
     def save_session(self, app, session, response):
         domain = self.get_cookie_domain(app)
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -779,6 +779,10 @@ sessions work::
     @app.route('/login', methods=['GET', 'POST'])
     def login():
         if request.method == 'POST':
+            # Regenerate session to prevent session fixation
+            app.regenerate_session(session)
+
+            # Handle login
             session['username'] = request.form['username']
             return redirect(url_for('index'))
         return '''
@@ -790,8 +794,8 @@ sessions work::
 
     @app.route('/logout')
     def logout():
-        # remove the username from the session if it's there
-        session.pop('username', None)
+        # Destroy the session to prevent reuse (e.g. session fixation)
+        app.destroy_session(session)
         return redirect(url_for('index'))
 
     # set the secret key.  keep this really secret:
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -449,6 +449,66 @@ def test_session_cookie_setting():
     run_test(expect_header=False)
 
 
+def test_session_cookie_regenerate():
+    app = flask.Flask(__name__)
+    app.testing = True
+    app.secret_key = 'dev key'
+
+    @app.route('/set', methods=['POST'])
+    def set():
+        flask.session['value'] = flask.request.form['value']
+        return 'value set'
+
+    @app.route('/get')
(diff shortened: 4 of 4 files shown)


flask/app.py
```diff
@@ -919,6 +919,35 @@ def save_session(self, session, response):
...
+        """Regenerate the session if it is session id based.  For the default
+        implementation, this is a noop.  Instead of overriding this
+        method we recommend replacing the :class:`session_interface`.
+
+        This should be used upon login to prevent session fixation.
+
+        https://www.owasp.org/index.php/Session_fixation
+
+        :param session: the session to be saved (a
+                        :class:`~werkzeug.contrib.securecookie.SecureCookie`
```
ThiefMaster: it's not necessarily a SecureCookie session

RonnyPfannschmidt:
at first glace this looks like a "breaking" change to me (external session implementer will break)

as such i think this needs a major release and/or compat handling

ThiefMaster:
Since session interfaces are supposed to inherit from `SessionInterface` it'd only break custom interfaces that don't inherit from that class.

flask/sessions.py
```diff
@@ -330,6 +342,16 @@ def open_session(self, app, request):
         except BadSignature:
             return self.session_class()
 
+    def regenerate_session(self, app, session):
+        # There isn't anything related to a session id so we do nothing
+        pass
+
+    def destroy_session(self, app, session):
+        # Empty store contents
+        session.clear()
```
ThiefMaster: Would this be a good default behavior in the base class?

twolfson: No, we shouldn't implement anything in the base class. When destroying a session we need to both destroy the cookie and its underlying store value. Otherwise, a user could "log out" (destroy their browser cookie) but an attacker could reuse the browser cookie and still access their session (since the underlying store data exists).

flask/app.py
twolfson: We used the same description as `save_session`. I think we should keep it as `SecureCookie` for consistency and update those in another PR

twolfson:
It won't be a breaking change as people must call `regenerate` and `destroy` themselves (e.g. at login, logout). Since these are new methods, nobody will be using them yet so nobody will receive the `NotImplementedError`

RonnyPfannschmidt:
that makes breakage more devious, because the downstreams that implement the interface wont have a indication, and up-streams trying to use those projects will be unable to do so because the api changed in a non-indicative way

twolfson:
I think you are conflating an issue of documenting changes with the feature itself. As stated before, these methods are not automatically invoked so nothing can break.

If someone wants to use these features, then they must make sure that all the corresponding libraries using these features know to use these features. But that isn't a breaking change in this library, it's a breaking change when shifting around dependencies (in the projects that use said dependencies).

untitaker:
I don't understand why the code in a potential `destroy_session` implementation
couldn't be added to the implementation of `clear`. Sure there is a semantic
difference between the two, but I can't imagine a situation where the
application developer would really like to just clear the session instead of
destroying it. I would therefore suggest that `clear` does the job of
`destroy_session`. That's better than having some sort of opt-in security this
PR provides.

The same can be said for `regenerate_session`.  Assuming `clear` changes the
session ID like I suggested above, you might as well just call `session.clear`
before logging the user in. You would have to slightly change the
implementation of `clear` to actually reset the session when it is called, not
when the session is empty. But in the end you would achieve all the desired
results without adding extra API surface.

Apart from that I don't consider it effective to call `regenerate_session` in
your code to mitigate the potential catastrophe an XSS in your application
could do (which is, as far as I understand, the only way you could steal the
cookie here). It is not clear at all where one would have to put those
`regenerate_session` calls. Doing it just before login doesn't fully mitigate
the effects an XSS could have. Better idea is to configure your CSP to avoid
the execution in the first place.

twolfson:
@untitaker These methods are to prevent session fixation attacks. The simplest example is:

- Attacker sets a user's session id before they log in
- User visits site, logs in, and their session id doesn't change
- Attacker can use the session id they gave to user to visit site as though they are user

More content here: https://www.owasp.org/index.php/Session_fixation

With respect to using `clear()` as a replacement, `clear()` would remove the contents of the session but the session id would stay the same.

With `destroy_session`, we both unset the session id cookie from the user's browser as well as delete it from the underlying store. This is typically used on log out. Due to erasure from the underlying store, it prevents the attacker from accessing the session content.

With `regenerate_session`, we generate a new session id, keep the same session data, update the session id cookie in the user's browser, and update the session's key in the underlying store. This is typically used on log in. Due to key shuffling, it prevents the attacker from reusing the previously given session id to access the logged in session.

A scenario where we would want  a `regenerate_session` instead of a `clear()` would be something like restoring a user's page on log in (e.g. visit `/item/123` but it requires log in, after they log in we redirect back to `/item/123`)

untitaker:
On Fri, Apr 21, 2017 at 12:17:08PM -0700, Todd Wolfson wrote:
> @untitaker These methods are to prevent session fixation attacks. The simplest example is:
> 
> - Attacker sets a user's session id before they log in
> - User visits site, logs in, and their session id doesn't change
> - Attacker can use the session id they gave to user to visit site as though they are user
> 
> More content here: https://www.owasp.org/index.php/Session_fixation

I know.

> 
> With respect to using `clear()` as a replacement, `clear()` would remove the contents of the session but the session id would stay the same.

I already said that `clear` might as well change the session ID. I know this
isn't the case at the moment.

The rest of what you wrote misses the point I tried to make, which is that
there are much more points in time where an XSS could be used to steal an
existing session. Defending against all of them would imply that you have to
sprinkle `regenerate_session` everywhere, which is a non-solution. The correct
solution is to prevent hijacking via e.g. CSP, not mitigating its effects.

davidism:
What session id are we referring to? Flask doesn't set one. Any implementation that does can easily clear it in `clear`. I agree with untitaker, this can already be done with the existing methods but should really be handled using CSP headers.

untitaker:
When I referred to session ids I was assuming a backend that had those. 

On 21 April 2017 21:58:55 GMT+02:00, David Lord <notifications@github.com> wrote:
>What session id are we referring to? Flask doesn't set one. Any
>implementation that does can easily clear it in `clear`. I agree with
>untitaker, this can already be done with the existing methods but
>should really be handled using CSP headers.
>
>-- 
>https://github.com/pallets/flask/pull/1603#issuecomment-296292040

-- 

twolfson:
Typically these methods are used sparsely (i.e. on login, forgot password success, and log out). I think "sprinkingly everywhere" is making it sound a lot more common than it is.

While it's great to consider having no XSS issues as a solution to prevent leaking session ids, these are 2 different attacks; one sets the session id, the other extracts it. I would hope that adding these methods leads to more documentation and convention around always using `regenerate_session()` and `destroy_session()`, thus educating newer developers instead of handwaving that `clear()` should be used.

With respect to CSP, it's only supported by modern browsers whereas session destruction/regeneration is supported everywhere.

For reference, these methods already exist in a similar Node.js framework:

- https://github.com/expressjs/session/tree/v1.15.2#sessionregeneratecallback

untitaker:
What is handwavy about assigning this purpose to `clear()`?

twolfson:
It's not semantic making it hard for experienced developers to find and doesn't educate developers unfamiliar with session fixation attacks

untitaker:
Again, CSP fully prevents this problem while regenerating session IDs fixes a very specific kind of attack that can be done via XSS. The vulnerability (the XSS) is still there, and can be exploited in other ways.

The worst about this though that it's basically opt-in security, and even then a very manual kind. It requires developers to add method calls "strategically".

Therefore I will close this. I'm open to other ways to prevent session fixation, but this one isn't it.

davidism:
Two good steps would be:

* patch flask-session to reset the id on `clear`
* adjust our tutorial to use `clear` during login and logout instead of `pop`-ing specific keys

untitaker:
Probably. I'm still unsure if we really want to encourage a practice that *only might decrease the likelyhood* of an exploit running, for a lack of data.

PR #1631, opened 2015-11-28
Extend documentation about uwsgi/nginx deployment

stephane:
This patch intends to provide more information about uwsgi/nginx deployment when the app callable is in (``yourapp/__init__.py``) and the application is attached to the '/' URL.
I'm not English native so I expect some remarks ;)

Related to previous #1464


jeffwidman:
@stephane At first glance this looks useful. Can you rebase?

PR #1634, opened 2015-12-02
add before_init param to View.as_view

guyskk:
I am writing a flask ext and need to do something before View init, 
but I can't do that via app.before_request because I need to test my View without request.

### DIFF
--- a/flask/views.py
+++ b/flask/views.py
@@ -31,6 +31,16 @@ class View(object):
 
         app.add_url_rule('/hello/<name>', view_func=MyView.as_view('myview'))
 
+    You can dosomething before View init::
+
+        def dosomething():
+            # you can return a Response then View will not init
+            # return None to continue
+            return None
+
+        view_func=MyView.as_view('myview', before_init=dosomething)
+        app.add_url_rule('/hello/<name>', view_func=view_func)
+
     When you want to decorate a pluggable view you will have to either do that
     when the view function is created (by wrapping the return value of
     :meth:`as_view`) or you can use the :attr:`decorators` attribute::
@@ -70,7 +80,7 @@ class View(object):
         raise NotImplementedError()
 
     @classmethod
-    def as_view(cls, name, *class_args, **class_kwargs):
+    def as_view(cls, name, before_init=None, *class_args, **class_kwargs):
         """Converts the class into an actual view function that can be used
         with the routing system.  Internally this generates a function on the
         fly which will instantiate the :class:`View` on each request and call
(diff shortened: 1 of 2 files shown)


davidism:
Please provide a concrete example of what you're trying to do with this as well as why you can't test a view with a request.  I'm not clear how this is different than just overriding `__init__` to call `do_something` first, or otherwise subclassing `View`.  Additionally, this patch needs tests and must pass existing tests as well.

guyskk:
I have  a Resource class subclassing View, and a Api class, just like flask-restful.
Api need to parse request before Resource init, such as check user's permission.

And the test is like this:

            with api.test_client() as c:
                rv,code,header = c.resource.action(data)
                assert code == 200
                assert rv == {"hello":"world"}
                assert c.resource.action_need_login(data).code == 403

            with api.test_client(user_id) as c:
                assert c.resource.action_need_login(data).code == 200
                assert c.resource.action_need_login(data).rv == {"hello":"guyskk"}

It will pass user_id and data directly.

guyskk:
A example:

    app = Flask(__name__)
    api = Api(app)

    class Hello(Resource):

        def get(self):
            
            return {"hello": "name"}

    api.add_resource(Hello)

In api, `_restaction_test` means on testing.

    def _before_resource_init(self):

        if not getattr(g, "_restaction_test", False):
            resource, action = parse_request()
            me = self.parse_auth_header()
            request_data = get_request_data()
            g.resource = resource
            g.action = action
            g.me = me
            g.request_data = request_data

        user_role = self._fn_user_role(g.me["id"])
        permit, res_role = self._permit(user_role, g.resource, g.action)
        g.me["role"] = res_role
        if not permit:
            if g.me["id"] is None:
                abort(403, "permission deny: your token is invalid")
            else:
                abort(403, "You don't have permission: user_role=%s" % user_role)

davidism:
What is `c.resource`?  It's not a standard part of Flask.  Your example doesn't demonstrate the need for your patch.  Class creation should not have side effects that you might need to bypass.  Use decorators to check for permissions, which is already supported by `View`.  This seems like an incorrect design choice rather than a missing feature on Flask's part.

guyskk:
You are right
thank you

flask/views.py
```diff
@@ -31,6 +31,16 @@ def dispatch_request(self, name):
 
         app.add_url_rule('/hello/<name>', view_func=MyView.as_view('myview'))
 
+    You can dosomething before View init::
```
ThiefMaster: typo (dosomething)

PR #1636, opened 2015-12-03
update flaskr example design

mscalora:
convert flaskr example to responsive design using bootstrap 3


davidism:
The point of the examples is to be concise and straightforward.  There's a place for an example of how to use bootstrap, but not in the Flask repo.

PR #1640, opened 2015-12-05
Allow application configurations to control the debugger

nZac:
I was surprised when running my application from the CLI that it didn't respect the applications debug flag.  This patch expands `DispatchApp`'s api to return the applications debug flag which is used by `run_simple` to activate the debugger and supply an informative message.

I explored enabling the reloader along with enabling the debugger but that caused a double execution. If I understand it correctly, the double load is the result of `eager_loading` not being switched properly because the reload status (if activated by the applications debug flag) wouldn't be known until after loading has occurred.

Since I am not positive what side-effects (none to my limited understanding) double execution would have I left that code out; it is a trivial addition if a double execution is acceptable in this case.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -105,6 +105,15 @@ class DispatchingApp(object):
     of the Werkzeug debugger means that it shows up in the browser.
     """
 
+    @property
+    def debug(self):
+        try:
+            return self._app.debug
+        except AttributeError as e:
+            print('Application not created yet, can\'t determine debug '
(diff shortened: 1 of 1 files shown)


nZac:
Unfortunately, this wont work as I would like. When `eager_loading=False` the application won't be ready until after `run_simple` is called meaning `app.debug` can't be detected. 

The sentiment still remains, it would be nice allow configuration to control the debugger and reloader though that might not be possible.

PR #1653, opened 2015-12-15
Fix wrong indentation error and get rid of PEP8 warnings in helpers.py

rochacbruno:
- There was wrong indentation error raising in line 587 https://github.com/mitsuhiko/flask/blob/master/flask/helpers.py#L587

- Also fixed some PEP8 warnings


davidism:
continued in #1702

PR #1657, opened 2015-12-18
Readme

Satender052:
I have updated README now its looking a bit good :)

### DIFF
--- a/README
+++ b/README
@@ -1,47 +1,47 @@
 
 
-                          // Flask //
+                          ##Flask 
 
-              web development, one drop at a time
+              #web development, one drop at a time
 
 
-    ~ What is Flask?
+##What is Flask?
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
First of all, these four commits should be squashed into a single one. Their commit messages are pretty useless but times four even more so. "Format README" would be a better message.

Anyway, @mitsuhiko has a strong preference for a simple plaintext README, so I doubt this PR will be merged.

ThiefMaster:
Just noticed that you created lots of PRs that just do minor changes to README files (also in other projects) that don't actually improve anything.

Please, don't do that! It's just noise for the maintainers of those projects. If you want to contribute to a project, pick some issue that's actually useful and try to solve/implement it and create a PR for that.

PR #1666, opened 2015-12-27
Fixed mistake in CONTRIBUTING.rst file, regarding pip install.

AvivC:
Changed  'pip install --editable .' to 'pip install --editable flask', which works. Also added an explanatory comment, since as a beginner to pip I found this confusing.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -49,6 +49,7 @@ Clone this repository::
 
 Install Flask as an editable package using the current source::
 
+    cd flask
     pip install --editable .
 
 Then you can run the testsuite with::


ThiefMaster:
I think `cd flask` after cloning would make more sense.

AvivC:
Very well, than I will add a cd flask to the code.

ThiefMaster:
BTW, it's usually a good idea to keep git commits formatted like this:

```
short summary, no more than 51 chars

more verbose information if needed
...
...
```

various tools and also GitHub show the first line as a summary, and if it's too long it gets truncated which is quite ugly

AvivC:
You mean a line break between the summary and the rest of the info? And thanks for the feedback, rather new to GitHub contribution :)

ThiefMaster:
http://chris.beams.io/posts/git-commit/ is a pretty good guide on how to write great commit messages

davidism:
These instructions assume you are in the project root (because you've been fixing a bug in the project and want to contribute it).  `cd flask` at that point would be incorrect, you'd end up in the inner `flask` package.  `pip install -e flask` is definitely wrong, since that would install the latest distribution in editable mode.  I don't think this needs to change.

PR #1669, opened 2015-12-30
Update `tox` installation instructions to point to PyPI

jeffwidman:
No longer need to grab `tox` from master


untitaker:
Thanks!

PR #1676, opened 2016-01-02
Refactored method Flask.make_response to improve readability

AvivC:
The method now looks at the type of the object returned from the view function, and returns a `self.response_class` object based on that type.

The docstring for the method has also been improved and contains a more
detailed explanation for what can be returned from a view function.

Also refactored the test for response creation.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1672,11 +1672,11 @@ class Flask(_PackageBoundObject):
         """
         return False
 
-    def make_response(self, rv):
+    def make_response(self, view_func_rv):
         """Converts the return value from a view function to a real
         response object that is an instance of :attr:`response_class`.
 
-        The following types are allowed for `rv`:
+        The following types are allowed for `view_func_rv`:
 
         .. tabularcolumns:: |p{3.5cm}|p{9.5cm}|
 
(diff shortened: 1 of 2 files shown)


flask/app.py
```diff
@@ -1688,51 +1688,79 @@ def make_response(self, rv):
...
-            else:
-                rv.status_code = status_or_headers
-        if headers:
-            rv.headers.extend(headers)
+    def _make_response_from_tuple(self, retval_from_view_func):
+        response_body, status, headers = self._extract_response_parts_from_tuple(retval_from_view_func)
 
-        return rv
+        if isinstance(response_body, self.response_class):
+            status, headers  \
```
ThiefMaster: I think the backslash usually comes after the `=` and with only one space before it

flask/app.py
```diff
@@ -1688,51 +1688,79 @@ def make_response(self, rv):
...
+
+        second_param_is_headers = isinstance(retval_from_view_func[1], (list, dict))
+        if second_param_is_headers:
+            headers, status = retval_from_view_func[1], None
+        else:
+            headers, status = [], retval_from_view_func[1]
+
+        return response_body, status, headers
+
+    def _possibly_use_status_and_headers_that_came_with_response_object(self, response_body, status, headers):
```
ThiefMaster: That method name sounds *way* too verbose

AvivC:
The build seems to fail on Python 3. Any idea why?

flask/app.py
AvivC: Will fix that.

flask/app.py
AvivC: Will fix.

tests/test_basic.py
```diff
@@ -908,53 +908,76 @@ def from_unicode():
...
+    assert response4.headers['X-Foo'] == 'Bar'
+    assert response4.headers['X-Bar'] == 'Foo'
+    assert response4.status == '404 NOT FOUND'
+    assert response4.status_code == 404
+
+    try:
+        client.get('/tuple5')
+    except ValueError as e:
+        assert 'Tuples returned from view functions must be composed as one of the following options'\
+               in str(e)
```
ThiefMaster: `with pytest.raises as exc_info:` might be more appropriate

flask/app.py
```diff
@@ -1672,11 +1672,11 @@ def should_ignore_error(self, error):
         """
         return False
 
-    def make_response(self, rv):
+    def make_response(self, retval_from_view_func):
```
ThiefMaster: `view_func_rv`? Not a huge fan of that either but "from" in a variable name usually reeks of too much verbosity

AvivC: Agree

AvivC:
In this project should I *always* prefix ascii strings with `b`? The build seems to fail because I didn't.

tests/test_basic.py
AvivC: For some reason using `pytest.raises` lets the exception propagate and fail the test. So I had to use `try except`.

AvivC:
I fixed the test where I previously used `try .. except: assert` instead of `pytest.raises`. The problem was that I forgot to do `app.testing = True`. Otherwise exceptions aren't propagated to the test client, so the `pytest.raises` assertion failes. Now fixed.

AvivC:
@ThiefMaster Hey :) Any other improvements to this pull request?

AvivC:
@guyskk 

> the origin code rv, status_or_headers, headers = rv + (None,) * (3 - len(rv)) is better, shorter and cleaner.

I respectfully disagree. I think using `status_or_headers` makes the code very hard to follow. At any point it may hold different types depending on the input, and It is also used as a sort of flag throughout the function to infer what was passed in from the view function. It's very difficult to understand all the possible paths in the code.

IMHO my refactoring turns the function into a simple series of `if - elif - else` clauses. The logic is now far clearer.

guyskk:
the origin rv tuple
```
response,
response, status
response, headers
response, status,  headers
```

after append None
```
response, None,    None
response, status,  None
response, headers, None
response, status,  headers
```

so the 3 items is
 `response`, `status_or_headers `,`headers`

AvivC:
@guyskk Yes, I understand the idea in `rv, status_or_headers, headers = rv + (None,) * (3 - len(rv))`.

I just think the same thing can be achieved with a little more code and far better readability. As demonstrated in the PR.

IMHO this function could be implemented simply with a series of `elif`s, so I don't see why not :)

guyskk:
Your implement of _make_response_from_tuple is too verbose, you need not split it into some functions, it just a little bit of code and you make it complicated. 

What's more, you should not change code and tests in the same commit, can your refactor pass the origin tests?

The 2th item can be string(status) or int(statuc_code), I wonder you refactor changed this behavior.

ThiefMaster:
> What's more, you should not change code and tests in the same commit

Why? By doing so you have working code with passing tests in each commit.

guyskk:
this is a refactor other than new feature，so the behavior shouldn't change，and should pass the origin tests.

AvivC:
@guyskk I agree, I might have refactored into too many sub-functions.

Other than that, IMHO the refactoring is good because it gives the function a structure of simple `if - elif` statements, instead of the weird structure it had before.

The 'rules' governing the behavior of the function are now visible to the eye, as opposed to the way the original is structured.

However as I said, I think you are right that I refactored into too many function levels.

@ThiefMaster Should I update the PR to not split into many sub-functions?

AvivC:
I have simplified the PR.

`make_response` is now a simple chain of `if - elif` statements. It calls a single function, `_make_response_from_tuple`. This function is a single block of code and doesn't call any other functions.

If you look at `make_response`, I think you'll agree it's clean and understandable.

    def make_response(self, view_func_rv):
        """... docs ..."""
        if view_func_rv is None:
            raise ValueError('View function did not return a response')

        if isinstance(view_func_rv, self.response_class):
            return view_func_rv
        elif isinstance(view_func_rv, (text_type, bytes, bytearray)):
            return self.response_class(view_func_rv)
        elif isinstance(view_func_rv, tuple):
            return self._make_response_from_tuple(view_func_rv)
        else:
            return self.response_class.force_type(view_func_rv, request.environ)

guyskk:
Good job!

guyskk:
But the tests `tuple1` `tuple2` ... didn't changed, the numbers is meaningless and has bad taste.

I suggest that split into 2 commit, the first one just refactor code, don't touch tests, so that we can confirm that the refactor didn't change the origin behavior. the second one improve tests.

davidism:
Thanks for looking into this. I agree that it was hard to follow. However, this introduced a bug where `None` was accepted as a response in a tuple, and the response in the tuple couldn't be a WSGI callable (although that's probably beyond rare).

I expanded on your idea by breaking out the tuple unpacking, adding more error messages, improving the documentation, and adding more tests. See #2256.

PR #1678, opened 2016-01-02
Refactored method preprocess_request in Flask.

AvivC:
Code duplication was eliminated. The docstring was also improved.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1791,31 +1791,30 @@ class Flask(_PackageBoundObject):
         raise error
 
     def preprocess_request(self):
-        """Called before the actual request dispatching and will
-        call each :meth:`before_request` decorated function, passing no
-        arguments.
-        If any of these functions returns a value, it's handled as
-        if it was the return value from the view and further
-        request handling is stopped.
+        """Called before the request dispatching.
+        Triggers two set of hook functions that should be invoked prior to a request dispatching:
+        self.url_value_preprocessors and self.before_request_funcs.
+        In both cases, triggers only the functions that are either global or registered to the current blueprint.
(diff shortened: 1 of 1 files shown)


davidism:
I'm on board with improving the docs, but I don't see any reason this needs to be split into four functions.

AvivC:
@davidism Agree. I removed the two extra functions. The PR now simply eliminates code duplication.

davidism:
Two for loops is not code duplication, and this function is unlikely to change.  Documentation only is fine.

flask/app.py
```diff
@@ -1791,32 +1791,39 @@ def handle_url_build_error(self, error, endpoint, values):
...
-        This also triggers the :meth:`url_value_processor` functions before
-        the actual :meth:`before_request` functions are called.
+        If any of the before_request_funcs returns a value, it's handled as if it was
+        the return value from the view function, and further request handling is stopped.
         """
-        bp = _request_ctx_stack.top.request.blueprint
-
-        funcs = self.url_value_preprocessors.get(None, ())
-        if bp is not None and bp in self.url_value_preprocessors:
-            funcs = chain(funcs, self.url_value_preprocessors[bp])
```
AvivC: @davidism The duplication I was referring to isn't the for loops, but the three lines of code that come before each one.
They are exactly identical except for the variables, so IMO it makes sense to factor them out to a function.

davidism:
Thanks, but I don't think this needs to be refactored.  It doesn't improve anything.  If you'd like to resubmit the documentation clarification, that's fine.

AvivC:
I will :)

PR #1679, opened 2016-01-02
Werkzeug should not block propagated exceptions from Flask

### LINKED ISSUE #1674: When using an external debugger, exceptions are not properly propagated
In trying to set up a small Flask application to work with an external debugger, I noticed that the way exceptions are handled vs. propagated isn't working as documented in the [Working with Debuggers](http://flask.pocoo.org/docs/0.10/errorhandling/#working-with-debuggers) section of the documentation.

The documentation specifies that to use an external debugger you have to play with the `debug`, `use_debugger` and `use_reloader` settings. In the following table I show all possible combinations of `debug` and `use_debugger` and what the current behavior is (Note that I'm assuming `use_reloader` is always set to `False` for this discussion, as this setting is not part of the problem):

    | debug | use_debugger | Who catches exceptions | Correct behavior?               |
    +-------+--------------+------------------------+---------------------------------+
    | True  | True         | Werkzeug's debugger    | Yes                             |
    | False | True         | Flask, returns 500     | Yes (I think?)                  |
    | True  | False        | Werkzeug's WSGI server | No, exceptions should propagate |
    | False | False        | Flask, returns 500     | Yes                             |

The appropriate configuration for using an external debugger is the one shown in the 3rd row of the above table, but note that with this configuration, exceptions are not propagated all the way up and never reach an external debugger. With this configuration Flask does the right thing and reraises the exception, but the Werkzeug WSGI server then catches it and handles it by logging the stack trace and moving on, because it was not told otherwise.

For this set up to work as documented, Werkzeug's server should have `passthrough_errors` set to `True`. If I add this setting to `app.run()` then exceptions are allowed to reach an external debugger.

My impression is that the value of the `PROPAGATE_EXCEPTIONS` configuration item should be synchronized with Werkzeug's `passthrough_errors`, so that Flask and Werkzeug consistently handle or reraise exceptions.

I'm happy to submit a pull request with the fix (which is quite simple), if you accept this as a valid issue. Please let me know what you think.

miguelgrinberg:
Fixes #1674.

By having Werkzeug always configured to let exceptions through we achieve the correct behavior for Flask. If Flask is configured to propagate exceptions (implicitly or explicitly, doesn't matter), then Werkzeug will get these exceptions and pass them through as well. If Flask is configured to not propagate exceptions, then Werkzeug will never see them, so no harm done if it is configured to pass them.

It looks like there is no testing for the new cli, so I have only added a test for the `app.run()` way of starting a server.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1222,6 +1222,26 @@ def test_exception_propagation():
         t.join()
 
 
+@pytest.mark.parametrize('debug', [True, False])
+@pytest.mark.parametrize('use_debugger', [True, False])
+@pytest.mark.parametrize('use_reloader', [True, False])
+@pytest.mark.parametrize('propagate_exceptions', [None, True, False])
+def test_werkzeug_passthrough_errors(monkeypatch, debug, use_debugger,
+                                     use_reloader, propagate_exceptions):
+    rv = {}
+
+    # Mocks werkzeug.serving.run_simple method
+    def run_simple_mock(*args, **kwargs):
+        rv['passthrough_errors'] = kwargs.get('passthrough_errors')
+
+    app = flask.Flask(__name__)
+    monkeypatch.setattr(werkzeug.serving, 'run_simple', run_simple_mock)
+    app.config['PROPAGATE_EXCEPTIONS'] = propagate_exceptions
+    app.run(debug=debug, use_debugger=use_debugger, use_reloader=use_reloader)
+    # make sure werkzeug always passes errors through
+    assert rv['passthrough_errors']
+
+
 def test_max_content_length():
     app = flask.Flask(__name__)
     app.config['MAX_CONTENT_LENGTH'] = 64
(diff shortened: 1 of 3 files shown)


tests/test_basic.py
```diff
@@ -1222,6 +1222,26 @@ def index():
...
+def test_werkzeug_passthrough_errors(monkeypatch):
+    rv = {}
+
+    # Mocks werkzeug.serving.run_simple method
+    def run_simple_mock(*args, **kwargs):
+        rv['passthrough_errors'] = kwargs.get('passthrough_errors')
+
+    app = flask.Flask(__name__)
+    monkeypatch.setattr(werkzeug.serving, 'run_simple', run_simple_mock)
+    for debug in [True, False]:
```
ThiefMaster: woultn't `pytest.mark.parametrize` be much better for this?

miguelgrinberg: Yep, makes sense. I'm not super familiar with py.test, I wasn't aware you can do that. I'll change it.

jeffwidman:
:+1:

davidism:
I don't know why I never noticed that PyCharm wasn't breaking on exceptions, but it works as expected and as described in the docs now.  Thanks for fixing this.

PR #1684, opened 2016-01-04
Clarified the docstring in method Flask.preprocess_request.

AvivC:
The doc now clearly states that the method invokes two set of hook
functions, and how these are filtered before execution.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1791,16 +1791,18 @@ class Flask(_PackageBoundObject):
         raise error
 
     def preprocess_request(self):
-        """Called before the actual request dispatching and will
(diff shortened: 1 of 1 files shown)


AvivC:
Respectfully ping :)

joequery:
Is this PR stating the current docs are out of date or incorrect? I find the previous docs more concise and easier to follow. 

For example, when the new version of the doc says the hooks "should be invoked", does that mean the dev has the responsibility of invoking it?

davidism:
Thanks for looking into this. I found the wording a little verbose and awkward, so I improved it in #2258 as well as making some changes to related docs.

PR #1696, opened 2016-01-16
Remove copyright year

pra85:
As per the discussion on 33534bb4a9937e6faba5ec


davidism:
Closing this in favor of setting the year to the earliest year specific to each file, and the doc year to a range.

PR #1698, opened 2016-01-20
Use indexation instead of slicing for getting the first element.

StyXman:
It's less confusing and it's slightly faster:
```sh
$ python2.7 -m timeit -s 'foo="a"*1024' 'foo[:1]'
10000000 loops, best of 3: 0.0352 usec per loop
$ python2.7 -m timeit -s 'foo="a"*1024' 'foo[0]'
10000000 loops, best of 3: 0.0291 usec per loop
```
```sh
$ python3.4 -m timeit -s 'foo="a"*1024' 'foo[:1]'
10000000 loops, best of 3: 0.0473 usec per loop
$ python3.4 -m timeit -s 'foo="a"*1024' 'foo[0]'
10000000 loops, best of 3: 0.0302 usec per loop
```

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -270,7 +270,7 @@ def url_for(endpoint, **values):
         url_adapter = reqctx.url_adapter
         blueprint_name = request.blueprint
         if not reqctx.request._is_old_module:
-            if endpoint[:1] == '.':
+            if endpoint[0] == '.':
                 if blueprint_name is not None:
                     endpoint = blueprint_name + endpoint
                 else:


cariaso:
If endpoints was an empty string, I think you'd be throwing an IndexError with this change, which wasn't true of the previous code. Not sure if it's realistic to ever have that, nor if the test cases check that.

StyXman:
Hmm, true. Given it's the only occurrence of this, I would add a comment about it.

ThiefMaster:
`url_for()` with an empty first argument seems pretty pointless, but I guess without the IndexError you get a much nicer error when calling it with an empty first argument.

Anyway, I don't think the time difference is really relevant. You won't call `url_for()` often enough (even in a large application) to make a difference of less than 20ns relevant

StyXman:
Agreed, with this change you get a `IndexError` instead of a `BuildError`, and preemptively raising `BuildError` would sidestep the whole error handling. I'm closing this.

PR #1702, opened 2016-01-25
Fix PEP8 issues and enforce PEP8 via pytest-pep8 (i.e. also during CI)

ThiefMaster:
E301 (no empty line before function/class) is disabled for files where having compact code (e.g. non-decorator based properties) makes sense.

### DIFF
--- a/flask/__init__.py
+++ b/flask/__init__.py
@@ -20,21 +20,22 @@ from jinja2 import Markup, escape
 
 from .app import Flask, Request, Response
 from .config import Config
-from .helpers import url_for, flash, send_file, send_from_directory, \
-     get_flashed_messages, get_template_attribute, make_response, safe_join, \
-     stream_with_context
-from .globals import current_app, g, request, session, _request_ctx_stack, \
-     _app_ctx_stack
-from .ctx import has_request_context, has_app_context, \
-     after_this_request, copy_current_request_context
+from .helpers import (url_for, flash, send_file, send_from_directory,
+                      get_flashed_messages, get_template_attribute,
+                      make_response, safe_join, stream_with_context)
+from .globals import (current_app, g, request, session, _request_ctx_stack,
+                      _app_ctx_stack)
+from .ctx import (has_request_context, has_app_context, after_this_request,
+                  copy_current_request_context)
 from .blueprints import Blueprint
 from .templating import render_template, render_template_string
 
 # the signals
-from .signals import signals_available, template_rendered, request_started, \
-     request_finished, got_request_exception, request_tearing_down, \
-     appcontext_tearing_down, appcontext_pushed, \
-     appcontext_popped, message_flashed, before_render_template
+from .signals import (signals_available, template_rendered, request_started,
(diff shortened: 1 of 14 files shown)


untitaker:
Interesting, why not flake8?

ThiefMaster:
py.test integration (even though it probably could have been done via tox too)

untitaker:
We should switch to one solution uniformly because otherwise contributors have
to follow different guidelines for different projects.

Atm Werkzeug uses flake8 in tox (with very weak settings)

ThiefMaster:
ok i'll have a look and use flake8 here too

ThiefMaster:
hm.. with flake8 there's no way of disabling certain checks per-file

davidism:
I don't like the formatting where there is major indentation on all lines, everything gets very crowded at the right.  Instead, move the first argument to a new line as well and use normal wrapping rules.  The automatic wrapping detection in tools like PyCharm's formatter produces some weird lines, so manually verify that the lines actually look good.

~~~python
# from
permanent_session_lifetime = ConfigAttribute('PERMANENT_SESSION_LIFETIME',
                                             get_converter=_make_timedelta)
# to
permanent_session_lifetime = ConfigAttribute(
    'PERMANENT_SESSION_LIFETIME',
    get_converter=_make_timedelta
)
~~~

Also, can we disable the line length, or at least extend the it the modern guideline of 100 characters?  I've never been a fan of the 80 character hard limit, and it would alleviate the wrapping problem described above.

There are still some instances of `\` continuation.

ThiefMaster:
If we go to >79 (which I'd highly appreciate, IMO the default limit results in less readable code), why not set 120 as a maximum for code and ~70 as a soft limit for comments/docstrings (they are more readable, but wrapping a long word because of 1 or 2 characters is pretty ugly)?

davidism:
I'd rather have just one 120 character limit for everything.  But that might be a little to extreme.

ThiefMaster:
@mitsuhiko what's your opinion on max-line-length=120 (or 100 in case you don't like 120)?

jeffwidman:
It'd be great to see this merged. @mitsuhiko any input on max line length?

untitaker:
I'm unsure if this is the proper way to enforce PEP-8 style. Maybe it should be gradually improved, to avoid massive merge conflicts between maintenance branches (and PRs)?

JelteF:
Although Flake8 does not have native exclusion different filetypes it can be done using the flake8-putty extension:
https://github.com/jayvdb/flake8-putty

davidism:
Closing this due to large number of merge conflicts. Unless we can close all other outstanding PRs first, I don't think a blanket change is the way to go, as it has a high potential of causing merge conflicts. Strategy I see going forward:

* Do some GitHub / git magic and figure out what files are not part of any open PR, and fix those.
* Add a tox test that filters flake8's output based on the lines git reports as changed and only fails if changed lines have lint issues. See [git-lint](https://github.com/sk-/git-lint/), for example, although it hasn't been updated recently and does not support flake8.

PR #1711, opened 2016-01-30
terminate running port by application in case of exit process

erkanay:
I have noticed that it would be better to have such a simple tool as decorator in code base to terminate port running by app.

A simple example:

```
from flask import Flask

app = Flask(__name__)

@app.reboot()
@app.route("/")
def hello():
    return "Hello world"

if __name__ == "__main__":
    app.run()
```

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -779,6 +779,30 @@ class Flask(_PackageBoundObject):
             rv.update(processor())
         return rv
 
+    def reboot(self):
+        """A decorator that is used to terminate running port
+        by application.
+
+            @app.reboot()
+            @app.route('/')
+            def index():
+                return 'Hello World'
+
+        """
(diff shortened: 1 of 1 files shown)


pussbb:
Simpler solution is here
http://flask.pocoo.org/snippets/67/

ThiefMaster:
You usually use the reloader during development so you don't need a manual restart. Also, a decorator makes not much sense for this. The Python code quality is also pretty low. Import should not be local, `commands.getoutput` most likely  uses a shell instead of an argv-style list and calling `kill` from Python makes no sense since Python has `os.kill()` for this.

All in all, this belongs not in Flask. Why not do something like this locally if you need it?

erkanay:
Thanks for update. I am going to close it.

PR #1716, opened 2016-02-03
Fix typo

lord63:
* Use the compatible way to handle the exception. You can find the
source code wsgi_app in app.py, and it use the compatible way, so update it
* Fix typo in config.py
* Fix typo in app.py

---

One more question, in the [flask.Flask.run](http://flask.pocoo.org/docs/dev/api/#flask.Flask.run) documentation, it says:

> It is not recommended to use this function for development with automatic reloading as this is badly supported. Instead you should be using the flask command line script’s runserver support.

I guess it should be:

> ...the flask command line script’s run support.

since flask cli has the `run` command not the `runserver` command.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -421,7 +421,7 @@ class Flask(_PackageBoundObject):
         #: A dictionary with lists of functions that should be called after
         #: each request.  The key of the dictionary is the name of the blueprint
         #: this function is active for, ``None`` for all requests.  This can for
-        #: example be used to open database connections or getting hold of the
+        #: example be used to close database connections or getting hold of the
         #: currently logged in user.  To register a function here, use the
         #: :meth:`after_request` decorator.
         self.after_request_funcs = {}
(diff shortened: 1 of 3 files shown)


flask/app.py
```diff
@@ -421,7 +421,7 @@ def __init__(self, import_name, static_path=None, static_url_path=None,
         #: A dictionary with lists of functions that should be called after
         #: each request.  The key of the dictionary is the name of the blueprint
         #: this function is active for, ``None`` for all requests.  This can for
-        #: example be used to open database connections or getting hold of the
+        #: example be used to close database connections or getting hold of the
         #: currently logged in user.  To register a function here, use the
         #: :meth:`after_request` decorator.
```
davidism: The mention of current user doesn't really make sense here either.  Maybe just change it to "close database connections and other open resources"?  Or just remove the second part entirely.

lord63: Although I'm OK with the original way, still can't come out a good example to show the second part. I've remove it.

I'll squash the commits when ready to merge.

kennethreitz:
:sparkles: :cake: :sparkles:

PR #1720, opened 2016-02-07
Non developer readme contribution

EricLOsorio:
I added a novice introduction to expand on some of the concepts that a novice might need to start using flask


ThiefMaster:
- Way too many commits that clutter the history (squash them into a single one)
- Unreadably-long lines (plaintext line should rarely exceed ~80 chars)

Besides that, I'm not sure if it makes sense to make the readme file that verbose. It sounds more like information that belongs into RTD (and I guess many of it is already in the flask docs)

PR #1728, opened 2016-02-16
Add JSONIFY_MIMETYPE configuration variable

sloria:
It is often useful to be able to return a vendor media type for JSON responses. For example, [JSON API](http://jsonapi.org/format/#content-negotiation) requires that all responses have the `Content-Type: applications/vnd.api+json` header.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -183,6 +183,7 @@ The following configuration values are used internally by Flask:
                                   if they are not requested by an
                                   XMLHttpRequest object (controlled by
                                   the ``X-Requested-With`` header)
+``JSONIFY_MIMETYPE``              MIME type used for jsonify responses.
 ``TEMPLATES_AUTO_RELOAD``         Whether to check for modifications of
                                   the template source and reload it
                                   automatically. By default the value is
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1019,6 +1019,18 @@ def test_jsonify_prettyprint():
         assert rv.data == pretty_response
 
 
+def test_jsonify_mimetype():
+    app = flask.Flask(__name__)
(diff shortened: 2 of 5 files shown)


tests/test_basic.py
```diff
@@ -1029,6 +1029,18 @@ def test_jsonify_prettyprint():
         assert rv.data == pretty_response
 
 
+def test_jsonify_mimetype():
+    app = flask.Flask(__name__)
+    app.config.update({"JSONIFY_MIMETYPE": 'application/vnd.api+json'})
```
ThiefMaster: why not just `app.config['JSONIFY_MIMETYPE'] = '...'`?

docs/config.rst
```diff
@@ -183,6 +183,7 @@ The following configuration values are used internally by Flask:
                                   if they are not requested by an
                                   XMLHttpRequest object (controlled by
                                   the ``X-Requested-With`` header)
+``JSONIFY_MIMETYPE``              Mimetype used for jsonify responses.
```
ThiefMaster: "MIME type"?

tests/test_basic.py
sloria: I am following the conventions used in similar tests. For example: https://github.com/mitsuhiko/flask/pull/1728/files#diff-9e9ddedef65dec3da86063a067839e65R1021

docs/config.rst
sloria: Good call. Fixed.

jeffwidman:
LGTM

untitaker:
LGTM, but the changelog entry is now for the wrong version.

sloria:
OK, I've rebased with the latest `master` and put the changelog entry in the 1.0 section.

PR #1730, opened 2016-02-17
make safe_join behave like os.path.join with *args

geusebi:
Make safe_join behave like os.path.join, which accept multiple path components with *args.
Note that only the ``directory'' 's bound is respected and the path components are only joined together.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,8 @@ Version 0.12
 - the cli command now responds to `--version`.
 - Mimetype guessing for ``send_file`` has been removed, as per issue ``#104``.
   See pull request ``#1849``.
+- Make ``flask.safe_join`` able to join multiple paths like ``os.path.join``
+  (pull request ``#1730``).
 
 Version 0.11.1
 --------------
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -563,8 +563,9 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
     return rv
 
 
-def safe_join(directory, filename):
-    """Safely join `directory` and `filename`.
+def safe_join(directory, *pathnames):
+    """Safely join `directory` and zero or more untrusted `pathnames`
+    components.
 
     Example usage::
 
@@ -574,20 +575,23 @@ def safe_join(directory, filename):
             with open(filename, 'rb') as fd:
                 content = fd.read()  # Read and process the file content...
 
-    :param directory: the base directory.
-    :param filename: the untrusted filename relative to that directory.
-    :raises: :class:`~werkzeug.exceptions.NotFound` if the resulting path
-             would fall out of `directory`.
+    :param directory: the trusted base directory.
+    :param pathnames: the untrusted pathnames relative to that directory.
+    :raises: :class:`~werkzeug.exceptions.NotFound` if one or more passed
+            paths fall out of its boundaries.
     """
-    filename = posixpath.normpath(filename)
-    for sep in _os_alt_seps:
-        if sep in filename:
+    for filename in pathnames:
+        if filename != '':
+            filename = posixpath.normpath(filename)
+        for sep in _os_alt_seps:
+            if sep in filename:
+                raise NotFound()
+        if os.path.isabs(filename) or \
+           filename == '..' or \
+           filename.startswith('../'):
             raise NotFound()
-    if os.path.isabs(filename) or \
-       filename == '..' or \
-       filename.startswith('../'):
-        raise NotFound()
-    return os.path.join(directory, filename)
+        directory = os.path.join(directory, filename)
+    return directory
 
 
 def send_from_directory(directory, filename, **options):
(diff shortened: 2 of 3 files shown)


untitaker:
Is there a concrete usecase for this? I would expect this to cause confusion as it's no longer clear which components are trusted.

geusebi:
I often use an another version which protect each single path components from crossing its boundary, depend on the case.

I issued a pull request with this version because it was the easier one.

e.g. (with base_dir coming from config and the other variables from the user)
`safe_join(base_dir, users_dir, user_name, "some_file")`
whatever the user input it cannot leave base_dir.

Checking every single path would be even better for me as it's the version I use most.

untitaker:
I still don't understand which usecase this has.

geusebi:
If you have many path to join you can do:
`safe_join(base_dir, users_dir, user_name, "some_file")`
or:
`paths = (base_dir, users_dir, user_name, "some_file")`
`safe_join(*paths)`
instead of:
`safe_join(base_dir, safe_join(users_dir, safe_join(user_name, "some_file")))`

Also you can switch from os.path.join to safe_join without problems even if you use *args.
I've always used os.path.join this way and I missed that feature when I used safe_join for the first time.

timster:
Couldn't you just do

`safe_join(base_dir, os.path.join(users_dir, user_name, "some_file"))`

geusebi:
Sure, but it would be nice to have safe_join to accept arguments exactly as its counterpart (I'm assuming it's os.path.join(a, *p)). It's easy to implement this behavior without breaking anything and so I proposed the pull request.

If you see no value or no need to merge, no problem :)

A version which checks every path components could be something like this
https://gist.github.com/geusebi/b152f1ee17a66a9bbeee (not tested)
With this version 
`safe_join(base_dir, safe_join(users_dir, safe_join(user_name, "some_file")))`
could be written as I did in my last comment i.e.
`paths = (base_dir, users_dir, user_name, "some_file")`
`safe_join(*paths)`

untitaker:
You've still not given me a concrete usecase. I'm worried that the proposed API isn't secure in situations where it's assumed to be.

On 22 February 2016 08:09:17 CET, Giampaolo Eusebi <notifications@github.com> wrote:
>Sure, but it would be nice to have safe_join to accept arguments
>exactly as its counterpart (I'm assuming it's os.path.join(a, *p)).
>It's easy to implement this behavior without breaking anything and so I
>proposed the pull request.
>
>If you see no value or no need to merge, no problem :)
>
>A version which checks every path components could be something like
>this
>https://gist.github.com/geusebi/b152f1ee17a66a9bbeee (not tested)
>With this version 
>`safe_join(base_dir, safe_join(users_dir, safe_join(user_name,
>"some_file")))`
>could be written as I did in my last comment i.e.
>`paths = (base_dir, users_dir, user_name, "some_file")`
>`safe_join(*paths)`
>
>---
>https://github.com/mitsuhiko/flask/pull/1730#issuecomment-187044768

-- 

ThiefMaster:
I think for security-related functions it makes sense to keep them simple. Extra complexity just makes it more likely that someone uses them in an insecure fashion

untitaker:
To be clear, I'd rather see a behavior where only the basepath is trusted, and the other args are treated as potentially malicious.

On 22 February 2016 10:27:16 CET, Adrian <notifications@github.com> wrote:
>I think for security-related functions it makes sense to keep them
>simple. Extra complexity just makes it more likely that someone uses
>them in an insecure fashion
>
>---
>https://github.com/mitsuhiko/flask/pull/1730#issuecomment-187090518

-- 

geusebi:
Here only the basepath is trusted https://gist.github.com/geusebi/b152f1ee17a66a9bbeee .

Usecase: sometimes I happen to manipulate paths and I have to collect some of these paths from the user and from other untrusted sources. In the end I have a tuple or a list of paths to join, safe_join allow me to do it safely but I have to call it using reduce or with a loop to make it work with multiple paths. This puzzled me the first time as I presumed it was fully compatible with path.join.

If I am the only one which thinks it can be useful and if so many (very reasonable) doubts are coming from this simple addition it should probably be refused :)

untitaker:
I'll merge if you update the PR from the gist you posted and add a changelog.

(Also you use a generator expression and a loop in the gist, it could be only one)

joequery:
:+1: from me for this PR. With the current behavior of only allowing a single file to be joined to the basepath, we're increasing the likelihood that someone using this function has manually constructed their own insecure "basepath" (which wouldn't truly be a basepath at that point) to accommodate the function definition.

untitaker:
This PR is insecure as well. It treats `*pathnames` as "trusted" in the sense that it allows `..` in one `pathnames` part to eradicate the previous one.

https://gist.github.com/geusebi/b152f1ee17a66a9bbeee is the correct approach.

untitaker:
@geusebi Could you rebase and add tests? Thanks!

geusebi:
Rebased and added some tests, let me know if you need something else.

P.S.
It would be great if somebody else could take a double look at the code before the merge.

Many thanks

untitaker:
Excellent, thanks!

PR #1733, opened 2016-02-19
Update MANIFEST.in with simpler template commands

welchbj:
Please see https://docs.python.org/2/distutils/sourcedist.html#commands for reference.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,13 +1,11 @@
 include Makefile CHANGES LICENSE AUTHORS
-recursive-include artwork *
-recursive-include tests *
-recursive-include examples *
-recursive-include docs *
-recursive-exclude docs *.pyc
-recursive-exclude docs *.pyo
-recursive-exclude tests *.pyc
-recursive-exclude tests *.pyo
-recursive-exclude examples *.pyc
-recursive-exclude examples *.pyo
+
+graft artwork
+graft tests
+graft examples
+graft docs
+
+global-exclude *.py[co]
+
 prune docs/_build
 prune docs/_themes


untitaker:
Seems reasonable. Any idea in which version those commands were added?

welchbj:
Admittedly, the documentation on these commands is pretty poor. Looking at the distutils repo on BitBucket, though, showed that they have been around since very early on in the lifespan of distutils: https://bitbucket.org/carljm/python-distutils/src/5bacc3d7000d1ba1e686a66df2e700c90535a6c7/filelist.py?fileviewer=file-view-default#filelist.py-164

davidism:
Can you verify that the result of `python setup.py sdist` and `python setup.py bdist_wheel` contain the same files before and after this change?

welchbj:
This script yields for me that the same files are packaged for sdist. I would assume the same logic is used by setuptools for the bdist_wheel command, as well.
```
git clone https://github.com/mitsuhiko/flask.git

cd flask

python setup.py sdist --format=zip >/dev/null
echo ++++++++++++++++
echo + Old contents +
echo ++++++++++++++++
python -c "import glob; import zipfile; print('\n'.join(zipfile.ZipFile(glob.glob('dist/*')[0]).namelist()))"

rm -rf dist

git fetch origin pull/1733/head:pr1733
git checkout pr1733
python setup.py sdist --format=zip >/dev/null
echo ++++++++++++++++
echo + New contents +
echo ++++++++++++++++
python -c "import glob; import zipfile; print('\n'.join(zipfile.ZipFile(glob.glob('dist/*')[0]).namelist()))"
```

tuukkamustonen:
> I would assume the same logic is used by setuptools for the bdist_wheel command, as well.

Binary packages (`bdist_*`) don't read `MANIFEST.in` file, instead they rely on what you define in `setup(...)`.

welchbj:
Thanks for the insight @tuukkamustonen. So would you say the only command this change would affect is sdist?

tuukkamustonen:
@welchbj That's how it goes. You can build wheels without `MANIFEST.in` file and having one doesn't affect what goes in. I don't have reference for it, but that's my experience. So yeah, this PR should only change sdist packaging.

untitaker:
Thanks @welchbj!

PR #1739, opened 2016-03-01
Add encoding type parameter to fix 'UnicodeDecodeError: cp949'

liks79:
I have a problem with loading a utf-8 encoded config file using config.from_pyfile() function.

However,  It raised an following error message.

> UnicodeDecodeError: 'cp949' codec can't decode byte 0xec in position 99: illegal multibyte sequence.

So I add encoding parameter to solve the problem at from_pyfile() function.

> Test Environments;

- OS : MS Window 7 32bit - Korean Edition
- Python 3.5.1
- Flask 0.10.1

### DIFF
--- a/flask/config.py
+++ b/flask/config.py
@@ -108,7 +108,7 @@ class Config(dict):
                                variable_name)
         return self.from_pyfile(rv, silent=silent)
 
-    def from_pyfile(self, filename, silent=False):
+    def from_pyfile(self, filename, silent=False, encoding='utf-8'):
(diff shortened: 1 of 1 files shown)


liks79:
I'll fix it and pull request again. 
😅😅😅

ThiefMaster:
FYI you can just forcepush to a PR's branch instead of opening a new one

liks79:
@ThiefMaster Thanks for mentioning it. 😀

Issue #1740, opened 2016-03-03
Not able to set encoding in open_resource


JiriBalcar:
If the resource from the application's resource folder is encoded in different encoding than locale.getpreferredencoding the UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 217 occures. We should be able to set encoding on open_resource

Python 3.5.1, Flask 0.10.1

PR #1741, opened 2016-03-03
Fix for #1740

JiriBalcar:
#1740

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -873,7 +873,7 @@ class _PackageBoundObject(object):
         return send_from_directory(self.static_folder, filename,
                                    cache_timeout=cache_timeout)
 
-    def open_resource(self, resource, mode='rb'):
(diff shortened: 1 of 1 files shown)


JiriBalcar:
Added encoding to open_resource with default None so it is not breaking change

davidism:
Use [`codecs.open`](https://docs.python.org/3/library/codecs.html#codecs.open) for compatibility with 2 and 3.    Needs documentation and a test would be nice too.

davidism:
On second look, I can't figure out why this fails for you.  It's opening the file in binary mode, it shouldn't matter what the encoding is at that point.

JiriBalcar:
if you will open the file in 'r' mode instead of 'rb' it will fail. For now, I changed the opening mode to 'rb' and decoding it myself. I will take a look on codecs.open and will try to update this pull request, because I think it will look better if the encoding can be defined in open_resource.

PR #1742, opened 2016-03-04
make hostname configurable

ferstar:
make hostname configurable if it defined in the ``SERVER_NAME`` config

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -807,7 +807,8 @@ class Flask(_PackageBoundObject):
 
         :param host: the hostname to listen on. Set this to ``'0.0.0.0'`` to
                      have the server available externally as well. Defaults to
-                     ``'127.0.0.1'``.
+                     ``'127.0.0.1'`` or the hostname defined in the ``SERVER_NAME``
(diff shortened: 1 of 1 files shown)


untitaker:
Why?

You can do this yourself, after all you're the one calling `app.run`.

ferstar:
Yes, but maybe it would be convenient to read config from the configuration file :-)

ThiefMaster:
Why, it's just for the dev server anyway, you rarely want to run it on anything but localhost

ferstar:
@ThiefMaster Of course, I was actually tested on my own , and this `patch` is very simple.

untitaker:
I'm against this. There are infinite ways to fetch/compute the values for `host` and `port`, and all of them can already be computed before launching `app.run`.

ferstar:
@untitaker Got, Thank you for your explanation and I will change it myself. BTW, happy weekend! :+1:

untitaker:
Thanks for your understanding, and a happy weekend to you too. :)

PR #1747, opened 2016-03-08
Fixed stream_with_context not working when decorated function takes parameters

fengsp:


ThiefMaster:
The commit message is quite long. The first line of a Git commit message should usually be <=51 chars (if that's not enough, leave line 2 empty and put whatever verbose explanation is needed in the following lines)

*"Fix stream_with_context if function has arguments"* would fit for example

PR #1748, opened 2016-03-08
Update deprecated references

gaetan-petit:
validators.Required() is marked as deprecated in favor of validators.DataRequired() and will be remove in WTForms 3.


ThiefMaster:
:+1:

untitaker:
I have no idea how wtforms works but I guess @ThiefMaster knows what this is about. Thanks @gaetan-petit!

ThiefMaster:
There's `DataRequired` (non-falsy value after processed by the field, so it would not work for an integer field where 0 is allowed) and `InputRequired` (any value submitted, i.e. `'foo' in request.form`)

PR #1753, opened 2016-03-14
Add two missing converters for flask in the docs

lord63:
All the converters are from werkzeug's builtin converters.

Documentation: http://werkzeug.pocoo.org/docs/dev/routing/#builtin-converters


lord63:
Ping @davidism, is there any problems before merging this PR?

PR #1754, opened 2016-03-14
Fixed minor typo error

bdh1011:
remove ;


untitaker:
Not really a typo, but it could be nicer. Don't really know if this is better though. Are there any references that would invalidate this punctuation?

bdh1011:
I think it is awkward. Though semicolon can be used as a kind of supercomma, this sentence is not that kine of thing. Oh, replacing it with comma would be more better. isn't it?

davidism:
The comma would work better, but this is fine as is.  It's two instructions: navigate through a series of directories, then run a file.

PR #1755, opened 2016-03-14
Replace ';' to ','

bdh1011:
I think it is awkward. Though semicolon can be used as a kind of supercomma, this sentence is not that kine of thing. Replacing it with comma would be more better. isn't it?


untitaker:
:confetti_ball:

PR #1760, opened 2016-03-27
Run bdist_wheel as a part of release process

mattrobenolt:
The setup.cfg declares the ability to compile a wheel, but release process isn't actually compiling a wheel.


untitaker:
Oh wow, thanks! Probably similar bugs are in the release scripts of all dependencies of Flask.

PR #1762, opened 2016-03-30
helpers.py - Ensure file is closed if flask opens it

devdave:


devdave:
Ran into a problem of helpers.send_file not closing a file it had opened, made sure it closes it now.

PR #1777, opened 2016-04-08
The comment didn't make any sense

danielquinn:
I'm not sure that I've understand the purpose of this property, but on the off chance that I'm correct, this statement makes more sense than what came before.

### DIFF
--- a/flask/views.py
+++ b/flask/views.py
@@ -48,7 +48,7 @@ class View(object):
     generated view function!
     """
 
-    #: A list of which methods this view can handle.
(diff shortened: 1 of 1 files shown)


flask/views.py
```diff
@@ -48,7 +48,7 @@ def dispatch_request(self):
     generated view function!
     """
 
-    #: A for which methods this pluggable view can handle.
+    #: A list of which methods this view can handle.
```
ThiefMaster: -wtich

PR #1781, opened 2016-04-13
minor rewording of get_json documentation for clarity

miguelgrinberg:
I did a minor rewording of the docstring for `request.get_json()`. The current documentation is misleading, by having a mention of `on_json_loading_failed` followed by a sentence that references "this function", without clearly stating if the reference is to `get_json` or `on_json_loading_failed`.

The description reads much better in my opinion, after I moved the `on_json_loading_failed` sentence to the end.

### DIFF
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -123,11 +123,12 @@ class Request(RequestBase):
         return False
 
     def get_json(self, force=False, silent=False, cache=True):
-        """Parses the incoming JSON request data and returns it.  If
-        parsing fails the :meth:`on_json_loading_failed` method on the
-        request object will be invoked.  By default this function will
-        only load the json data if the mimetype is :mimetype:`application/json`
-        but this can be overridden by the `force` parameter.
+        """Parses the incoming JSON request data and returns it.  By default
+        this function will return ``None`` if the mimetype is not
+        :mimetype:`application/json` but this can be overridden by the
+        ``force`` parameter. If parsing fails the
+        :meth:`on_json_loading_failed` method on the request object will be
+        invoked.
 
         :param force: if set to ``True`` the mimetype is ignored.
         :param silent: if set to ``True`` this method will fail silently


davidism:
Should probably mention that "only load if" means "will return None otherwise".  Otherwise looks good.

miguelgrinberg:
The default behavior is to return a bad request error, actually. This is described in the docstring for the `on_json_loading_failed`, which is linked here. Do you think it needs to be repeated on this docstring?

davidism:
It's late, so I might be totally reading this wrong, but

~~~python
from flask import Flask, request

app = Flask(__name__)

@app.route('/')
def index():
    data = request.get_json()
    print(data)
    return 'hello'

app.run(debug=True)
~~~

prints None and doesn't raise an error, because force is False and the request isn't json.  If force were true, it would try to read the (non-existent) json and raise an error.

miguelgrinberg:
Okay, yes, you are right. I didn't actually test it, was going from the documentation, which is still misleading, so yeah, I need to find a better way to explain what's going on.

flask/wrappers.py
```diff
@@ -123,11 +123,11 @@ def is_json(self):
         return False
 
     def get_json(self, force=False, silent=False, cache=True):
-        """Parses the incoming JSON request data and returns it.  If
-        parsing fails the :meth:`on_json_loading_failed` method on the
-        request object will be invoked.  By default this function will
-        only load the json data if the mimetype is :mimetype:`application/json`
-        but this can be overridden by the `force` parameter.
+        """Parses the incoming JSON request data and returns it.  By default
+        this function will return `None` if the mimetype is not
```
davidism: Inline code uses two backticks in rst: \`\`None\`\` and force below.

miguelgrinberg:
Oops, right. The `force` mention has single backticks in the current docs, I just copied without thinking.

joequery:
:+1: on the clarification of this. I would personally phrase the docstring like so:

>  By default this function will immediately return ``None`` if the provided mimetype is not :mimetype:`application/json`. Use `force=True` to force parsing of the request data regardless of the mimetype

PR #1797, opened 2016-05-04
#1782: Add flag to handle file-like objects

g3wanghc:
A flag to deal with file-like object while keeping the `.name` and `.filename` properties.


ThiefMaster:
Haven't reviewed it closer yet, but those comments should most likely be squashed into a single one.

g3wanghc:
@ThiefMaster Should I make it into a single commit?

g3wanghc:
Need to work on a better solution.

PR #1798, opened 2016-05-05
Fix a grammar mistake in the docs

wojcikstefan:


untitaker:
Thanks!

thomasst:
You should also remove `themselves`, which is unnecessary :)

PR #1806, opened 2016-05-12
Forward ported CLI tests from Flask-CLI and fixed a bug with the CLI's name.

jezdez:
To clarify, looking at the [signature of the click's Group class](https://github.com/pallets/click/blob/6.6/click/core.py#L1135-L1138) I found that passing the whole Flask app isn't what it expects, but just a command group name.


untitaker:
Excellent, thanks!

PR #1810, opened 2016-05-15
Added X-Accel-Redirect support

deelaka:
`X-Accel-Redirect` is very much alike Apache's `XSendfile` it instead however needs a `URI` pointing to the file than an absolute path. This is a tricky problem to solve and there  have been discussions on implementing a `nginx_sendfile` method, this however is disastrous since in development machines `nginx_sendfile` won't send the file since it is on a default flask server, thus every time the developer wishes to test an app on his own machine, it is required to change all `nginx_sendfile` calls to normal `flask.send_file` calls manually changing url's and such.

How is this patch different? As seen below in the test app simply set the config  `USE_X_ACCEL` to true and then when calling `send_file` set the `x_accel_uri` parameter to the internal `uri` configured in your nginx config. And when testing simply keep `USE_X_ACCEL` at the default value `False`, this will make `send_file` ignore the `x_accel_uri` even if it is set. :smile: 

## A test app:

```
from flask import Flask, send_file

app = Flask(__name__)
app.use_x_accel = True  
# app.config['USE_X_ACCEL'] = True

@app.route('/dl')
def download():
        return send_file('files/1GB.zip', as_attachment=True, x_accel_uri='/protected/')

app.run(debug=True, port=8080)
```

## For the following nginx config:

```
server {
        listen 80;

        location / {
                proxy_pass http://localhost:8080/;
        }

        location /protected {
                internal;
                alias /home/test/files;
        }
}
```

### Headers retrieved using `Curl`:

```
deelaka@deelaka-Lap:~$ curl -I your.server.ip/dl
HTTP/1.1 200 OK
Server: nginx/1.4.6 (Ubuntu)
Date: Sun, 15 May 2016 08:42:49 GMT
Content-Type: application/zip
Content-Length: 1073741824
Last-Modified: Sat, 14 May 2016 16:51:22 GMT
Connection: keep-alive
Content-Disposition: attachment; filename=1GB.zip
Cache-Control: public, max-age=43200
Expires: Sun, 15 May 2016 20:42:49 GMT
ETag: "5737578a-40000000"
Accept-Ranges: bytes

```

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -266,6 +266,15 @@ class Flask(_PackageBoundObject):
     #: ``USE_X_SENDFILE`` configuration key.  Defaults to ``False``.
     use_x_sendfile = ConfigAttribute('USE_X_SENDFILE')
 
+    #: Enable this if you want to use Nginx's X-Accel feature.
+    #: This only affects files sent with the :func:`send_file` method.
+    #:
+    #: .. versionadded:: 0.10
+    #:
+    #: This attribute can also be configured from the config with the
+    #: ``USE_X_ACCEL`` configuration key.  Defaults to ``False``.
+    use_x_accel = ConfigAttribute('USE_X_ACCEL')
+
     #: The name of the logger to use.  By default the logger name is the
     #: package name passed to the constructor.
     #:
@@ -296,6 +305,7 @@ class Flask(_PackageBoundObject):
         'SECRET_KEY':                           None,
         'PERMANENT_SESSION_LIFETIME':           timedelta(days=31),
         'USE_X_SENDFILE':                       False,
+        'USE_X_ACCEL':                          False, # kula
         'LOGGER_NAME':                          None,
         'LOGGER_HANDLER_POLICY':               'always',
         'SERVER_NAME':                          None,
--- a/flask/app.py
+++ b/flask/app.py
@@ -305,7 +305,7 @@ class Flask(_PackageBoundObject):
         'SECRET_KEY':                           None,
         'PERMANENT_SESSION_LIFETIME':           timedelta(days=31),
         'USE_X_SENDFILE':                       False,
-        'USE_X_ACCEL':                          False, # kula
+        'USE_X_ACCEL':                          False,
         'LOGGER_NAME':                          None,
         'LOGGER_HANDLER_POLICY':               'always',
         'SERVER_NAME':                          None,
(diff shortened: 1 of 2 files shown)


ThiefMaster:
First of all, this kind of PR should be squashed into a single commit.

Then, I'm not sure if it makes lots of sense to put this inside Flask - but certainly not as a part of `send_file` which means you need to change your application code to support nginx/X-Accel-Redir.

This could be done much better as a WSGI middleware where you setup a mapping between xsendfile/filesystem paths and URLs:

```python
class XAccelMiddleware(object):
    """A WSGI Middleware that converts X-Sendfile headers to X-Accel-Redirect
    headers if possible.

    If the path is not mapped to a URI usable for X-Sendfile we abort with an
    error since it likely means there is a misconfiguration.
    """

    def __init__(self, app, mapping):
        self.app = app
        self.mapping = mapping.items()

    def __call__(self, environ, start_response):
        def _start_response(status, headers, exc_info=None):
            xsf_path = None
            new_headers = []
            for name, value in headers:
                if name.lower() == 'x-sendfile':
                    xsf_path = value
                else:
                    new_headers.append((name, value))
            if xsf_path:
                uri = self.make_x_accel_header(xsf_path)
                if not uri:
                    raise ValueError('Could not map %s to an URI' % xsf_path)
                new_headers.append(('X-Accel-Redirect', uri))
            return start_response(status, new_headers, exc_info)

        return self.app(environ, _start_response)

    def make_x_accel_header(self, path):
        for base, uri in self.mapping:
            if path.startswith(base + '/'):
                return uri + path[len(base):]
```

This doesn't even need changes in Flask, just one line in your application to add the middleware.

deelaka:
@ThiefMaster yes this should've been squashed but what I don't get is why you insist implementing X-accel support through WSGI middleware? is it for some sort of performance reason? and most importantly why has this so called "WSGI middleware for X-Accel" support not yet implemented in flask or werkzeug?

ThiefMaster:
I agree that it would make sense to add this middleware to Werkzeug or Flask; I simple never ended up doing it (since that also involves writing better documentation for it).

It's not really a performance thing but mostly about keeping the `send_file` API clean of webserver-specific code (`X-Sendfile` is much more generic since it just passes the filesystem path). Doing it on the WSGI level also makes sense since now it doesn't even need to know from where in the app the send-file request comes from - as soon as there's the header it'll handle it.

deelaka:
@ThiefMaster I agree, it is important to keep the code clean, and since that is a priority I'll close this pull request. But It'd be really helpful if you could provide insight on how to easily integrate your given `XAccelMiddleware` to a simple app such the `test app` I've posted above. :)

ThiefMaster:
app.wsgi_app = XAccelMiddleware(app.wsgi_app, app.config['X_ACCEL_MAP'])

Then simply add a dict called `X_ACCEL_MAP` to your config mapping filesystem paths to URLs as setup in the nginx config.

PR #1815, opened 2016-05-22
Add special handling of ./ template strings

JelteF:
As is shown in https://github.com/pallets/flask/issues/1361, there is a desire for easy separation of templates for different blueprints. 

Currently it is not possible to assign a separate sub directory to be used for a specific blueprint. All templates are searched relatively to the root template directory. Although it seems that the `template_folder` Blueprint argument would have this meaning this is not actually the case. 
This argument will only add the `template_folder` as one of the directories to search in. 

With this change a template passed to `render_template` can be prefixed with `./` to indicate that it should not search in the root template directory, but in a subdirectory of it with the name of the blueprint.

### DIFF
--- a/flask/templating.py
+++ b/flask/templating.py
@@ -11,8 +11,10 @@
 from jinja2 import BaseLoader, Environment as BaseEnvironment, \
      TemplateNotFound
 
-from .globals import _request_ctx_stack, _app_ctx_stack
+from .globals import _request_ctx_stack, _app_ctx_stack, request
 from .signals import template_rendered, before_render_template
+from ._compat import string_types
+from .ctx import has_request_context
 
 
 def _default_template_ctx_processor():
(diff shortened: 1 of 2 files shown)


flask/templating.py
```diff
@@ -113,12 +114,26 @@ def render_template(template_name_or_list, **context):
...
+    If the template path starts with ``./`` and the request is handled by a
+    `Blueprint` it will replace the ``.`` with the name of the Blueprint.
+
     :param template_name_or_list: the name of the template to be
                                   rendered, or an iterable with template names
                                   the first one existing will be rendered
     :param context: the variables that should be available in the
                     context of the template.
     """
+    blueprint = request.blueprint
```
ThiefMaster: This breaks `render_template` outside a request context

JelteF: So I should check if `request` is `None` first?

JelteF: I changed it a bit together with changing `six` for `._compat`.

flask/templating.py
```diff
@@ -113,12 +114,26 @@ def render_template(template_name_or_list, **context):
...
+    `Blueprint` it will replace the ``.`` with the name of the Blueprint.
+
     :param template_name_or_list: the name of the template to be
                                   rendered, or an iterable with template names
                                   the first one existing will be rendered
     :param context: the variables that should be available in the
                     context of the template.
     """
+
+    if request is not None and request.blueprint is not None:
```
davidism: `if has_request_context() and ...`

davidism:
This won't work for `extends` or `import`, which will lead to even more confusion.  And it doesn't solve the problem that some developers incorrectly expect the blueprints to take priority.

JelteF:
It's true that this won't work for `extends` and `import`, but that is simply because jinja has no notion of requests itself. If you can think of a solution for that I will gladly implement it.

As for the problem of incorrectly expecting blueprints to take priority. This is probably mostly caused by the fact that there is a `url_prefix`, but no `template_prefix` and thus it is assumed that `template_folder` fulfills that function. So maybe `template_prefix` should be added instead.

nopdotcom:
What would the documentation change look like?

(I notice the [development version of Flask](http://flask.pocoo.org/docs/dev/blueprints/#templates) has **much** clearer documentation than [0.10.1](http://flask.pocoo.org/docs/0.10/blueprints/#templates).)

davidism:
I am not saying that we will include this behavior, but *if* we were to, this change should probably happen in `loader.get_source`: https://github.com/pallets/flask/blob/master/flask/templating.py#L54.  Open a new PR for that please.  However, I'm inclined to leave this as it is.

davidism:
https://github.com/pallets/flask/issues/1361#issuecomment-220850342

PR #1819, opened 2016-05-25
fix bug in example/flaskr/flaskr.py

a112121788:


davidism:
This is not a bug.  Flask master has the cli built in, Flask-CLI is a backport for earlier versions.

a112121788:
oh , I see

PR #1826, opened 2016-05-29
Match the hello.py with the one on homepage

koallen:
The `hello.py` script in quickstart is different from the one on the homepage. Therefore, it cannot be run with `python hello.py`, which conflicts with the next paragraph in quickstart.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -19,6 +19,9 @@ A minimal Flask application looks something like this::
     @app.route('/')
     def hello_world():
         return 'Hello, World!'
+    
+    if __name__ = "__main__":
+        app.run()
 
 So what did that code do?
 


untitaker:
Hello @koallen, thanks for finding this!

The text (and the frontpage) rather should be adapted. Launching apps should be rather done through the `flask` command nowadays. See also #1102

koallen:
Ok thanks. I'm new to flask, so I'm not aware of that issue. That looks like a better solution to me!

koallen:
So what should I do with this pull request?

untitaker:
I just pushed https://github.com/pallets/flask/commit/f91aea2aa0c83964cebdd71350318932baffe700 -- if you find any other instances where we talk about directly opening `hello.py`, feel free to open a PR!

PR #1840, opened 2016-05-31
a few more python3-compatible print

fchapoton:
some simple changes towards py3

I hope that the future import is allowed here. It is already used elsewhere in the code of flask.


untitaker:
Thanks!

PR #1843, opened 2016-06-01
Mention the template name conflict issue in blueprint templates docs

RaHus:
re #266

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -182,17 +182,18 @@ resource folder.
 The template folder is added to the search path of templates but with a lower 
 priority than the actual application's template folder. That way you can 
 easily override templates that a blueprint provides in the actual application. 
-This also means that you'll have to make sure that no blueprint template shares 
-the same relative path with any other blueprint or actual application template 
-else a name conflict will occur due to the way flask collects references to 
-template files.
+This also means that if you don't want a blueprint template to be accidentally
+overridden, make sure that no other blueprint or actual application template
+has the same relative path. When multiple blueprints provide the same relative
+template path the first blueprint registered takes precedence over the others.
+
 
 So if you have a blueprint in the folder ``yourapplication/admin`` and you
 want to render the template ``'admin/index.html'`` and you have provided
 ``templates`` as a `template_folder` you will have to create a file like
 this: :file:`yourapplication/admin/templates/admin/index.html`. The reason
-for the extra ``admin`` folder is to avoid the name conflict issue mentioned
-above with a template named ``index.html`` in the actual application template 
+for the extra ``admin`` folder is to avoid getting our template overridden
+by a template named ``index.html`` in the actual application template
 folder.  
 
 To further reiterate this: if you have a blueprint named ``admin`` and you
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -176,16 +176,24 @@ the `template_folder` parameter to the :class:`Blueprint` constructor::
 
     admin = Blueprint('admin', __name__, template_folder='templates')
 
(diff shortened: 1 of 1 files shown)


ThiefMaster:
The commit message is overly long (first line shouldn't be more than 51 chars) and the reference to the issue doesn't seem to be correct (considering that github doesn't link it)

docs/blueprints.rst
```diff
@@ -176,16 +176,24 @@ the `template_folder` parameter to the :class:`Blueprint` constructor::
 
     admin = Blueprint('admin', __name__, template_folder='templates')
 
-As for static files, the path can be absolute or relative to the blueprint
-resource folder.  The template folder is added to the searchpath of
+For static files, the path can be absolute or relative to the blueprint
+resource folder.  
+
+The template folder is added to the searchpath of
```
ThiefMaster: "search path"

RaHus:
@ThiefMaster Noted, i'll fix it on a new PR. Does the content seem ok?

ThiefMaster:
No need for a new PR. Simply amend the commit and force-push.

docs/blueprints.rst
```diff
@@ -176,16 +176,24 @@ the `template_folder` parameter to the :class:`Blueprint` constructor::
...
-provides in the actual application.
+For static files, the path can be absolute or relative to the blueprint
+resource folder.  
+
+The template folder is added to the search path of templates but with a lower 
+priority than the actual application's template folder. That way you can 
+easily override templates that a blueprint provides in the actual application. 
+This also means that you'll have to make sure that no blueprint template shares 
+the same relative path with any other blueprint or actual application template 
+else a name conflict will occur due to the way flask collects references to
```
jeffwidman: The way this is currently worded might imply:
1. the "name conflict" will throw an exception
2. You must always make sure the names are unique, when in fact you might (rarely) wish to leverage this behavior. 

Maybe instead say something like:
> This also means that if you don't want a blueprint template to be accidentally overridden, make sure that no other blueprint or actual application template has the same relative path.

docs/blueprints.rst
```diff
@@ -176,16 +176,24 @@ the `template_folder` parameter to the :class:`Blueprint` constructor::
...
-As for static files, the path can be absolute or relative to the blueprint
-resource folder.  The template folder is added to the searchpath of
-templates but with a lower priority than the actual application's template
-folder.  That way you can easily override templates that a blueprint
-provides in the actual application.
+For static files, the path can be absolute or relative to the blueprint
+resource folder.  
+
+The template folder is added to the search path of templates but with a lower 
+priority than the actual application's template folder. That way you can
```
jeffwidman: What about adding an additional sentence here explaining how the priority order is resolved when two blueprint templates conflict?

jeffwidman:
Big fan of this PR. There's plenty comments on SO and the issue tracker from people surprised by this behavior, so it's great to further clarify it.

docs/blueprints.rst
RaHus: In  [flask.templating.DispatchingJinjaLoader#_iter_loaders](https://github.com/pallets/flask/blob/master/flask/templating.py#L87) looks like the first loader checked is the main app loader and then it walks the blueprint loaders by the order they were registered

Also [this answer](http://stackoverflow.com/a/8737529/1631135) suggests the same.

> First it checks if the app's global loader, which is app.jinja_loader can locate the file. Failing that, it checks all application blueprints (in order of registration, AFAIK) for blueprint.jinja_loader in an attempt to locate the file. 

But i am still wondering about how i should phrase this

docs/blueprints.rst
RaHus: That is much better, thank you for your input!

PR #1855, opened 2016-06-02
Adds simplejson as a testing target.

JamesTFarrington:
Adds one test specifically for a simplejson bug.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -19,6 +19,10 @@ from werkzeug.exceptions import BadRequest
 from werkzeug.http import parse_cache_control_header, parse_options_header
 from werkzeug.http import http_date
 from flask._compat import StringIO, text_type
+try:
+    import simplejson
+except ImportError:
+    import json as simplejson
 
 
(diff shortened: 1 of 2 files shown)


untitaker:
This does not test simplejson within travis! It only adds a tox env.

untitaker:
Also adding this as an alternative value to "devel", "release" is wrong.

jeffwidman:
My mistake, I'm going to hold off on merging any more code changes during this sprint then, as I'm clearly not doing a good job on reviewing right now.

davidism:
We need to test both with and without `simplejson`, since both are supported right now.  If we added those two options on top of the others, we'd now be running 30 envs with `tox`.  That seems a bit excessive.

davidism:
Travis has an example in the docs with 56 envs, I don't feel too bad about this now.

untitaker:
@jeffwidman My suggestion would be to have a policy to require two maintainers to sign off; https://github.com/pallets/meta/issues/18

Issue #1852, opened 2016-06-02
Remove version info from Werkzeug in the docs


dawran6:
The following version information is coming from the Werkzeug docs: "New in version 0.6: The location can now be a unicode string that is encoded using the iri_to_uri() function." See http://flask.pocoo.org/docs/0.11/api/#flask.redirect.

There might be more of this kind of issue in the documentation. For this particular one, you can trace the code to docs/api.rst,: 

    .. autofunction:: redirect

davidism:
@untitaker really not sure what to do here.  Maybe there's a way to indicate the package as well as the version for externals we're re-documenting?  Or disable the versions in those cases?  This came up while looking at #1704.

untitaker:
We should set a manual description for this function, and refer to Werkzeug docs (see intersphinx for that).

Another reason for a necessary rephrasing is that "return a WSGI application" is unnecessarily low-level for Flask's docs, and only appropriate for Werkzeug's.

davidism:
Similar to how I wrote a plugin to collapse old version directives, maybe we can write a plugin that adds "Werkzeug version ...". That would be a lot more maintainable.

tristanfisher:
I apologize if this is more general, but what level of technical detail is desired when writing Flask docs?  I wouldn't consider mentioning WSGI to be low-level, but can understand if that's meant to be abstracted.

davidism:
I don't have a particular problem with how the docs read now. If we need to simplify something that can be done at a later time.

Issue #1704, opened 2016-01-26
Remove old version added and changed tags


davidism:
Some of these lists are getting a bit long and irrelevant.  We should remove everything that wasn't added or changed in 1.0 for the 1.0 release.  This will reset the baseline that things will be added or changed from in the future.

davidism:
Should at least leave 0.10 and 0.11 tags.  Should ensure that everything mentioned in the changelog has a tag in the docs.

dawran6:
I'm working on it in PyCon

untitaker:
I'm not sure what to do with those. Ideally they would be only hidden. We should look at what other projects do.

dawran6:
What do you guys think about commenting out those version tags instead of hiding those?

ThiefMaster:
IMO that just adds noise. Committing commented-out stuff is rarely a good idea

untitaker:
@ThiefMaster @dawran6 I thought about this too. I think it's a reasonable approach, but I'd rather consider all options first. So here they are:

* removing
* commenting out
* Hiding old version markers automatically
    * via a Sphinx extension
    * via CSS

jeffwidman:
My vote is to remove them because I don't think they're relevant to anyone reading the current version of the docs. Flask 0.9 was released nearly four years ago, so very unlikely to be used by anyone for new projects. 

Anyone who needs them for legacy projects can still dig them up by pulling up the old versions of the docs. There's been a number of changes in the docs since 0.9, so legacy projects will still need to pull up those old versions anyway.

PR #1856, opened 2016-06-02
Remove old version added and changed tags #1704

dawran6:
Removed the versionadded and versionchanged tags except for the one linked from external docs. Created issue #1852 to remove those tags from external docs.


jeffwidman:
LGTM, squash and I'll merge

untitaker:
Hang on, do we have any sort of consensus on what to do with those markers? Merging seems a bit too eager

jeffwidman:
No problem, I'll hold off.

Over in #1704 both myself and @davidism were on board with doing this and no one objected, so I assumed we were good to go.

ThiefMaster:
+1 for keeping them but making old ones look less important (or collapsing them). Removing them seems more like a good idea for a post-1.0 version once 1.0 is widespread and previous versions are considered obsolete and a bad idea to use

PR #1860, opened 2016-06-02
Coverage for test_static_path_deprecated and test_static_url_path

jphilipsen05:
Made tests in tests_basic.py

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1116,7 +1116,7 @@ def test_static_files():
     rv.close()
 
 
-def test_static_path_depricated():
+def test_static_path_deprecated():
     with pytest.deprecated_call():
         app = flask.Flask(__name__, static_path='/foo')
     app.testing = True
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1116,6 +1116,25 @@ def test_static_files():
     rv.close()
 
 
+def test_static_path_depricated():
+    with pytest.deprecated_call():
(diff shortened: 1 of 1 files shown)


tests/test_basic.py
```diff
@@ -1116,6 +1116,25 @@ def test_static_files():
     rv.close()
 
 
+def test_static_path_depricated():
```
jeffwidman: Typo should be `deprecated` (without an "i")

jphilipsen05: That's what I thought and how I typed it until I used it from this doc:
http://pytest.org/latest/recwarn.html#ensuring-a-function-triggers-a-deprecation-warning

and the python builtin spells it with an e. So I'm fine either way. Do you need me to push it again or is this something that you were able to fix/change if you need?

ThiefMaster: Simply amend the commit and force-push

jphilipsen05: Thanks for fixing that. Were you able to fix it or do I need to fix and push again?

ThiefMaster:
Check your git config to ensure your email is correct - it's either wrong or just not linked to your GitHub account. The latter is not a big issue but it's nicer if your commits are attributed to your GitHub user!

jphilipsen05:
Typo is fixed.

ThiefMaster:
It'd be better to amend the commit - no need to have separate commits for this

PR #1861, opened 2016-06-02
Document Runtime Error when running outside of application context

### LINKED ISSUE #1521: Better docs for app context and it's solutions
I've been using Flask for about a month now and i can say that it's pretty awesome.

The one thing that annoys me is the "application context" and it's lack of proper documentation.

I understand that app context makes you able to instantiate multiple apps in a single python process. But i've been thinking for quite a while about what is a proper use case for this? can the documentation provide an example about why and how to do this?

In addition, if faced with "working outside application context" exception, how do you overcome this? for example in a celery task. Do we use something like, [Flask Celery Helper](https://github.com/Robpol86/Flask-Celery-Helper)? Because even using the Celery object alone would still result in such exception being raised inside whatever we do inside the celery task itself. That has been kind of a debugging nightmare.

Many Thanks

ryanbackman:
Adds a little documentation of the Runtime Error that you get when working outside an applications context.

fix #1521

### DIFF
--- a/docs/appcontext.rst
+++ b/docs/appcontext.rst
@@ -74,6 +74,11 @@ The application context is also used by the :func:`~flask.url_for`
 function in case a ``SERVER_NAME`` was configured.  This allows you to
 generate URLs even in the absence of a request.
 
+If a request context has not been pushed and an application context has 
+not been explicitly set, a ``RuntimeError`` will be raised.
+::
+    RuntimeError: Working outside of application context.
+
 Locality of the Context
 -----------------------
 


docs/appcontext.rst
```diff
@@ -74,6 +74,11 @@ The application context is also used by the :func:`~flask.url_for`
 function in case a ``SERVER_NAME`` was configured.  This allows you to
 generate URLs even in the absence of a request.
 
+If a request context has not been pushed and an application context has
```
ThiefMaster: "If no request or application context has been pushed, ..."?

Issue #221, opened 2011-04-19
Subclassing of Flask should be documented


dag:
* It should be made clear that for some problems (such as a custom Jinja environment), subclassing is *the* solution and Flask is explicitly designed for it.
* Some examples and best-practices would be helpful, maybe as a 'pattern'. How to extend rather than replace the defaults, how to deal with immutable containers etc.
* The reasons for the need to subclass, and the design choice of immutable containers, could be explained and defended. (Needed at construction time?)

On IRC, questions for which subclassing is the answer, are not uncommon, and at the same time few realize that you can and *should* do this for some things, and they're often confused as to *how* it's properly done.

If I find the energy I might try to draft some of the above.

antsar:
I believe this can be closed, see PR #1868

jeffwidman:
Indeed. Thanks @antsar for your hard work at the Pycon Sprint.

PR #1862, opened 2016-06-02
Add subclassing pattern/example

antsar:
Wrote up a basic example of subclassing Flask to set up an altered Jinja2 environment. This should fix issue #221

### DIFF
--- a/docs/patterns/subclassing.rst
+++ b/docs/patterns/subclassing.rst
@@ -1,0 +1,22 @@
+Subclassing Flask
+=================
+
+The :class:`~flask.Flask` class is designed for subclassing.
+
+One reason to subclass would be customizing the Jinja2 :class:`~jinja2.Environment`. For example, to add a new global template variable::
+
+    from flask import Flask
+    from datetime import datetime
+
+    class MyFlask(Flask):
+        """ Flask with more global template vars """
+
+        def create_jinja_environment(self):
+            """ Initialize my custom Jinja environment. """
+            jinja_env = super(MyFlask, self).create_jinja_environment(self)
+            jinja_env.globals.update(
+                current_time = datetime.datetime.now()
+            )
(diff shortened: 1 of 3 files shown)


docs/patterns/subclassing.rst
```diff
@@ -0,0 +1,22 @@
...
+
+The :class:`~flask.Flask` class is designed for subclassing.
+
+One reason to subclass would be customizing the Jinja2 :class:`~jinja2.Environment`. For example, to add a new global template variable::
+
+    from flask import Flask
+    from datetime import datetime
+
+    class MyFlask(Flask):
+        """ Flask with more global template vars """
```
ThiefMaster: There shouldn't be surrounding whitespace

docs/patterns/subclassing.rst
```diff
@@ -0,0 +1,22 @@
...
+
+        def create_jinja_environment(self):
+            """ Initialize my custom Jinja environment. """
+            jinja_env = super(MyFlask, self).create_jinja_environment(self)
+            jinja_env.globals.update(
+                current_time = datetime.datetime.now()
```
ThiefMaster: kwargs don't have space around the `=` and this example doesn't seem particularly useful; especially considering that this probably runs once at setup time and there are already proper APIs in Flask to add new template globals, filters or tests without subclassing.

antsar:
Thanks for the quick feedback. Any suggestions for a better subclassing use-case?

docs/patterns/subclassing.rst
antsar: Thanks for the quick feedback. Any suggestions for a better subclassing use-case?

ThiefMaster: I'm not sure if there is any good reason to override this specific method.

Issue #1847, opened 2016-06-01
Issue using Flask CLI with module layout recommended in Large Applications


sean-lynch:
Example project that shows issue is available here: https://github.com/sean-lynch/flask-cli-issue . I'm not sure if this is a flask cli issue or an issue with documentation. Happy to correct the docs if it's the latter.

I'm attempting to use the large app structure seen in the [Flask Larger Applications documentation](http://flask.pocoo.org/docs/0.11/patterns/packages/)

First, set up a virutal environment and install flask 0.11 into that environment (requirements.txt
has all that for you).

Can successfully launch app using old `app.run()` mechanism found in `run.py` and `old_launch.sh`

```
$ python run.py
```

But can't seem to get to work with flask-cli. You can see both of the methods in `new_launch.sh`.

This throws `ImportError - No module named server`:

```
$ FLASK_APP=server FLASK_DEBUG=1 flask run
```

This loads, but doesn't register any of the views even though the `Loading views...` print output makes it to the console.

```
$ FLASK_APP=server/__init__.py FLASK_DEBUG=1 flask run
```

How should I be loading a module with flask-cli? Or alternatively, how should I be changing the module structure to support it properly?

mitsuhiko:
That's because you did not actually install your app.

A few options:

* export PYTHONPATH=`pwd`
* pip install --editable . (assuming you have a setup.py, did not look if you have)

An argument can be made that this behavior should be improved. I'm up for that.

sean-lynch:
Ah, hmm. Is installing an app in development a common pattern? I haven't run into it yet but admittedly, all my flask apps have been deployed by me rather than being open sourced for distribution. I was making this change based on the recommended guidance in the release notes to abandon `app.run()` in favor of the cli.

While `export PYTHONPATH=pwd` didn't seem to help, reading the [documentation](http://flask.pocoo.org/docs/0.11/cli/) again, I found that `python -m flask run` fixed the short term problem for me. My final command:

    FLASK_APP=server FLASK_DEBUG=1 python -m flask run

This works fine. Only feel like I'm not really using any of the benefits of the new cli feature now 😄

mitsuhiko:
> Is installing an app in development a common pattern?

Yes! Without `pip install --editable .` you will not get any meta information about your package. Functionality we should make sure that `python -m flask` and `flask` behave the same WRT to the Python path.

dorianpula:
Looking into this as part of the PyCon sprints.

taion:
[...]

PR #1863, opened 2016-06-02
Fix Issue 1847 - Adding Current Path when Invoking Flask-Cli Run

dorianpula:
Fix for: #1847 

Adds the current directory to the Python path, when invoking a local development server.  `flask run` was unable to find the app's module in the PYTHONPATH, while `python -m flask run` adds the current working directory to the PYTHONPATH automatically.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -398,6 +398,8 @@ def run_command(info, host, port, reload, debugger, eager_loading,
     The reloader and debugger are by default enabled if the debug flag of
     Flask is enabled and disabled otherwise.
     """
+    import sys
+    sys.path.append('.')
     from werkzeug.serving import run_simple
 
     debug = get_debug_flag()


ThiefMaster:
Strong :-1:; IMO people should add a setup.py

flask/cli.py
```diff
@@ -398,6 +398,8 @@ def run_command(info, host, port, reload, debugger, eager_loading,
     The reloader and debugger are by default enabled if the debug flag of
     Flask is enabled and disabled otherwise.
     """
+    import sys
```
ThiefMaster: This import is not even needed; `sys` is already imported on the top level.

dorianpula:
Sure, but the fact is that the behaviour between `python -m flask run` and `flask run` is inconsistent.  Not an ideal fix, but then again not every flask app should need a setup.py, if it not being distributed via PyPI.

flask/cli.py
dorianpula: Sure, but the fact is that the behaviour between python -m flask run and flask run is inconsistent. Not an ideal fix, but then again not every flask app should need a setup.py, if it not being distributed via PyPI.

dorianpula:
Sure, but the fact is that the behaviour between python -m flask run and flask run is inconsistent. Not an ideal fix, but then again not every flask app should need a setup.py, if it not being distributed via PyPI.

ThiefMaster:
A setup.py makes perfect sense even if you never intend to publish your package on PyPI. It's how your register a package properly within your Python environment.

untitaker:
You can already set `FLASK_APP=file.py` (or `FLASK_APP=package/__init__.py`), and the appropriate path will be added to `sys.path`.

`setup.py` is very useful in a lot of cases, but it is not required.

All of this is poorly documented and a few instances in the quickstart are completely wrong (i.e. nonfunctional) in 0.11 and only fixed in master. If you find any other instances where this behavior could be clarified, feel free to point them out. Thanks!

davidism:
@untitaker there appears to be some path issues with pointing at `__init__.py`.  https://github.com/pallets/flask/issues/1847#issuecomment-223439632  Pointing at `__init__` seems unintuitive, given that you would normally never do that in imports or `-m` calls.

davidism:
I would expect to be able to point to `FLASK_APP=package`, and I'm sure the vast majority of "didn't use setup.py" cases are running from the app directory.  Given that `FLASK_APP=package python -m flask run` works in this case, and is doing the same thing to the path, this seems like an ok workaround to make both behave the same.

Maybe we could be smarter about when we add the current directory to the path, or output a warning to use setup.py instead when we do.

davidism:
Either way, we need a clearer message when this fails, rather than just dumping an `ImportError` traceback.

> Could not locate a Flask app at \[location\].  Install your app in development mode, or see \[the docs\] for more information.

jeffwidman:
Not a fan of forcing folks to use a `setup.py`, I prefer that we let folks get up and running with a simple `flask run` within their app directory.

untitaker:
See https://github.com/pallets/flask/pull/1872, I think this got fixed.

untitaker:
@jeffwidman I don't think that's possible at all. How would the command know which package/file to

jeffwidman:
Yeah, you're right. I am so used to accessing `run` via a simple custom `manage.py run` command that I wrote that I didn't think through how that might work if someone doesn't write their own custom command.

Issue #801, opened 2013-07-23
Given absolute URL, testclient discards query params


ryanolson:
The following fails at the assert inside the index subroutine.

``` python
import unittest
from flask import Flask, url_for, request, jsonify

class TestFlask(unittest.TestCase):

    def test_get_with_arguments(self):
        app = Flask(__name__)
        app.debug = True

        @app.route('/echo')
        def index():
            assert 'key' in request.args
            return jsonify( dict(key=request.args['key']) )

        url = None
        with app.test_request_context():
            url = url_for('index', key='test', _external=True)

        assert url == 'http://localhost/echo?key=test'

        rv = app.test_client().get(url)
        assert 'key' in rv.data
```

manuelgodoy:
When I build my dict and pass it on as `data` the args are not received: 

```javascript
import unittest
from flask import Flask, url_for, request, jsonify

class TestFlask(unittest.TestCase):

    def test_get_with_arguments(self):
        app = Flask(__name__)
        app.debug = True

        @app.route('/echo')
        def index():
            assert 'key' in request.args
            return jsonify( dict(key=request.args['key']) )
 
        rv = app.test_client().get('/echo', data = dict(key = 'test') )
        assert 'key' in rv.data
```

Frozenball:
This behaviour seems a bit unexpected. What's the reasoning behind this behaviour?

adambyrtek:
I noticed this issue in other circumstances, without using the `_external` flag at all. Most of our view tests don't use request context explicitly, for example:

```
def test_foo_bar(client):
    response = client.get(url_for('endpoint'))
    assert response.status_code == 200
```

In order to make `url_for` work without the request context we had to explicitly set `app.config['SERVER_NAME'] = 'localhost'` in test setup, but besides that this works perfectly and is much less verbose.

The problem is that in this case `url_for` generates absolute URLs (e.g. `http://localhost/endpoint`) which suffer from the issue in question. In other words the obvious code doesn't work:

```
def test_foo_bar(client):
    response = client.get(url_for('endpoint', param='foo'))
    assert response.status_code == 200
```

And has to be replaced with much less obvious:

```
def test_foo_bar(client):
    response = client.get(url_for('endpoint'), query_string={'param': 'foo'})
    assert response.status_code == 200
```

Worst of all there is no warning whatsoever so you don't even know why a test is failing for a URL that seems perfectly correct.

sprin:
[...]

PR #1866, opened 2016-06-02
url_for now supports no request context and no SERVER_NAME case

jasonslay:
#801

The simplest way to accomplish this was to bring in a modified version of the wrapper found here:
https://github.com/miguelgrinberg/flack/blob/master/flack/utils.py#L11

This wrapper also needed to cover the case where an application context existed with `SERVER_NAME` defined. 

I added two tests (one to cover when `SERVER_NAME` doesn't exist, and one when it does). The removal of `test_url_generation_requires_server_name` is because the error case it covered no longer exists.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -195,7 +195,6 @@ def url_for(*args, **kwargs):
     wrapper based on https://github.com/miguelgrinberg/flack/blob/master/flack/utils.py
     that also covers the case where application context exists with SERVER_NAME defined.
     """
-
     reqctx = _request_ctx_stack.top
     appctx = _app_ctx_stack.top
     if reqctx is None and (appctx is None or appctx.url_adapter is None):
@@ -317,7 +316,6 @@ def _url_for(endpoint, **values):
                                'adapter for request independent URL generation. '
                                'You might be able to fix this by setting '
                                'the SERVER_NAME config variable.')
-
         external = values.pop('_external', True)
 
     anchor = values.pop('_anchor', None)


dorianpula:
Looks good to me.

flask/helpers.py
```diff
@@ -190,7 +190,23 @@ def index():
     return current_app.make_response(args)
 
 
-def url_for(endpoint, **values):
+def url_for(*args, **kwargs):
+    """
+    wrapper based on https://github.com/miguelgrinberg/flack/blob/master/flack/utils.py
```
ThiefMaster: This pretty much screws up the `url_for` documentation

flask/helpers.py
```diff
@@ -190,7 +190,23 @@ def index():
...
+    that also covers the case where application context exists with SERVER_NAME defined.
+    """
+    reqctx = _request_ctx_stack.top
+    appctx = _app_ctx_stack.top
+    if reqctx is None and (appctx is None or appctx.url_adapter is None):
+        if kwargs.get('_external', False):
+            raise RuntimeError('Cannot generate external URLs without a '
+                               'request context.')
+        with current_app.test_request_context():
```
ThiefMaster: This feels extremely dirty. It's not the purpose of `url_for` to push a context.

davidism:
First, sorry it took so long to finish this. 😩

I'm reviewing the linked flack project, and I can't figure out why it's necessary in that project. All the calls to `url_for` happen in views (during a request context), or in `@async` decorated tasks, which wraps them in a request context. So the original function works fine as is.

The linked issue is about something slightly different, but the comment at the end outlines a less radical change that would still fix it. Closing this in favor of that.

PR #1868, opened 2016-06-02
Rewrite subclassing example with a better use-case.

antsar:
As @ThiefMaster pointed out in PR #1862, template globals can be added without subclassing. Here is a different use-case, inspired by [recent IRC conversations](https://botbot.me/freenode/pocoo/2016-05-30/?msg=66949866&page=2).


jeffwidman:
LGTM

untitaker:
Excellent, thanks!

PR #1869, opened 2016-06-02
Tests with and without simplejson for every existing testenv

JamesTFarrington:
changes configuration for tox and travis. Assist from @davidism


jeffwidman:
LGTM

PR #1872, opened 2016-06-02
Clitest update

jphilipsen05:
There was a bad if/elif statement and the elif os.path ... would never get hit. So I swapped the order of those two statements and added testing for the prepare_exec_for_file function.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -55,10 +55,10 @@ def prepare_exec_for_file(filename):
     module = []
 
     # Chop off file extensions or package markers
-    if filename.endswith('.py'):
-        filename = filename[:-3]
-    elif os.path.split(filename)[1] == '__init__.py':
+    if os.path.split(filename)[1] == '__init__.py':
         filename = os.path.dirname(filename)
+    elif filename.endswith('.py'):
+        filename = filename[:-3]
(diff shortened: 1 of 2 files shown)


ThiefMaster:
second commit should be amended into the first one

also, the first line of a commit should be no longer than 51 chars (right not even github truncates it) - see http://chris.beams.io/posts/git-commit/ for some useful guidelines

untitaker:
@ThiefMaster We can squash ourselves so commit styling is no longer that relevant.

@jphilipsen05 Can you verify (on your machine) that the tests fail without the bugfix?

jphilipsen05:
Yes, If I swap the if/elif statements back to the original I get this error in the tests. Showing that the first if statement for an __init__.py is hitting the if statement that handles python files and does not reach the expected __init__.py elif statement. 

When I have my bug fix in place there are no errors. 

Let me know if you have any other questions.

test_apps = None

    def test_prepare_exec_for_file(test_apps):
        assert prepare_exec_for_file('test.py') == 'test'
>       assert prepare_exec_for_file('/usr/share/__init__.py') == 'share'
E       assert '__init__' == 'share'
E         - __init__
E         + share

PR #1873, opened 2016-06-02
Using `setup.cfg` with fabric

### LINKED ISSUE #129: Thoughts about the fabric docs
Maybe use of one of these should be suggested:

    setup.py egg_info ...
    --tag-svn-revision (-r)  Add subversion revision ID to version number
    --tag-date (-d)          Add date stamp (e.g. 20050528) to version number
    --tag-build (-b)         Specify explicit tag to add to version number

A problem with this is that ``setup.py --fullname`` does not include the tag. Not really an issue with ``-b``, a useful value might be ``git log -1 --format=%h`` for example.

Another thought: why go through all that trouble to extract the tarball?

    easy_install myapp.tar.gz

Also: why not call ``pack()`` in deploy?

SkiftCreative:
Explaining how to use `setup.cfg` to configure test builds when trying to deploy with fabric, and using `sdist`. Fixes #129

### DIFF
--- a/docs/patterns/fabric.rst
+++ b/docs/patterns/fabric.rst
@@ -156,6 +156,25 @@ location where it's expected (eg: :file:`/var/www/yourapplication`).
 Either way, in our case here we only expect one or two servers and we can
 upload them ahead of time by hand.
 
+Configuring egg_info
+--------------------
+If you need to configure your fabric build with tags, you can create a `setup.cfg`
+file in the root of your app. An example would be:
+
+    [egg_info]
+    tag_svn_revision = 1
+    tag_build = .dev
+    tag_date = 1
+
+    [aliases]
+    release = egg_info -RDb ''
+
+And now when running `python setup.py sdist ...`  in your fabric file, it will
+pick up on these settings and tag appropriately. And when making a release build
+it will ignore these build tags as expected.
+
+
+
 First Deployment
 ----------------
 


ThiefMaster:
Can you please amend the commit message? Ideally the first line should be no longer than 51 chars; the current one is so long that even GitHub truncates it. The issue reference could go in a later line for example - when looking at the short git log the issue number is not really relevant.

http://chris.beams.io/posts/git-commit/ has some good guidelines on writing nice commit messages.

docs/patterns/fabric.rst
```diff
@@ -156,6 +156,25 @@ location where it's expected (eg: :file:`/var/www/yourapplication`).
 Either way, in our case here we only expect one or two servers and we can
 upload them ahead of time by hand.
 
+Configuring egg_info
+--------------------
+If you need to configure your fabric build with tags, you can create a `setup.cfg`
+file in the root of your app. An example would be:
+
+    [egg_info]
+    tag_svn_revision = 1
```
ThiefMaster: is that just poor naming or only relevant when using SVN

davidism: Yeah, `tag_svn_revision` only really works with svn.  'dev<date>` is generally useful.
```
tag_build = .dev
tag_date = 1
```

davidism:
This would be better under the setuptools page, it's useful for packaging in general.

cyberpulseraj:
Zt

Sent from my iPhone

OfOn 3 Jun 2016, at 16:55, David Lord <notifications@github.com<mailto:notifications@github.com>> wrote


This would be better under the setuptools page, it's useful for packaging in general.

-

davidism:
33212309a22d3aeed725e3138d48593dc0a5b47f, then moved in 14a5a9e5547895f9e77dfd2dccd7fdb63fe2b8da

PR #1874, opened 2016-06-02
Comment out old version added and changed tags. Fix #1704

### LINKED ISSUE #1704: Remove old version added and changed tags

dawran6:
Comment out the "versionadded" and "versionchanged" tags by adding # in
the front. The line (or block) would be considered as comments.

This is another attempt at #1856 based on #1704 (comment)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -205,7 +205,7 @@ To access the current session you can use the :class:`session` object:
 Session Interface
 -----------------
 
-.. versionadded:: 0.8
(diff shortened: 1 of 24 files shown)


dawran6:
Copied @ThiefMaster 's comment here:

Committing commented-out code is usually noise...

PR #1875, opened 2016-06-02
Add a link to Extension Development

hyunchel:
Provides a link to [Flask Extension Development][0] page from [Flask Extension page][1].
Flask Extension page talks about `finding`, `using` extensions, it might as well talk about `building` one too. Found #532 as the most relevant issue that is open.

[0]: http://flask.pocoo.org/docs/0.11/extensiondev
[1]: http://flask.pocoo.org/docs/0.11/extensions

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -25,6 +25,13 @@ importable from ``flask_foo``::
 
     import flask_foo
 
+Building Extensions
+-------------------
+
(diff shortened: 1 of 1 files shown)


untitaker:
#532 is not totally relevant, which is more about explaining the usage of some popular extensions in the official documentation.

But anyway this change seems golden. Thanks!

PR #1878, opened 2016-06-03
Use recwarn everywhere

untitaker:
...instead of custom fixture. Also assert that no warnings are left over
after the test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@
     :license: BSD, see LICENSE for more details.
 """
 import flask
+import gc
 import os
 import sys
 import pkgutil
@@ -129,4 +130,5 @@ def purge_module(request):
 @pytest.yield_fixture(autouse=True)
 def catch_deprecation_warnings(recwarn):
     yield
+    gc.collect()
     assert not recwarn.list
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -126,8 +126,9 @@ def inner(name):
     return inner
 
 
-@pytest.fixture
-def catch_deprecation_warnings():
-    import warnings
-    warnings.simplefilter('default', category=DeprecationWarning)
-    return lambda: warnings.catch_warnings(record=True)
+@pytest.fixture(autouse=True)
+def catch_deprecation_warnings(recwarn, request):
```
ThiefMaster: I think `yield_fixture` would be cleaner and more readable:

```python
@pytest.yield_fixture(autouse=True)
def catch_deprecation_warnings(recwarn):
    yield
    assert not recwarn.list
```

untitaker:
@ThiefMaster @davidism Please review.

tests/conftest.py
untitaker: Interesting, wasn't aware of that.

untitaker:
Interesting, we're getting a ResourceWarning somewhere.

PR #1882, opened 2016-06-03
Enhance tests.test_cli.test_find_best_app

hyunchel:
This commit adds a test case for `test_find_best_app` where
Module object does not contain Flask application.
Also cleans the function little bit to provides more meaningful comment.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -29,24 +29,27 @@ def test_cli_name(test_apps):
 
 
 def test_find_best_app(test_apps):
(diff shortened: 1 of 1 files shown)


jeffwidman:
LGTM, I don't think there's any name conflicts with using `Module` but not 100% sure

hyunchel:
I can call it `SampleModule` or `FlaskModule` just to be safe.
Should I force push with the change above?

jeffwidman:
No, let's wait for another maintainer to take a look.

PR #1886, opened 2016-06-03
Enhance code coverage for Blueprint.endpoint

dawran6:
Add basic test for the endpoint decorator for the Blueprint object.

Mentioned on the API [docs](http://flask.pocoo.org/docs/0.11/api/#flask.Blueprint.endpoint): 

> ...If the endpoint is prefixed with a . it will be registered to the current blueprint, otherwise it’s an application independent endpoint.

This commit only added test for the _without-prefixed-dot_ endpoint


dawran6:
I need help to understand how ___If the endpoint is prefixed with a . it will be registered to the current blueprint___ works

Issue #434, opened 2012-03-12
sessions documentation (client side vs server side)


wilsaj:
more explicit docs on client side sessions vs server side sessions (and how to implement server-side sessions)

recent mailing list discussion: http://flask.pocoo.org/mailinglist/archive/2012/3/1/is-it-wise-for-flask-to-store-session-variables-on-the-client/
unhappy feedback: http://feedback.flask.pocoo.org/message/223

jeffwidman:
I think the best fix is to add a sentence to the [quickstart's section on sessions](http://flask.pocoo.org/docs/dev/quickstart/#sessions) saying that if you want to handle sessions server-side rather than client-side, there are several Flask extensions that support this. I wouldn't link directly to a single extension, as extensions come and go. Mostly we just want to make folks reading this aware of the existence, and then they can google to learn more...

dawran6:
Hi guys, could you review my commit 486ed09? Thanks!

jeffwidman:
@dawran6 Not a big deal, but just FYI  typically you just point toward a PR, as the individual commit may change based on feedback, resulting in a different hash for the newest version. Plus, when you force-push to the PR branch that particular commit link is broken. You can link to a PR with `#34` or whatever number it is and GH will auto-link it.

frederickOtus:
Is there a reason that this issue is still open?

PR #1888, opened 2016-06-04
sessions documentation (client side vs server side) #434

dawran6:
Mention the existence of Flask extentions that handle server-side sessions

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/1888)

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -817,6 +817,9 @@ values do not persist across requests, cookies are indeed enabled, and you are
 not getting a clear error message, check the size of the cookie in your page
 responses compared to the size supported by web browsers.
 
+Besides the default client-side based sessions, if you want to handle
+sessions on the server-side instead, there are several
+Flask extensions that support this.
 
 Message Flashing
 ----------------


dawran6:
PR for #434

untitaker:
We should go into a lot more detail and actually list extensions that do this.

untitaker:
Also the text above is very explicit about where sessions get stored (cookies). It can't harm to restate for clarity, but as of now the reading flow is a bit awkward IMO

jeffwidman:
@untitaker I'm hesitant to go  into much detail here because this is the quickstart, so we want to keep it quick, especially when Flask's default client-side sessions suffice for most use cases. Primarily I thought we could add a sentence to make n00b web-devs aware of the existence of server-side sessions and they can google if they want more.

untitaker:
I overlooked that this was the quickstart. At some point sessions might have to get its own page, if we want to list extensions. 

I still insist about the reading flow being bad. The paragraph above talks about sessions being stored entirely client side and this one opens with "Sessions are stored clientside" as if that wasn't already established.

untitaker:
Hey @dawran6, are you still interested in pushing this PR forward?

dawran6:
Hey @untitaker , @jeffwidman , sorry for taking so long. I only changed a bit on the transition words. What do you guys think?

untitaker:
LGTM, definetly an improvement. Thanks!

PR #1889, opened 2016-06-04
Test side effect

hyunchel:
Function `prepare_exec_for_file` has a side effect where a path is added to
`sys.path` list.
This commit enhances an exisiting test case for `prepare_exec_for_file`
by testing the side effect of the function and adding necessary
comments.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -12,6 +12,8 @@
 # Copyright (C) 2015 CERN.
 #
 from __future__ import absolute_import, print_function
+import os
+import sys
 
 import click
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -50,10 +52,23 @@ class mod:
 
 
 def test_prepare_exec_for_file(test_apps):
-    assert prepare_exec_for_file('test.py') == 'test'
+    """Expects correct path to be set and correct module name to be returned.
+
+    :func:`prepare_exec_for_file` has a side effect, where
+    parent directory of given file is added to `sys.path`.
+    """
+    realpath = os.path.realpath('/usr/share/test.py')
```
ThiefMaster: I'd use `/tmp` as an example path. IMHO that makes more sense than a path that'd probably never contain a python file

tests/test_cli.py
```diff
@@ -50,10 +52,23 @@ class mod:
 
 
 def test_prepare_exec_for_file(test_apps):
-    assert prepare_exec_for_file('test.py') == 'test'
+    """Expects correct path to be set and correct module name to be returned.
+
+    :func:`prepare_exec_for_file` has a side effect, where
+    parent directory of given file is added to `sys.path`.
```
ThiefMaster: +the

tests/test_cli.py
```diff
@@ -50,10 +52,23 @@ class mod:
 
 
 def test_prepare_exec_for_file(test_apps):
-    assert prepare_exec_for_file('test.py') == 'test'
+    """Expects correct path to be set and correct module name to be returned.
```
ThiefMaster: *Expect

hyunchel:
@ThiefMaster Thank you for the review! I've force pushed with the corrections.

PR #1891, opened 2016-06-06
Fix spelling

zkanda:
Minor spelling fix.

### DIFF
--- a/docs/python3.rst
+++ b/docs/python3.rst
@@ -13,7 +13,7 @@ You should use the latest versions of all Flask-related packages.
 Flask 0.10 and Werkzeug 0.9 were the first versions to introduce Python 3 support.
 
 Python 3 changed how unicode and bytes are handled,
-which complicated how low level code handles HTTP data.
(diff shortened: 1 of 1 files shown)


untitaker:
No, this is past-tense.

zkanda:
Ahh didn't really understand that. Thanks for clarifying it @untitaker :+1:

davidism:
Technically either is correct here, although I'd probably change it to "complicates" since the change happened in the past but the effect is continuous.  The rest of the paragraph is present tense as well.  Of course, now I'm questioning the choice of *affect* and *effect*.  :-/

untitaker:
Agreed:

* complicate -> complicates
* affect -> effect

davidism:
http://chat.stackoverflow.com/transcript/message/30984060#30984060 Consensus seems to be "affect" in both places.

untitaker:
Yes, a typo of mine. Fixing both now.

untitaker:
Fixed!

PR #1892, opened 2016-06-06
Updating url in errorhandling.rst to a valid link

avborhanian:
Was originally set to github.com/getsentry/sentry, which was a relative link to a page that doesn't exist. I prepended https:// to fix this problem.


untitaker:
Thanks!

PR #1896, opened 2016-06-08
Added make target for test coverage, simplified contributing doc

antsar:
The commands to run a coverage report are somewhat clunky, and hard to remember.
1. `make cov` will now generate the report, display it, and save an HTML version.
2. `make test` already ensures that test dependencies are installed. This seemed clever and helpful for beginners, so I did the same for `make tox-test` and `make cov`. (Also, added `tox` and `pytest-cov` to test-requirements.txt)
3. Updated contribution guide to replace the `py.test`, `tox`, and coverage instructions with the `make` versions. Consistent/similar commands should be easier to remember? Also ensures everyone runs the same tests every time.

Apologies if I globbed too much into one PR, happy to edit if needed.


jeffwidman:
I like most everything here except the docs changes (item 3). Personally, I prefer that we continue to recommend using the normal tool commands rather than hiding them behind `make x`. For example,  `py.test` rather than `make test`.

If we bury everything under `make x` commands, then while it's easier to run the default case, it makes it more difficult to customize things adhoc when developing. These instructions target developers who want to hack on flask, so I don't think it's an unreasonable bar for them to have to learn the basics of these tools. I do like having `make cov` as a shortcut, so I would be in favor of saying "Here's the instructions.... If you want a shortcut for the above that runs everything, you can run `make x`."

> Also, added tox and pytest-cov to test-requirements.txt

It doesn't look we're doing this, but sometimes projects use `test-requirements.txt` in their `tox.ini` file to tell Tox what to install... so you can accidentally end up telling tox to install tox. I just checked, and not an issue for Flask.

antsar:
Good point, probably not helpful *remove* info from the docs. Rewritten (approximately) as suggested.

antsar:
Sorry for the spam. Mis-clicked a delete button.

kennethreitz:
needs a rebase!

kennethreitz:
i merged this manually

Issue #1907, opened 2016-06-11
Jinja templates do not auto reload if Flask app debug enabled solely via app.run(debug=True)


davb5:
When enabling debug mode via `app.run(debug=True)`, code changes result in an auto-reload but Jinja templates are cached until the app is manually is restarted.

At https://github.com/pallets/flask/blob/21d595bee73e65a28201709bd1a8b97c42aff5d2/flask/app.py#L695 Flask checks if the `config["TEMPLATE_AUTO_RELOAD"]` is explicitly set. If not, Jinja's `options["auto_reload"]` is set to `app.debug`.

However, if you instantiate your Flask object _then_ call `app.run(debug=True)` without loading a config, template auto reloading is not enabled (`app.debug` was `False` when the Flask app was instantiated).

At https://github.com/pallets/flask/blob/21d595bee73e65a28201709bd1a8b97c42aff5d2/flask/app.py#L839 we enable reloading code and enable the debugger. My proposed fix would be to also enable template reloading (`jinja_env.auto_reload = self.debug`) here.

This would result is less confusion ("Debug is enabled but my templates don't auto reload") and make the template reloading behaviour consistent with the code reloading behaviour ("If no config is set or `debug` is not explicitly set in config, and the site calls `app.run(debug=True)`, then enable the relevant auto reloading features").

jeffwidman:
Your proposed fix makes sense to me.

davb5:
Some further clarification...

In attempting to put together a simple proof of concept for the issue, I've further narrowed down exactly where it's triggered.

In a very simple PoC, if I ...

1. Instantiate my `Flask` app
2. Add some routes
3. Call `app.run(debug=True)`

... the issue does not manifest. I can alter templates and they are reloaded as expected.

This works because the `jinja_env` isn't actually initialised until *after* `app.run(debug=True)` is called (so it's seeing `app.debug == True`).

However, if I ...

1. Instantiate my `Flask` app
2. Add some routes
3. **Access my `app.jinja_env` object (thereby initialising the Jinja2 environment)**
4. Call `app.run(debug=True)`

... template auto-reloading doesn't work as expected.

This is because accessing `app.jinja_env` creates the Jinja environment, reading `app.debug` before `app.run(debug=True)` sets it to `True`.

I think the issue still stands despite the narrower impact of only affecting users who access the `jinja_env`.

[...]

PR #1910, opened 2016-06-14
Enable template auto-reloading in app.run()

davb5:
When Flask app debugging is enabled (app.debug==True), and Jinja2
template auto-reloading is not explicitly disbaled, template
auto-reloading should be enabled.

If the app is instantiated, the jinja_env object is accessed (thereby
initialising the Jinja2 environment) and the server is then started with
app.run(debug=True), template auto-reloading is _not_ enabled.

This is because reading the jinja_env object causes the environment
initialisation function to set auto_reload to app.debug (which isn't yet
True). Calling app.run(debug=True) should correct this in order to
remain consistent with Flask code reloading (which is enabled within
app.run() if debug == True).

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -839,6 +839,8 @@ class Flask(_PackageBoundObject):
         options.setdefault('use_reloader', self.debug)
         options.setdefault('use_debugger', self.debug)
         options.setdefault('passthrough_errors', True)
+        if debug:
+            self.jinja_env.auto_reload = True
         try:
             run_simple(host, port, self, **options)
         finally:
--- a/tests/test_templating.py
+++ b/tests/test_templating.py
@@ -349,11 +349,9 @@ def test_templates_auto_reload():
 
 def test_templates_auto_reload_debug_run(monkeypatch):
     # debug is None in config, config option is None, app.run(debug=True)
-    rv = {}
-
     # Mocks werkzeug.serving.run_simple method
     def run_simple_mock(*args, **kwargs):
-        rv['passthrough_errors'] = kwargs.get('passthrough_errors')
+        pass
 
     app = flask.Flask(__name__)
     monkeypatch.setattr(werkzeug.serving, 'run_simple', run_simple_mock)
--- a/tests/test_templating.py
+++ b/tests/test_templating.py
@@ -14,6 +14,7 @@ import pytest
 import flask
 import logging
 from jinja2 import TemplateNotFound
+import werkzeug.serving
 
 
 def test_context_processing():
@@ -344,6 +345,22 @@ def test_templates_auto_reload():
     app = flask.Flask(__name__)
     app.config['DEBUG'] = True
     app.config['TEMPLATES_AUTO_RELOAD'] = True
+    assert app.jinja_env.auto_reload is True
+
+def test_templates_auto_reload_debug_run(monkeypatch):
+    # debug is None in config, config option is None, app.run(debug=True)
+    rv = {}
+
+    # Mocks werkzeug.serving.run_simple method
+    def run_simple_mock(*args, **kwargs):
+        rv['passthrough_errors'] = kwargs.get('passthrough_errors')
+
+    app = flask.Flask(__name__)
+    monkeypatch.setattr(werkzeug.serving, 'run_simple', run_simple_mock)
+
+    assert app.config['TEMPLATES_AUTO_RELOAD'] is None
+    assert app.jinja_env.auto_reload is False
+    app.run(debug=True)
     assert app.jinja_env.auto_reload is True
 
 def test_template_loader_debugging(test_apps):


jeffwidman:
This looks fine to me. Can you add a test though?

davb5:
No problem, I'll add a test to this later this evening. Thanks.

flask/app.py
```diff
@@ -839,6 +839,8 @@ def run(self, host=None, port=None, debug=None, **options):
         options.setdefault('use_reloader', self.debug)
         options.setdefault('use_debugger', self.debug)
         options.setdefault('passthrough_errors', True)
+        if debug:
```
justanr: Shouldn't this depend on the reloader being active, not just debug?

davb5:
I don't think so - that's the root of the problem.

The template reloader gets disabled at Flask app instantiation if it's not explicitly enabled and debug mode isn't turned on at that point.

However, in the case if the code reloader, reloading is enabled in `app.run()` it is not explicitly disabled and debug is enabled.

I think the patch harmonises the behaviour between template and code reloading (if no explicit reloading configuration and debug is enabled via `app.run(debug=True)`, reloading is set to match `debug`).

untitaker:
I think you're confusing the template reloader of Jinja (enabled by default) with the server reloader from Werkzeug (disabled by default).

untitaker:
And TEMPLATES_AUTO_RELOAD controls how Jinja's `auto_reload` is set. The default `None` value means that it depends on debug mode (see `create_jinja_environment`, https://github.com/pallets/flask/blob/5c12721730d18e2eddfe083cf1f6398af8915643/flask/app.py#L695)

untitaker:
Sorry, I see that you've figured all of this already out in https://github.com/pallets/flask/issues/1907#issuecomment-225743376. In that case I would say that we need to deduplicate that logic anyway, i.e. remove the logic for TEMPLATE_AUTO_RELOAD out of `create_jinja_environment` into `run`.

davb5:
@untitaker That's the problem. If you don't explicitly enable debug via configuration at Flask app instantiation then access the `jinja_env`, template auto reloading is disabled. If you then subsequently enable debug by starting the server with `app.run(debug=True)`, Werkzeug debugging and reloading is enabled (regardless of the fact it wasn't otherwise explicitly enabled via config) but template auto reloading isn't enabled.

I think the intention is that calling `app.run(debug=True)` should enable debugging and auto reloading. It's confusing that this enables code reloading (when no explicit option is set) but doesn't enable template reloading (when no explicit option is set).

Sorry if I'm not explaining this very well..

untitaker:
Yes I oversaw your post in #1907, it all makes sense.

But with this fix there's another inconsistency: If I explicitly set `app.jinja_env.auto_reload = False` and then run `app.run(debug=True)`, the value gets ignored?

justanr:
@davb5 What i meant specifically was using `options['use_reloader']` instead of just the debug flag itself. There's instanceswhen I've enabled debug but disabled reloading intentionally.

untitaker:
You are confusing Werkzeug server reloading with Jinja env reloading.

On 14 June 2016 23:36:46 CEST, Alec Nikolas Reiter <notifications@github.com> wrote:
>@davb5 What i meant specifically was using `options['use_reloader']`
>instead of just the debug flag itself. There's instanceswhen I've
>enabled debug but disabled reloading intentionally.   
>
>---
>https://github.com/pallets/flask/pull/1910#issuecomment-226023828

-- 

davb5:
Ah, I understand now @untitaker.

Sorry, I missed that comment. You're right in that we're clobbering any configuration set value from `create_jinja_environment`, and this probably isn't the best course of action.

The difficulty here is that between instantiating our Flask app and calling `app.run()` there are a few places that template auto reloading could be enabled or disabled.

My proposed patch works for most scenarios but fails for the one you've highlighted - if we explicitly disable template auto reloading (using any of the above mechanisms) then call `app.run(debug=True)`, we clobber our explicit disablement of template reloading. And that's probably not ideal.

@justanr's comment that we might want to use `options['use_reloader']` rather than `debug` makes some sense, but could this cause further confusion by tying Jinja template autoreloading to Werkzeug reloading in one place (`app.run`) but no-where else?

I think you're right in that moving the logic for TEMPLATE_AUTO_RELOAD into `app.run` might be the solution. We'd have to make sure that nothing else fails and no behaviours change as a result of template auto-reloading not being configured one way or another before `app.run` (i.e. make sure nothing expects some value to be there before `app.run`).

What does Jinja do if  `auto_reload` is not set? Does it have a sensible default behaviour?

If so, we could probably just move that block from `create_jinja_environment` into `run` pretty much wholesale. Would this be a better solution?

davb5:
Actually, that might not work without some more changes.

I don't believe `run()` is called at all in WSGI apps, so that couldn't be the only place that we set `jinja_env.auto_reload`.

davb5:
Actually, https://github.com/pallets/flask/pull/1910#issuecomment-226035937 might not matter as I don't think the Werkzeug auto reloader works in a WSGI environment (could someone please confirm?).

If we don't auto reload code in a WSGI environment, keeping the template auto reloader configuration in `app.run()` would make sense and keep things consistent.

davb5:
I've added a test for the original patch (wether or not we use that solution, it's a starting point for the unit test).

davidism:
Except `app.run` isn't the recommended way to run in dev mode either now.  `FLASK_APP=app flask run` is, and that's a different code path.

nguyenkims:
hey does anyone have an update on this pull request?

illume:
A quick workaround is to enable it manually:
```class YourFlask(Flask):
    def create_jinja_environment(self):
        self.config['TEMPLATES_AUTO_RELOAD'] = True
        return Flask.create_jinja_environment(self)
```

I guess this issue is blocking a lot of people who use flask to develop templates.

markrjr:
Thank goodness I found this, I've been tearing my hair out for a few hours trying to figure out why my templates weren't auto updating. I guess my machine must have updated to 0.11.1 by itself. What release are you guys targeting for this fix?

cihantas:
Any updates on when this will be merged?

deelaka:
Adding ` app.jinja_env.auto_reload = True` along with `app.config['TEMPLATES_AUTO_RELOAD'] = True` seems to fix.

kennethreitz:
needs a rebase!

kennethreitz:
closing due to inactivity. please consider re-opening this pull request :)

davidism:
Sorry for close/reopen spam, was trying to see if "allow edits from maintainers" was enabled for the PR. I'm going to continue this in another PR.

GhastlyParadox:
So, I'm at a loss. I've tried every proposed solution I've been able to find, without luck. My Flask app completely ignores any changes made to my html template files, no matter what I do. 

I've tried every permutation of setting debug and auto_reload I can think of. I currently have it running with a solution proposed in this thread (code below) - no errors, but still it completely ignores any changes made to template html files. 

Am I going mad? What could I be missing? 

```
class YourFlask(Flask):
    def create_jinja_environment(self):
        self.auto_reload = True
        self.config['TEMPLATES_AUTO_RELOAD'] = True
        return Flask.create_jinja_environment(self)

app = YourFlask(__name__)
app.config.update(DEBUG=True, SECRET_KEY=os.getenv("FLASK_APP_KEY"))

....

if __name__ == "__main__":

    connect_to_db(app)
    app.run()
```

Issue #1902, opened 2016-06-09
Import error with some examples


rnelsonchem:
I'm trying to run the "flaskr" and "minitwit" examples, but I'm running into an import error issue. I'm trying to follow the instructions from the README file. When running `flask initdb`, I get the following error:

```
Traceback (most recent call last):
  File "/home/nelson/apps/miniconda/bin/flask", line 6, in <module>
    sys.exit(flask.cli.main())
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/flask/cli.py", line 478, in main
    cli.main(args=args, prog_name=name)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/flask/cli.py", line 345, in main
    return AppGroup.main(self, *args, **kwargs)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/click/core.py", line 1055, in invoke
    cmd_name, cmd, args = self.resolve_command(ctx, args)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/click/core.py", line 1094, in resolve_command
    cmd = self.get_command(ctx, cmd_name)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/flask/cli.py", line 316, in get_command
    rv = info.load_app().cli.get_command(ctx, name)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/flask/cli.py", line 209, in load_app
    rv = locate_app(self.app_import_path)
  File "/home/nelson/apps/miniconda/lib/python3.4/site-packages/flask/cli.py", line 89, in locate_app
    __import__(module)
ImportError: No module named 'flaskr'
```

I'm using an Anaconda Python install on both Windows 10 (Python 3.5) and Linux (Python 3.4) with either the stock flask 0.11 install or an install from the latest git repo.  I'm running this code from appropriate directory and am able to import flaskr or minitwit from a terminal Python session. I can run the other examples that do not require the flask script.

untitaker:
The flaskr example should be modified to include a proper `setup.py`.

rnelsonchem:
Okay. I'm trying to figure this out, but I hit a snag. If I get this working, I could file a PR.

I moved flaskr.py into a new folder "flaskr" and created an empty \_\_init\_\_.py file. I then created the following setup.py file:

```
[...]

PR #1914, opened 2016-06-15
 Fix problems with examples and CLI #1902

ArchRoller:


ThiefMaster:
Please don't use `master` in your repos for PRs, and if you do rebase instead of doing merge commits. Otherwise you quickly get a mess that'd propagate into the upstream repo if not rebased/squashed before merging.

PR #1921, opened 2016-06-21
'be' was omitted

bngsudheer:
'be' was omitted


ThiefMaster:
looks ok, but just as a recommendation for the future: if the first line of the commit message explains everything it just adds noise to repeat the same text again in the detailed commit message.

bngsudheer:
Gotcha :)

Issue #1919, opened 2016-06-20
Remove easy_install from installation docs


jeffwidman:
As pointed out in #1915, http://flask.pocoo.org/docs/0.11/installation/ mentions both `pip` and `easy_install`. 

`pip` seems to be the canonical way to manage packages these days, so I vote we drop `easy_install` from the install docs.

WenchenLi:
vote for ```pip``` as well! It's well managed than `easy_install`

jeffwidman:
Fixed in #1920

PR #1920, opened 2016-06-20
Update installation documentation for using 'pip' command

alexyer:
Fix for #1919

PR #1927, opened 2016-06-23
Removed easy_install method from installation docs

shubhamoli:
As per issue #1919, documentation for installing **virtualenv** via **easy_install** has been successfully removed from **docs/installation.rst** file.


jeffwidman:
Thanks, afraid this is a dupe of #1920 though.

PR #1934, opened 2016-06-29
Helper method send_file not closing file handle.

pykaso:


davidism:
The open file handle is still being used by `wrap_file`. We can't close the file until it has been sent. This is handled by the WSGI server, which calls `close` on the `Response` object, which calls `close` on the wrapped file.

PR #1936, opened 2016-06-29
Add the ability to combine MethodViews

acatton:
I just had this bug on my application, let me give you this example:

``` python
class SingleObjectMixin(object):
    def get_object(self):
        # implementation detail

class ShowView(SingleObjectMixin, MethodView):
    def get(self):
        return self.render_response(obj=self.get_object())

class DeleteView(SingleObjectMixin, MethodView):
    def delete(self):
        orm.delete(self.get_object())
        return self.redirect_to_list()

class ShowDeleteView(ShowView, DeleteView):
    pass
```

Because the metaclass for `MethodView` defines a `methods` property on each view, inheriting from two `MethodView` doesn't combine the supported methods.

Right now you have to do:

``` python
class ShowDeleteView(ShowView, DeleteView):
    methods = ['GET', 'DELETE']
```

which is explicit, but which I think is not how `MethodView` was intended to work.

Of course this fix, does allows you to inherit from two `MethodView` and override the supported methods.

### DIFF
--- a/flask/views.py
+++ b/flask/views.py
@@ -102,12 +102,16 @@ class View(object):
         return view
 
 
+def get_methods(cls):
+    return getattr(cls, 'methods', []) or []
+
+
 class MethodViewType(type):
 
     def __new__(cls, name, bases, d):
         rv = type.__new__(cls, name, bases, d)
         if 'methods' not in d:
-            methods = set(rv.methods or [])
+            methods = set(m for b in bases for m in get_methods(b))
             for key in d:
                 if key in http_method_funcs:
                     methods.add(key.upper())
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -160,3 +160,45 @@ def test_endpoint_override():
 
     # But these tests should still pass. We just log a warning.
     common_test(app)
+
+def test_multiple_inheritance():
+    app = flask.Flask(__name__)
+
+    class GetView(flask.views.MethodView):
+        def get(self):
+            return 'GET'
+
+    class DeleteView(flask.views.MethodView):
+        def delete(self):
+            return 'DELETE'
+
+    class GetDeleteView(GetView, DeleteView):
+        pass
+
+    app.add_url_rule('/', view_func=GetDeleteView.as_view('index'))
+
+    c = app.test_client()
+    assert c.get('/').data == b'GET'
+    assert c.delete('/').data == b'DELETE'
+    assert sorted(GetDeleteView.methods) == ['DELETE', 'GET']
+
+def test_remove_method_from_parent():
+    app = flask.Flask(__name__)
+
+    class GetView(flask.views.MethodView):
+        def get(self):
+            return 'GET'
+
+    class OtherView(flask.views.MethodView):
+        def post(self):
+            return 'POST'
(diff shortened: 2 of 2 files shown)


mdomke:
+1

acatton:
Is there any update on this? Or is this a bad idea?

flask/views.py
```diff
@@ -102,12 +102,16 @@ def view(*args, **kwargs):
...
+    return getattr(cls, 'methods', []) or []
+
+
 class MethodViewType(type):
 
     def __new__(cls, name, bases, d):
         rv = type.__new__(cls, name, bases, d)
         if 'methods' not in d:
-            methods = set(rv.methods or [])
+            methods = set(m for b in bases for m in get_methods(b))
```
RonnyPfannschmidt: this always breaks for deeper inheritance

acatton: I'm not sure to understand. Let's assume this deep inheritance:

```python
class MaternalGrandMother(MethodView):
    def get(self):
        # implementation details

class MaternalGrandFather(MethodView):
    def post(self):
        # implementation details

class Mother(MaternalGrandMother, MaternalGrandFather):
    # method = ['POST', 'GET']

class PaternalGrandMother(MethodView):
    def delete(self):
        # implementation details
    def patch(self):
        # implementation details

class PaternalGrandFather(MethodView):
    def put(self):
        # implementation

class Father(PaternalGrandMother, PaternalGrandFather):
    methods = ['PUT', 'DELETE']  # Don't inherit PATCH
```

So now here's the behavior I expect, and I don't think, and that's what I meant in this code:

```python
class A(Father, Mother):
    # methods = ['PUT', 'DELETE', 'GET', 'POST']

class B(Mother, PaternalGrandFather, PaternalGrandMother):
    # methods = ['PUT', 'DELETE', 'PATCH', 'GET', 'POST']

class C(Mother, MaternalGrandMother, PaternalGrandMother):
    # methods = ['POST', 'GET', 'DELETE', 'PATCH']
```

Can you give me an example that I'm missing?

acatton:
Why is reviewable failing?

RonnyPfannschmidt:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pallets/flask/1936#-:-KUq0YJTMgCNhFDk8BKE:bnfp4nl)*

RonnyPfannschmidt:
wow, reviewable is broken ^^

flask/views.py
acatton: That's is a good point, I'll add a doc fix for this in this pull request and ping you once this is done.

davidism:
I don't think `methods` should be used while collecting the supported methods from the MRO. It doesn't really make sense to not using a method if the view defines it, but if you do want to do that, I think you should specify `methods` explicitly in the subclass.

davidism:
Given that's what your test actually tests, I'm going to adjust this to collect all methods provided by the MRO unless explicitly set.

davidism:
Continued in #2252.

PR #1937, opened 2016-06-30
FLASK_APP should be set to [app].py, not just [app]

joannejang:
Setting FLASK_APP to flaskr/minitwit as instructed in the README will cause an import error.


untitaker:
Hello @joannejang, sorry for the late reply: As part of #1945 and #1954 I think we're going to address this issue.

Issue #1933, opened 2016-06-27
JSON Parsing on empty request content with RequestParser


thanhtungka91:
From flask-restful/flask-restful#163.

tdpsk:
I agree that this is weird but acceptable behavior on the client side. Therefore I created a pull request to fix this. However there is a larger discussion surrounding this, see pull request.

untitaker:
This behavior is completely incompliant, if you really need this you should rather catch ValueError. But an empty file is not valid JSON.

PR #1939, opened 2016-07-01
Check for empty request data prior to loading JSON, fix for #1933

tdpsk:


tdpsk:
Due to the error reported in #1933 concerning some widely used clients I fixed this for GET requests.

But actually nowhere in RFC2616 does it say POST and / or PUT must have a body content. However there are currently test cases testing this behavior and I did not want to introduce a breaking change in a pull request without prior discussion. Can someone who is familiar with the discussion fill me in why it is required that a POST request with an empty body fails?

guyskk:
@tdpsk If request have no body content, the client shouldn't set `Content-Type: application/json` in request headers because empty string is not valid JSON. Thus, server should use `request.form` in that case, and use `request.json` only if client set  `Content-Type: application/json`.

PR #1940, opened 2016-07-01
Added a property is_body_json

smitthakkar96:


ThiefMaster:
What's the advantage of this? `request.get_json()` should cover all you need.

smitthakkar96:
@ThiefMaster I think we need a property that validates the json of body as empty body with headers application/json results into 400.

ThiefMaster:
You can catch the exception or use `silent=True`

smitthakkar96:
I didn't notice silent parameter, btw thanks.

Issue #1906, opened 2016-06-11
SessionInterface shouldn't hard code a fix for localhost when it comes to cookies


nickjj:
The `get_cookie_domain` function specifically fixes issues with webkit browsers by doing this:

```python
# https://github.com/pallets/flask/blob/0.11.1/flask/sessions.py#L198-222
if rv == '.localhost':
    rv = None
```

The problem is that nowadays many people are using Docker and other techniques (Vagrant, etc.) to create a better development environment so they are no longer running Flask on `localhost`.

In Docker's case, you'd likely end up with an IP address based host name which means you'd set your `SERVER_NAME` to something like `192.168.99.100`.

At this point Chrome and other webkit based browsers will freak out by reject setting the cookie which in turn will make your Flask app throw CSRF token related issues when submitting forms.

Since Flask is already hard coding a fix for this, I think you should allow users to configure which domain name gets skipped with a proper config option.

This could be named something like `SESSION_COOKIE_IGNORE_DOMAIN` (feel free to improve the name :)). It could default to localhost so that it's backwards compatible and nothing changes for most users.

Right now the work around is to modify your `/etc/hosts` file to map an IP address to a FQDN with a TLD if you're stuck using an IP based `SERVER_NAME`. That seems a little lame and preventable at the Flask level.

What do you think?

untitaker:
Better workarounds:

* explicitly set `SESSION_COOKIE_DOMAIN`
* Subclass `SessionInterface` and set `app.session_interface` to the new subclass

nickjj:
If you set `SESSION_COOKIE_DOMAIN` to an IP address wouldn't Chrome still reject it?

untitaker:
You can set it to `None` though.

I don't think this can be properly worked around at a Flask level. And actually I consider the workaround of assigning your dev server a proper hostname to be a good solution.

nickjj:
Setting it to `None` results in the same issue as not setting it.

untitaker:
Sorry, I misread the code

untitaker:
Actually, as far as I understand, `SERVER_NAME` does not support IP addresses at all, and setting cookies on IP addresses is not supported at all: https://tools.ietf.org/html/rfc6265#section-5.1.3

nickjj:
I have a bunch of students using an IP based `SERVER_NAME` and cookies are correctly set as long as they don't use a webkit based browser.

A couple of months ago I swear the spec was different. It was written such that IP addresses were valid.

[...]

PR #1946, opened 2016-07-06
Added 127.0.0.0/8 to the session cookie 'localhost' exception.

honsiorovskyi:
Added 127.0.0.0/8 IP range to the expression which checks whether the session cookie domain is 'localhost' in the method `get_cookie_domain`.

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/1946)

### DIFF
--- a/flask/sessions.py
+++ b/flask/sessions.py
@@ -9,6 +9,7 @@
     :license: BSD, see LICENSE for more details.
 """
 
+import re
 import uuid
 import hashlib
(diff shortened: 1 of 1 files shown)


untitaker:
I feel like this is an incomplete fix for https://github.com/pallets/flask/issues/1906

untitaker:
Actually, `SERVER_NAME` does not support IP addresses.

Issue #1926, opened 2016-06-23
Unhelpful Error Message on FLASK_APP to Bad Path


cancan101:
If the path specified in `FLASK_APP` does not exist (for example it is a typo), then the unhelpful error message is shown:
```
ImportError: Import by filename is not supported.
```

This is even though I have:
```
FLASK_APP=alex/server2.py
```
which is "path-like" and not module like.

Perhaps [this line](https://github.com/pallets/flask/blob/cd1a9b7d545a9566b6d8ab861572f2e80c67916e/flask/cli.py#L106) should be modified to return `True` for file-like objects.

robbielaldrich:
How about 

```python
if os.path.isfile(app) or os.path.isdir(app):
    return prepare_exec_for_file(app)
```

untitaker:
Fixed by https://github.com/pallets/flask/pull/1950

Issue #1829, opened 2016-05-29
"[Errno 2] No such file or directory" on flask run with FLASK_DEBUG=1


yoavram:
I get the following `[Errno 2] No such file or directory` with **flask 0.11** & werkzeug 0.11.10 on Windows 7 64bit with Python 3.5 but only when `FLASK_DEBUG=1`. 

``` sh
D:\tmp
λ python --version
Python 3.5.1 :: Continuum Analytics, Inc.

D:\tmp
λ python -c "import hello_flask"

D:\tmp
λ set FLASK_APP=hello_flask.py

D:\tmp
λ flask run
 * Serving Flask app "hello_flask"
 * Forcing debug mode off
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)

D:\tmp
λ set FLASK_DEBUG=1

D:\tmp
λ flask run
 * Serving Flask app "hello_flask"
 * Forcing debug mode on
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
 * Restarting with windowsapi reloader
d:\anaconda3\envs\flask\python.exe: can't open file 'D:\Anaconda3\envs\flask\Scripts\flask': [Errno 2] No such file or directory
```

`hello_flask.py` is the minimal app from the [docs](http://flask.pocoo.org/docs/0.11/quickstart/#a-minimal-application).

gledi:
I faced the same error. However it works when I run it as a module:

```console
$ python -m flask run
```

mitsuhiko:
This might fall under the same issue as #1847.

andersea:
There is a question on stackoverflow for this issue:

http://stackoverflow.com/questions/37650208/flask-cant-find-app-file

dgbillotte:
I'm having the same problem on Windows Server with Flask 0.11.1 except when I try

```
$ python -m flask run
```
I get: ImportError: cannot import name Flask

Currently I can only make it work with FLASK_DEBUG=0

I'd be happy to provide any other details...

TomIsPrettyCool:
This is caused as python is looking for a script named flask, but on windows it is called flask.exe.

As a temporary fix you can copy and paste flask.exe, and just rename it 'flask'

yoavram:
Good call @TomIsPrettyCoool.
So this is a bug in the reloader - note that windows finds `flask` when prompted to, although it's `flask.exe`. the werkzeug reloader probably calls `flask` instead of `flask.exe`.
[...]

PR #1950, opened 2016-07-08
adding in try around __import__ to catch invalid files/paths

nateprewitt:
I'll start this off by saying this may be a naive approach to what appears to be a common thread between a few recent issues (#1926, #1902, #1829). I'm attempting to solve the issue of mistyped or improperly defined app paths reaching the __import__ call for the flask app server. Given the way `cli.py` is currently structured, I can't find a bulletproof way to ensure only correct paths/module names reach the import. 

If that's the case, I believe the best response is to simply catch the error and return something more descriptive than `ImportError: Import by filename is not supported.` for debugging purposes.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -86,7 +86,13 @@ def locate_app(app_id):
         module = app_id
         app_obj = None
 
-    __import__(module)
+    try:
+        __import__(module)
+    except ImportError:
+        raise NoAppException('The file/path provided (%s) does not appear to '
+                             'exist.  Please verify the path is correct.  If '
+                             'app is not on PYTHONPATH, ensure the extension '
+                             'is .py' % module)
     mod = sys.modules[module]
     if app_obj is None:
         app = find_best_app(mod)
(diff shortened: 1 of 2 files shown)


untitaker:
I'm all for better error messages, but this is a bit misleading: Appending `.py` is not necessarily required, only if you want to import `.py` files whose parent dirs are not in your PYTHONPATH.

If you have your flask app `foobar` installed as package, `foobar` should suffice, without file extension.

nateprewitt:
Thanks for clarifying that @untitaker, I updated the exception message to reflect that a little closer.

untitaker:
Why did you close this?

nateprewitt:
Sorry, wasn't sure if it was really going to provide anything useful. Didn't want to waste any of your time. I'll reopen if needed.

untitaker:
I'll review whenever I get around to it, because I do this in my freetime.

nateprewitt:
@untitaker absolutely, no rush. I was just trying to clean up on my end, didn't mean to kill this prematurely.

untitaker:
Excellent, thanks!

PR #1951, opened 2016-07-09
close <li> tag in lines (16,18)

ahmedakef:
screenshot from 2016-07-09 21-51-02
i noticed that <li> tag haven't closed in lines 15,18 
which is bad practice as if i put "some thing : h1> some text /h1>" in the text-area  all the other articles become h1> so big and color blue

### DIFF
--- a/examples/flaskr/flaskr/templates/show_entries.html
+++ b/examples/flaskr/flaskr/templates/show_entries.html
@@ -13,9 +13,9 @@
   {% endif %}
   <ul class="entries">
   {% for entry in entries %}
-    <li><h2>{{ entry.title }}</h2>{{ entry.text|safe }}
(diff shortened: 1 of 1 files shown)


untitaker:
Thanks!

davidism:
Didn't we just go through this and make @mitsuhiko sad a couple weeks ago?  Closing tags are not required for `li`, and the style throughout is to omit them.

davidism:
The issue here is marking user input safe, not omitting optional close tags.

untitaker:
I think it's warranted here. The (non-security) issue here is that a (accidentally) malformed blog entry can break the layout of the entire site.

untitaker:
(which is possible in either case, but omitting the `li` tag makes it too easy)

PR #1945, opened 2016-07-06
Address #1902: Converts example/flaskr to have a setup.py

wgwz:
Updates the flaskr example, as well as the tutorial docs, for consistency. I tested that following the docs linearly works as expected.

PR #1953, opened 2016-07-12
Update minitwit & improve testing for examples

wgwz:
* Related to #1945
* Re-works minitwit to be installed and run as:

    pip install --editable .
    export FLASK_APP=minitwit.minitwit
    export FLASK_DEBUG=1
    flask initdb
    flask run

* added flaskr and minitwit to norecursedirs
  * tests not properly run when using pytest standards
  * see: http://stackoverflow.com/questions/38313171/configuring-pytest-with-installable-examples-in-a-project
* Both flaskr and minitwit now follow pytest standards.
* Tests can for them as `py.test` or `python setup.py test`


wgwz:
I chose the wrong base. Re-filed on the 0.11-maintenance branch.

jeffwidman:
In general, we commit against master and if needed a maintainer will backport to a maintenance branch. That way we avoid diverging from master as much as possible.

davidism:
@jeffwidman no, way easier to commit against what's actually being fixed, then merge the result forward.

jeffwidman:
Hmm, doesn't the problem exist in both master and maintenance branch? 

On the other projects I'm involved with when that's the case we always do the PR against master and then cherrypick as needed to the relevant maintenance branch. I'm fine with either flow, but not understanding the rationale here.

PR #1963, opened 2016-07-27
isolate templates between blueprints

teddy-hoo:
see the test


ThiefMaster:
This is a breaking change and in many cases not what you want, e.g. if you merely use blueprints to structure your application.

ThiefMaster:
If you need this in your application, override the jinja loader.

teddy-hoo:
Do you mean flask should not support this feature?  or my code should consider more cases?
Thx a lot!

ThiefMaster:
the former - it's something better done inside your application

teddy-hoo:
I can not understand.
I use blueprint as component of my app, but one blueprint can use templates in other blueprints. Why not  the components be independent?

ThiefMaster:
Because that's only one way to use blueprints. For example, your blueprint most likely wants to use templates from your app (like your base layout template) and depending on how you structure things some components require another component and also use some of that component's templates.

davidism:
#1537

PR #1965, opened 2016-07-28
Added a missing module to import statement

vioan:


davidism:
`after_this_request` was written a couple sections up the page, it's not an import.

vioan:
then perhaps the function should be added to the example as well? Because if somebody is just trying to copy/paste the example code, will not work as it is right now.

ThiefMaster:
@davidism Actually, it is an import.

vioan:
of course it is an import

untitaker:
Reopening as a reminder, but the section needs a larger rewrite to make use of Flask's own version.

PR #1977, opened 2016-08-03
Set __qualname__ on class-based view function

taion:
On Python 3, New Relic uses `__qualname__` to trace transactions. Without setting `__qualname__` in `as_view()`, I don't get meaningful accounting of transactions in New Relic on my class-based views.

`__qualname__` is meaningless in Python 2, but I think it's just a harmless no-op in this context.

### DIFF
--- a/flask/views.py
+++ b/flask/views.py
@@ -84,7 +84,7 @@ class View(object):
             return self.dispatch_request(*args, **kwargs)
 
         if cls.decorators:
-            view.__name__ = name
+            view.__qualname__ = view.__name__ = name
             view.__module__ = cls.__module__
             for decorator in cls.decorators:
                 view = decorator(view)
@@ -95,7 +95,7 @@ class View(object):
         # the view class so you can actually replace it with something else
         # for testing purposes and debugging.
         view.view_class = cls
-        view.__name__ = name
+        view.__qualname__ = view.__name__ = name
         view.__doc__ = cls.__doc__
         view.__module__ = cls.__module__
         view.methods = cls.methods
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -160,3 +160,16 @@ def test_endpoint_override():
 
     # But these tests should still pass. We just log a warning.
     common_test(app)
+
+def test_view_name():
+    app = flask.Flask(__name__)
+
+    class Index(flask.views.View):
+        methods = ['GET', 'POST']
+        def dispatch_request(self):
+            return flask.request.method
+
+    view_func = Index.as_view('index')
+
+    assert view_func.__name__ == 'index'
+    assert view_func.__qualname__ == 'index'


davidism:
Would it make more sense to try to set it to `cls.__qualname__` (if it's defined), in case the class is nested?

taion:
I feel like it would be a bit inconsistent to set `view.__qualname__` to `cls.__qualname__` but not set `view.__name__` to `cls.__name__`.

In my specific case these three strings are all actually the same, though, so I'd be fine with making that change if you think it'd make more sense.

davidism:
Could you go into more detail about why the current result isn't meaningful?

Looks like the qualname would be `'View.as_view.<locals>.view'`, which I guess would all get lumped together as the same thing in New Relic?

Maybe `View.as_view.<locals>.<name>` would be more accurate.

taion:
That's exactly it.

On Python 2, results would show up under e.g. `/my_app.views:WidgetView`, `/my_app.views:WidgetListView`, `/my_app.views:OtherView`, &c.

After upgrading to Python 3, all results now show up under `/my_app.views:View.as_view.<locals>.view`.

"How it looks in New Relic" is a bit of a specific case, but I think it makes sense to think of the view function as being an implementation detail of how class-based views work in Flask, and that in this context it doesn't help much to directly expose it to the user. At least, that's how it works with Python 2 right now, given that this code also overrides e.g. `__module__` on the view function.

davidism:
How's it getting that name in Python 2?  I would expect it to be `/my_app.views:{name}`, there's no standard attribute pointing back at the class from the generated view function.

Any other @pallets/devteam-flask want to weigh in?  I'm leaning towards accounting for the existing qualname, it's the most descriptive.

~~~python
view.__qualname__ = '.'.join(getattr(self.__qualname__, '').rsplit('.', 1)[-1], name)
~~~

taion:
Yes, to clarify, I am passing `name='WidgetView'`, so for me, `name`, `cls.__name__`, and `cls.__qualname__` are all the same (since my class isn't nested).

My argument against using `__qualname__` there is that you already override `view.__module__`, so the fully qualified path would be something like `my_app.views.View.as_view.<locals>.WidgetView`, which I think would be confusing, since `View` is not in `my_app.views`.

davidism:
I see what you mean now, since `View` subclasses don't typically override `as_view` the qualname still points at the parent class.

taion:
Worse – `__qualname__` proper is `View.as_view...`, so `view.__module__ + '.' + view.__qualname__` could actually give you something non-existent, since the module could be importing `MethodView` instead of `View`.

taion:
Any updates here? Should I work around this on my end for the time being?

davidism:
I'm sure the "subclass method has superclass qualname" issue is present elsewhere in New Relic, how does it handle this there?  I'd rather return a value that matches the spec, even if New Relic doesn't handle it well, rather than change what qualname means.

untitaker:
What if we do: `view.__qualname__ = view.__qualname__.rsplit('.', 1)[-1] + '.' + name`? This differs from @davidism's first proposal since it reuses the original qualname of the function.

taion:
I think the value of preserving `__qualname__` here is low, because the code there is already replacing `view.__module__`.

If anything I'd invert things – if my view is `my_app.views:MyView`, and given the existing behavior that sets `view.__module__` to `my_app.views`, and assuming we have `name` as `my_view`, I think we'd want `__qualname__` to be either:

- `MyView`
- `my_view`

I think including `View` in `__qualname__` wouldn't be useful because this code already overwrites `__module__`.

untitaker:
We should make `__qualname__` whatever the PEP requires it to be, which is why I think replacing the last part is most conformant.

taion:
Reading through PEP 395, I'm starting to think that maybe it'd be best to drop this PR leave `__qualname__` untouched. I'm going to take a closer look at what exactly New Relic is doing here and see if there isn't a better way to do this.

davidism:
It sounds like they expect `__qualname__` to be the closest object that a function "belongs" to, when it actually means the closest object where it is actually defined.  It might be better to work with them first to fix that disconnect, then figure out if this needs to change.  Closing this for now.

taion:
Yup. Thanks for your time here.

taion:
New Relic handled this on their end in https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-276055.

PR #1981, opened 2016-08-11
Fix line 31 for conformity with PEP-8

blancheta:
Hi,
I fixed this line to conform with PEP-8.
Regards,


davidism:
Thanks, but we're handling any conversion to PEP-8 internally.

PR #1984, opened 2016-08-19
JSON request/response mixin

untitaker:
Continuing #1416

All inline discussions have been resolved.

### DIFF
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -18,13 +18,6 @@ from .globals import _request_ctx_stack
 _missing = object()
 
 
-def _get_data(req):
-    getter = getattr(req, 'get_data', None)
-    if getter is not None:
-        return getter()
-    return req.data
-
-
 class JSONMixin(object):
     """Mixin for both request and response classes to provide JSON parsing
     capabilities.
@@ -57,6 +50,12 @@ class JSONMixin(object):
                                 'Use get_json() instead.'), stacklevel=2)
         return self.get_json()
 
+    def _get_data_for_json(req, cache):
+        getter = getattr(req, 'get_data', None)
+        if getter is not None:
+            return getter(cache=cache)
+        return req.data
+
     def get_json(self, force=False, silent=False, cache=True):
         """Parses the incoming JSON request data and returns it.  By default
         this function will return ``None`` if the mimetype is not
@@ -84,7 +83,7 @@ class JSONMixin(object):
         # been encoded correctly as well.
         charset = self.mimetype_params.get('charset')
         try:
-            data = _get_data(self)
+            data = self._get_data_for_json(cache)
             if charset is not None:
                 rv = json.loads(data, encoding=charset)
             else:
@@ -215,3 +214,10 @@ class Response(ResponseBase, JSONMixin):
     set :attr:`~flask.Flask.response_class` to your subclass.
     """
     default_mimetype = 'text/html'
+
+    def _get_data_for_json(req, cache):
+        # Ignore the cache flag since response doesn't support it
+        getter = getattr(req, 'get_data', None)
+        if getter is not None:
+            return getter()
+        return req.data
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -25,15 +25,88 @@ def _get_data(req, cache):
     return req.data
 
 
-def _is_mimetype_json(mimetype):
-    if mimetype == 'application/json':
-        return True
-    if mimetype.startswith('application/') and mimetype.endswith('+json'):
-        return True
-    return False
+class JSONMixin(object):
+    """Mixin for both request and response classes to provide JSON parsing
+    capabilities.
+
+    .. versionadded:: 0.12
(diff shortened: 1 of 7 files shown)


untitaker:
@homu r+

flask/wrappers.py
```diff
@@ -13,19 +13,110 @@
...
+        if not (force or self.is_json):
+            return None
+
+        # We accept MIME charset header against the specification as certain
+        # clients have been using this in the past.  For responses, we assume
+        # that if the response charset was set explicitly then the data had
+        # been encoded correctly as well.
+        charset = self.mimetype_params.get('charset')
+        try:
+            data = self._get_data_for_json(cache)
```
untitaker: Do we even want to pass `cache` on? If we do this, by default both `data` and `json` are stored on the request object. While this leads to the least surprise for newcomers (no empty return values), it is very memory inefficient.

flask/wrappers.py
```diff
@@ -13,19 +13,110 @@
...
+        """
+        from warnings import warn
+        warn(DeprecationWarning(
+            'json is deprecated.  Use get_json() instead.'), stacklevel=2)
+        return self.get_json()
+
+    def _get_data_for_json(self, cache):
+        getter = getattr(self, 'get_data', None)
+        if getter is not None:
+            return getter(cache=cache)
```
untitaker: We have a problem here: `werkzeug.Response.get_data` doesn't accept a `cache` flag.

mitsuhiko: Can we just add the flag there and require a newer werkzeug?

untitaker: Yes, see https://github.com/pallets/werkzeug/pull/992

sevankj:
Reviewed 5 of 7 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pallets/flask/1984)*

JordanP:
Any progress on this ? Could you rebase ?

I've been looking for better JSON support (a-la python-requests) and there's already several bug reports and PR in flask and werkzeug about this. This looks like a popular demand :)

jwg4:
This change would make a big difference to us. Sorry to see that it is nearly two years old, can I do something to help?

flask/wrappers.py
JordanP: The `cache` arg is already defined on the original `get_json()` method, so I don't think removing it or changing its behavior should be part of this PR.

leejunhoan:
Reviewed 5 of 7 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pallets/flask/1984)*

ramnes:
This request merged would be a huge relief for me and my team also.

@mitsuhiko, @untitaker, how can we make this happen?

kislyuk:
+1 for this feature. Any way to help this get in?

kennethreitz:
needs a rebase!

kennethreitz:
Maybe it would be better to use Requests as the test client :)

flask/wrappers.py
```diff
@@ -13,19 +13,110 @@
...
+
+    @property
+    def json(self):
+        """If this request/response is in JSON format then this property will
+        contain the parsed JSON data.  Otherwise it will be ``None``.
+
+        The :meth:`get_json` method should be used instead.
+        """
+        from warnings import warn
+        warn(DeprecationWarning(
```
kennethreitz: `.json` is important, and shouldn't be deprecated

kennethreitz: `is_json()` shouldn't be neccessary, as `.json` should return `None` if there is no JSON, correct?

kennethreitz:
closing, old pr, requests is actually a better idea for the client (though i'm sure that's not an obvious choice)

untitaker:
That is only your opinion at the moment, and the Flask docs don't show how to
use it as one. Before we settle on this we should make clear how to actually
use it in tests.

Right now you have to spin up a dev server for this.

kennethreitz:
i definitely don't think requests should go in, i'm mostly joking about that :)

kennethreitz:
may have been premature to close this — i reviewed every line of code, and it seemed solid until i got to the deprecation of `.json`. then, I saw that there may be some issues with werkzeug and saw the merge conflicts so decided to close.

untitaker:
I'm not sure either what to think of that.

We can have (parts of) requests as test client and it would probably be an UX
improvement, but right now I consider the overhead of its entire HTTP stack to
be too much.

kennethreitz:
Well, you'd write a connection adapter that would skip it speaking HTTP and have it talk WSGI instead. Should be <30LOC. It's alright though, people are used to having unique test clients, and people are def used to the one built in here.

davidism:
Continued in #2358.

PR #1990, opened 2016-08-22
Use the Celery result backend setting from the example

sanderl-mediamonks:


ThiefMaster:
Why? I don't think this is the result backend but the backend used for actual tasks.

sanderl-mediamonks:
No, it's the result backend. See http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html#keeping-results

ThiefMaster:
Yeah, saw it after a moment which is why i deleted my comment and merged it

PR #2002, opened 2016-09-01
Clean up tutorial docs for installable app pattern with flaskr

wgwz:
- reading sequentially through the tutorial works.
- fixes references to `export FLASK_APP=flaskr.flaskr`

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2002)

### DIFF
--- a/docs/tutorial/dbinit.rst
+++ b/docs/tutorial/dbinit.rst
@@ -44,6 +44,9 @@ Within the function, you can then access :attr:`flask.g` and other things as
 you might expect.  When the script ends, the application context tears down
 and the database connection is released.
 
+.. note:: In order to use the newly created ``flask initdb`` command, the 
+    application must be re-installed using ``pip install -e .``
+
 You will want to keep an actual function around that initializes the database,
 though, so that we can easily create databases in unit tests later on.  (For
 more information see :ref:`testing`.)
--- a/docs/tutorial/dbinit.rst
+++ b/docs/tutorial/dbinit.rst
@@ -44,9 +44,6 @@ Within the function, you can then access :attr:`flask.g` and other things as
 you might expect.  When the script ends, the application context tears down
 and the database connection is released.
 
-.. note:: In order to use the newly created ``flask initdb`` command, the 
-    application must be re-installed using ``pip install -e .``
-
 You will want to keep an actual function around that initializes the database,
 though, so that we can easily create databases in unit tests later on.  (For
 more information see :ref:`testing`.)
--- a/docs/tutorial/setuptools.rst
+++ b/docs/tutorial/setuptools.rst
@@ -9,19 +9,23 @@ tutorial you will see exactly how to extend the ``flask`` command line
 interface (CLI).
 
 A useful pattern to manage a Flask application is to install your app
-using `setuptools`_.  This involves creating a :file:`setup.py`
-in the projects root directory.  You also need to add an empty
-:file:`__init__.py` file to make the :file:`flaskr/flaskr` directory
-a package.  The code structure at this point should be::
+using `setuptools`_.  This involves creating two new files; 
+:file:`setup.py` and :file:`MANIFEST.in` in the projects root directory.  
+You also need to add an :file:`__init__.py` file to make the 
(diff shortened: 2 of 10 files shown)


davidism:
It would be nice to link to https://packaging.python.org rather than the docs for setuptools, since they're the official comprehensive documentation on creating installable packages.

wgwz:
@davidism does that mean you just want an extra sentence or two worked into the tutorial somewhere with that information? is there a particular instance of mentioning setuptools that you have in mind?

wgwz:
@davidism i think i understand what you want. change the link for setuptools and then add a couple lines describing packaging.python.org.

docs/tutorial/dbinit.rst
```diff
@@ -44,6 +44,9 @@ Within the function, you can then access :attr:`flask.g` and other things as
 you might expect.  When the script ends, the application context tears down
 and the database connection is released.
 
+.. note:: In order to use the newly created ``flask initdb`` command, the 
+    application must be re-installed using ``pip install -e .``
```
untitaker: Probably should come in the previous step. Also a ordering of all commands in the order they have to be typed is probably preferrable.

docs/tutorial/setuptools.rst
```diff
@@ -9,19 +9,23 @@ tutorial you will see exactly how to extend the ``flask`` command line
...
             __init__.py
             /static
             /templates
+            flaskr.py
+            schema.sql
         setup.py
+        MANIFEST.in
 
-The content of the ``setup.py`` file for ``flaskr`` is:
+The content of the ``setup.py`` file for ``flaskr`` will be:
```
untitaker: Not sure why this change, everything else is in present tense.

docs/tutorial/testing.rst
```diff
@@ -9,45 +9,53 @@ modifications in the future.  The application above is used as a basic
...
-======================
+Adding tests to flaskr
+----------------------
 
-Assuming you have seen the testing section above and have either written
+Assuming you have seen the :ref:`testing` section and have either written
 your own tests for ``flaskr`` or have followed along with the examples
 provided, you might be wondering about ways to organize the project.
 
-One possible and recommended project structure is::
```
untitaker: Why change this?

untitaker:
I'm a bit unhappy how a lot of changes are squashed into one commit.

wgwz:
@untitaker Thanks for the feedback, I'll definitely make the suggested changes. 

> I'm a bit unhappy about how a lot of changes are squashed into one commit.

What should I do next time? One commit/file? I'm a bit confused, I've seen comments on other PRs suggesting to squash commits. Let me know if you would like me to refile the PR.

untitaker:
In principle it's better to have separate PRs for unrelated changes, and
separate commits for sligthly related or interdependent changes. Sometimes we
ask people to squash all commits because there are a lot of commits that fix
typos in previous ones, or make changes wished by the reviewer.

docs/tutorial/testing.rst
wgwz: Which part exactly? "and recommended"? I did not want to say something that wasn't necessarily true. But if it is a recommended structure, sure I will put that back in. Also I think the "----" makes more sense then "====", for lack of better terms using a subsection as opposed to another section.

docs/tutorial/dbinit.rst
wgwz: Well, technically at its current state the commands are listed in the order they have to be typed and it is functional, when followed through linearly. But it would be nice to avoid the repetition. Which I guess is what you are getting at. But correct me if I am wrong.

untitaker: My point is twofold:

* We can have this instruction as normal text instead of a note/alert
* This instruction is not necessarily part of the db initialization. In particular, `pip install -e .` is already required for Step 2 (`flask run`)

docs/tutorial/testing.rst
untitaker: Yeah this application layout is now recommended, sorry about the confusion. I only meant this line specifically, not the headlines or anything.

untitaker:
Thanks!

Issue #2007, opened 2016-09-06
Prevent setting `SERVER_NAME` and `SESSION_COOKIE_DOMAIN` to an IP address


untitaker:
See #1946 and #1906. I think we should prevent setting those things to IP addresses in the first place and only allow `None` and domain names != `localhost`.

untitaker:
From #1955: We should also check that no port is contained.

davidism:
Given that `SERVER_NAME` is used with `url_for(..., _external=True)` when not in a request context, preventing ports would break links.  Maybe there's a more correct way to do external links outside request contexts, but I couldn't find it at the time.

untitaker:
>Given that SERVER_NAME is used with url_for(..., _external=True) when not in a request context, preventing ports would break links.

It rather seems to me that using ports there only accidentally worked and was never expected by the API authors. `SERVER_NAME` is [passed in Flask](https://github.com/pallets/flask/blob/9e41eca7f839db16cdf064f827fada0cabf6f54f/flask/app.py#L1781), yes. However, I don't think Werkzeug's `Map.bind` expects a port either, since 1.) the parameter is also called `server_name` 2.) `server_name` [is then idna-encoded](https://github.com/untitaker/werkzeug/blob/b617036ff9173fdc527177493ac518dae01503b4/werkzeug/routing.py#L1245), which is nonsensical for host + port combinations and actually breaks for certain hostnames:

    >>> 'foobar:443'.encode('idna')
    'foobar:443'
    >>> 'fööbar:443'.encode('idna')
    b'xn--fbar:443-n4aa'

However, apparently `app.run` derives its default port from `SERVER_NAME`.

This is so broken...

davidism:
This and the propagate exceptions issue are both cases of one option doing more than it should.  We should probably do a more thorough review of what each option does and maybe add some more to disambiguate stuff before 1.0.

untitaker:
Something like:

* Prevent ports in `SERVER_NAME`
* Add `SERVER_PORT`
* Change `app.run` behavior and `url_for` implementation accordingly.

davidism:
Might be nice to make `SERVER_NAME` separate from `bind`, so that multiple domains / proxies don't cause 404s when you're really just trying to use `SERVER_NAME` for `url_for`.

untitaker:
I don't understand, SERVER_NAME is already the frontend hostname.

On 10 September 2016 20:17:51 CEST, David Lord <notifications@github.com> wrote:
[...]

PR #2010, opened 2016-09-07
changing ip address to string for localhost

infinite-Joy:
For https://github.com/pallets/flask/issues/2007

which now is more inline with docs in the home page of http://flask.pocoo.org/

Before the change:
```
python hello_flask.py
127.0.0.1 - - [07/Sep/2016 11:08:55] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [07/Sep/2016 11:08:55] "GET /favicon.ico HTTP/1.1" 404 -
127.0.0.1 - - [07/Sep/2016 11:09:14] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [07/Sep/2016 11:09:14] "GET /favicon.ico HTTP/1.1" 404 -
```

After the change:
```
python hello_flask.py
 * Running on http://localhost:5000/ (Press CTRL+C to quit)
127.0.0.1 - - [07/Sep/2016 11:09:42] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [07/Sep/2016 11:09:59] "GET / HTTP/1.1" 200 -
```

As you can see its not "getting" the favicon. Is that an issue?

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2010)

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -75,8 +75,7 @@ class Flask(_PackageBoundObject):
     The name of the package is used to resolve resources from inside the
     package or the folder the module is contained in depending on if the
     package parameter resolves to an actual python package (a folder with
-    an :file:`__init__.py` file inside) or a standard module
-    (just a ``.py`` file).
+    an :file:`__init__.py` file inside) or a standard module (just a ``.py`` file).
 
     For more information about resource loading, see :func:`open_resource`.
 
@@ -98,7 +97,7 @@ class Flask(_PackageBoundObject):
         using a package, it's usually recommended to hardcode the name of
         your package there.
 
-        e.g. if your application is defined in :file:`yourapplication/app.py`
+        For example if your application is defined in :file:`yourapplication/app.py`
         you should create it with one of the two versions below::
 
             app = Flask('yourapplication')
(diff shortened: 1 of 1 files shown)


ThiefMaster:
> As you can see its not "getting" the favicon. Is that an issue?

Probably just means your browser already has one cached.

flask/app.py
```diff
@@ -827,7 +829,7 @@ def run(self, host=None, port=None, debug=None, **options):
         """
         from werkzeug.serving import run_simple
         if host is None:
-            host = '127.0.0.1'
+            host = 'localhost'
```
ThiefMaster: This is not equivalent to `127.0.0.1` - using `localhost` may result in `::1` being used instead isince `localhost` has both IPv4 and IPv6 addresses

davidism:
Please remove the unrelated grammar "fixes", they're not necessary.

infinite-Joy:
done

davidism:
From the linked issue:

> I think we should only allow `None` and `!= 'localhost'`.

So what exactly is this addressing by making it equal `localhost`?

untitaker:
I think y'all are misunderstanding #2007. I want to show an error when the user tries to set `app.config['SERVER_NAME']` to an IP address.

PR #2015, opened 2016-09-08
fixed deprecated syntax in setup.cfg

drewja:
I was inspired to fix this by the following warning in pytest:
"[pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead."

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2015)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,5 +4,5 @@ release = egg_info -RDb ''
 [wheel]
 universal = 1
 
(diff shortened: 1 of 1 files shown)


untitaker:
>I was inspired to fix this by the following warning in pytest:

Truly inspiring :P

Thanks!

PR #2018, opened 2016-09-09
Fix fileuploads manual page redirect URL

aleksihakli:
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2018)


ThiefMaster:
I think the idea is that you would create another endpoint that shows a success message or a list of uploaded files.

Issue #1879, opened 2016-06-03
test_main_module_paths fails on RHEL and Ubuntu when using Amazon EC2


karlw00t:
Running the unit tests against HEAD (e048aa4e19), I'm able to reproduce a single failing unit tests in:
> tests/test_instance_config.py::test_main_module_paths

I'm able to reproduce the bug on Amazon Linux 2016.03.1 and RHEL 7.2.  I'm running on a fresh EC2 instance in both cases, and the steps to reproduce are the same, here is my bash history:

> sudo yum install python-virtualenv
> sudo yum groupinstall 'Development Tools'
> virtualenv flask
> cd flask/
> git clone https://github.com/pallets/flask.git
> cd flask/
> source ../bin/activate
> pip install -e .
> pip install -r test-requirements.txt 
> py.test tests/test_instance_config.py::test_main_module_paths

failing output:

> modules_tmpdir = local('/tmp/pytest-of-ec2-user/pytest-0/test_main_module_paths0/modules_tmpdir'), purge_module = <function inner at 0x170c1b8>
> 
>     def test_main_module_paths(modules_tmpdir, purge_module):
>         app = modules_tmpdir.join('main_app.py')
>         app.write('import flask\n\napp = flask.Flask("__main__")')
>         purge_module('main_app')
>     
>         from main_app import app
>         here = os.path.abspath(os.getcwd())
> >       assert app.instance_path == os.path.join(here, 'instance')
> E       assert '/home/ec2-us...r/py-instance' == '/home/ec2-use...lask/instance'
> E         - /home/ec2-user/flask/var/py-instance
> E         ?                      ^ ^ ---
> E         + /home/ec2-user/flask/flask/instance
> E         ?                      ^^ ^^
> 
> tests/test_instance_config.py:33: AssertionError

This test seems to be passing fine on Travis, and on Debian and OSX.  It appears to be a Redhat/Amazon Linux specific issue.  I ran `git bisect` back to commits on May 26th, 2016 and the test still fails all attempts.

untitaker:
I'm sorry, but what am I supposed to do?

untitaker:
Are you sure that you installed Flask on both machines the same way? What are the respective setuptools, pip and Python versions?

jeffwidman:
[...]

PR #2035, opened 2016-09-21
Modified test_instance_config.py (test_main_module_paths) to reflect flask behavior

drewja:
fix for issue: #1879 
The test of checking default location of the instance folder when running flask app as __main__ fails from within VirtualEnv directory because the working directory shares a prefix with sys.prefix. That is partially how flask decides which kind of instance folder to return:

    "package_path"/var/"name"-instance
    or:
    "working_dir"/instance.

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2035)

### DIFF
--- a/tests/test_instance_config.py
+++ b/tests/test_instance_config.py
@@ -22,22 +22,30 @@ def test_explicit_instance_paths(modules_tmpdir):
     app = flask.Flask(__name__, instance_path=str(modules_tmpdir))
     assert app.instance_path == str(modules_tmpdir)
 
-
 def test_main_module_paths(modules_tmpdir, purge_module):
     app = modules_tmpdir.join('main_app.py')
-    app.write('import flask\n\napp = flask.Flask("__main__")')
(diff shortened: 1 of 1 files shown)


davidism:
Could you provide a project layout that produces the wrong result right now?  In the past people have had problems because they put their project in their `env` folder (wrong), rather than put the `env` in their project (right).

davidism:
Thanks for taking a look at this, but I don't think it's the right solution.  This makes the test dependent on how the dev environment is set up.  Putting the project folder in the env folder is incorrect, the env should either be completely separate or inside the project folder.  It would be useful to test the instance path for an installed package, but this test is just covering up incorrect directory structure.

drewja:
I agree actually, I usually don't use virtualenv as a sort of container for projects. But seeing this open issue I thought I would see what can be done. #1879 should be closed It is only re-producible by cloning the project into a venv directory.

PR #2040, opened 2016-09-26
fix unclosed HTML tags

ethan-funny:
fix unclosed HTML tags

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2040)


ThiefMaster:
Certain tags do not need to be closed, and IIRC. @mitsuhiko prefers the "standalone" version of e.g. `<p>` tags.

Issue #1467, opened 2015-05-21
Add support for default environ in test_client


haizaar:
It would be nice if `test_client()` would accept default environ dictionary. That would stop us doing hacks like [this](http://stackoverflow.com/questions/15278285/setting-mocking-request-headers-for-flask-app-unit-test) over and over.

For example:
```
test_client = app.test_client(defailt_environ = {
    "REMOTE_ADDR": "127.0.0.1"
})
```

`default_environ` should be then passed down all the way to `environment_base` args of `EnvironBuilder` in test.py.

Thank you!

mrecachinas:
@haizaar see #2047 to ensure this met your request, and if so, feel free to close it! :)

haizaar:
Looking at the code seems like it does. Unfortunately I don't have an access to the project where it was originally required. I'll close this issue.

Thank you for implementing it!

PR #2047, opened 2016-10-01
Default environ

mrecachinas:
Addresses #1467. Now you can pass `environ_base` dictionary into `app.test_client`. This also allows overriding per `open` call (or `make_test_environ_builder` call).

Perhaps a better solution would be to edit werkzeug directly and add it to that init (which, in turn, would allow us to remove the subclass init). I don't believe that would cause any unintended behavior, but I'm open to further conversation on it.

---
This change is [Reviewable](https://reviewable.io/reviews/pallets/flask/2047)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -19,8 +19,8 @@ Version 0.12
   well as error handlers.
 - Disable logger propagation by default for the app logger.
 - Add support for range requests in ``send_file``.
-- Default environment can be set in ``app.test_client`` instead of per
-  ``app.test_client.get`` request
+- ``app.test_client`` includes preset default environment, which can now be
+  directly set, instead of per ``client.get``.
 
 Version 0.11.2
 --------------
--- a/CHANGES
+++ b/CHANGES
@@ -19,6 +19,8 @@ Version 0.12
   well as error handlers.
 - Disable logger propagation by default for the app logger.
 - Add support for range requests in ``send_file``.
+- Default environment can be set in ``app.test_client`` instead of per
+  ``app.test_client.get`` request
 
 Version 0.11.2
 --------------
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -114,9 +114,7 @@ class FlaskClient(Client):
         as_tuple = kwargs.pop('as_tuple', False)
         buffered = kwargs.pop('buffered', False)
         follow_redirects = kwargs.pop('follow_redirects', False)
-        builder = make_test_environ_builder(self.application,
-                                            environ_base=environ_base,
-                                            *args, **kwargs)
+        builder = make_test_environ_builder(self.application, *args, **kwargs)
 
         return Client.open(self, builder,
                            as_tuple=as_tuple,
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -48,6 +48,11 @@ class FlaskClient(Client):
 
     preserve_context = False
 
+    def __init__(self, *args, **kwargs):
+        environ_base = kwargs.pop("environ_base", None)
+        super(FlaskClient, self).__init__(*args, **kwargs)
+        self.environ_base = environ_base
+
     @contextmanager
     def session_transaction(self, *args, **kwargs):
         """When used in combination with a ``with`` statement this opens a
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -110,7 +110,10 @@ class FlaskClient(Client):
         as_tuple = kwargs.pop('as_tuple', False)
         buffered = kwargs.pop('buffered', False)
         follow_redirects = kwargs.pop('follow_redirects', False)
-        builder = make_test_environ_builder(self.application, *args, **kwargs)
+        environ_base = kwargs.get(environ_base, None) or self.environ_base
+        builder = make_test_environ_builder(self.application,
+                                            environ_base=environ_base,
+                                            *args, **kwargs)
 
         return Client.open(self, builder,
                            as_tuple=as_tuple,
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -110,7 +110,7 @@ class FlaskClient(Client):
         as_tuple = kwargs.pop('as_tuple', False)
         buffered = kwargs.pop('buffered', False)
         follow_redirects = kwargs.pop('follow_redirects', False)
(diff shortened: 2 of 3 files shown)


untitaker:
I'd be happy to set some default WSGI env vars such as REMOTE_ADDR, HTTP_USER_AGENT, ...

mrecachinas:
@untitaker Ooh, good point. Where do you think that would be best suited? Werkzeug?

untitaker:
Not sure, but I guess Flask would be the better place for sensible but presumptive defaults.

mrecachinas:
@untitaker I've started putting something together in `FlaskClient.__init__`. Is there any helper function already in werkzeug or flask (or even wsgi) that lists all the available WSGI env vars?

Otherwise, where should I look to get this list, since it seems fairly exhaustive of HTTP headers among others?

flask/testing.py
```diff
@@ -105,7 +110,10 @@ def open(self, *args, **kwargs):
         as_tuple = kwargs.pop('as_tuple', False)
         buffered = kwargs.pop('buffered', False)
         follow_redirects = kwargs.pop('follow_redirects', False)
-        builder = make_test_environ_builder(self.application, *args, **kwargs)
+        environ_base = kwargs.get("environ_base", None) or self.environ_base
```
untitaker: You can just use `kwargs.setdefault` here, like at the top of `open`.

flask/testing.py
```diff
@@ -48,6 +48,11 @@ class FlaskClient(Client):
 
     preserve_context = False
 
+    def __init__(self, *args, **kwargs):
+        environ_base = kwargs.pop("environ_base", None)
+        super(FlaskClient, self).__init__(*args, **kwargs)
+        self.environ_base = environ_base
```
untitaker: I wouldn't necessarily add a keyword argument here, but only do something like:

    self.environ_base = {
        "REMOTE_ADDR": "127.0.0.1",
        "HTTP_USER_AGENT": "werkzeug/" + werkzeug.__version__
    }

untitaker:
There is no such list, because you can put any kind of data in there. We don't need to set a lot either...

flask/testing.py
mrecachinas: Are we switching from passing it in `app.test_client()` to env vars? I thought you meant in addition.

untitaker: I mean, currently to add your own `environ_base` you would do this:

    c = TestClient(environ_base={...})

But I only want to do this, because it's not that much of a common usecase:

    c = TestClient()
    c.environ_base = ...

untitaker: In either case we would have to add our default variables here.

mrecachinas: Gotcha!

flask/testing.py
mrecachinas: See commit #2ecdec5

flask/testing.py
mrecachinas: See commit 695701e below

mrecachinas:
Ah ok. Makes sense.

flask/testing.py
```diff
@@ -101,11 +109,14 @@ def session_transaction(self, *args, **kwargs):
...
         kwargs.setdefault('environ_overrides', {}) \
             ['flask._preserve_context'] = self.preserve_context
+        kwargs.setdefault('environ_base', self.environ_base)
 
         as_tuple = kwargs.pop('as_tuple', False)
         buffered = kwargs.pop('buffered', False)
         follow_redirects = kwargs.pop('follow_redirects', False)
-        builder = make_test_environ_builder(self.application, *args, **kwargs)
+        builder = make_test_environ_builder(self.application,
+                                            environ_base=environ_base,
```
untitaker: This line isn't needed then.

mrecachinas: Yup, just removed it.

mrecachinas:
@untitaker Add the changes to `flask/CHANGES`?

untitaker:
Yes please

mrecachinas:
@untitaker -- let me know if you'd like me to do anything else!

flask/testing.py
```diff
@@ -43,11 +44,34 @@ class FlaskClient(Client):
     information about how to use this class refer to
     :class:`werkzeug.test.Client`.
 
+    Uses default environment values
+
+    ::
+    
+        {
```
untitaker: Perhaps we should only say that we set some default env values, and that the exact values can be found at `client.environ_base`. Having them duplicated here may lead to one of the copies being out of date.

mrecachinas: No prob! Just added the change.

untitaker:
One last last change! :cold_sweat: Could you add tests?

jeffwidman:
@mrecachinas thanks for doing this, I'm looking forward to using it.

mrecachinas:
No problem @untitaker -- I had some that I had been testing with that I just committed. Let me know what you think and if there is anything else I can do.

mrecachinas:
@untitaker @jeffwidman what do you think about also allowing the user to set `environ_base` upon app.test_client() instantiation?

untitaker:
I don't expect that users will have to set `environ_base` that often.

CHANGES
```diff
@@ -19,6 +19,8 @@ Version 0.12
   well as error handlers.
 - Disable logger propagation by default for the app logger.
 - Add support for range requests in ``send_file``.
+- Default environment can be set in ``app.test_client`` instead of per
+  ``app.test_client.get`` request
```
untitaker: This should also mention that defaults are already set.

flask/testing.py
```diff
@@ -43,11 +44,21 @@ class FlaskClient(Client):
     information about how to use this class refer to
     :class:`werkzeug.test.Client`.
 
+    Default environment values can be set after instantiation of the
+    `app.test_client()` object in `client.environ_base`.
+
```
untitaker: Could you add a `versionchanged` directive?

CHANGES
mrecachinas: See 29371b8.

mrecachinas:
@untitaker Fair enough. Just wanted to double check.

untitaker:
Thanks!

Issue #2020, opened 2016-09-10
Default flask.cli commands are not override-able


allanlei:
The default commands `run` and `shell` cannot be overridden.

Using a custom command called `shell`, I expect the custom `shell` to override the builtin `shell`
```python
@app.cli.command('shell', short_help='My custom shell')
def shell():
    pass
```

But running `--help` still produces `flask.cli.shell`'s help
```bash
> python -m flask
Usage: python -m flask [OPTIONS] COMMAND [ARGS]...

  This shell command acts as general utility script for Flask applications.

  It loads the application configured (either through the FLASK_APP
  environment variable) and then provides commands either provided by the
  application or Flask itself.

  The most useful commands are the "run" and "shell" command.

  Example usage:

    $ export FLASK_APP=hello
    $ export FLASK_DEBUG=1
    $ flask run

Options:
  --help  Show this message and exit.

Commands:
  run    Runs a development server.
  shell  Runs a shell in the app context.
```

And running gives me `flask.cli.shell`'s shell

```bash
> python -m flask shell

Python 3.5.2 (default, Jul 17 2016, 00:00:00) 
[GCC 4.8.4] on linux
App: wallet
Instance: app/instance
>>> 
```

Also,

```python
@app.cli.command('shell_plus', short_help='bar')
def shell_plus_1():
    pass

@app.cli.command('shell_plus', short_help='foo')
def shell_plus_2():
    pass
```
has the behaviour I expect

```bash
Usage: python -m flask [OPTIONS] COMMAND [ARGS]...

  This shell command acts as general utility script for Flask applications.

  It loads the application configured (either through the FLASK_APP
  environment variable) and then provides commands either provided by the
  application or Flask itself.

  The most useful commands are the "run" and "shell" command.

  Example usage:

    $ export FLASK_APP=hello
    $ export FLASK_DEBUG=1
    $ flask run

Options:
  --help  Show this message and exit.

Commands:
  run         Runs a development server.
  shell       Runs a shell in the app context.
  shell_plus  foo
```

mrecachinas:
This appears to be the intended behavior to ensure basic script functionality in the event the application fails. See https://github.com/pallets/flask/blob/master/flask/cli.py#L329-L349. 

If there is internal interest in changing that, I have a patch ready for review.

davidism:
~~~python
from flask.cli import FlaskGroup, run_command

cli = FlaskGroup(add_default_commands=False, create_app=create_app)
cli.add_command(run_command)

[...]

PR #2053, opened 2016-10-09
Add possible overriding of builtin cli cmds

mrecachinas:
This PR is in reference to #2020 (cc @davidism). Basically, a CLI command was checked if it was a builtin, and if so, it executed it. I've changed this to now wait until it sees if the same command exists in the user-created commands.

Also, the test I added works both with and without the changes, so definitely let me know what I'm missing.


mrecachinas:
Ping @davidism -- what are your thoughts on this?

kblomqvist:
Would this also allow to call the builtin command from the overridden command?

davidism:
Would you need to? I assume if you're overriding it you can make it do whatever you want, including adding a flag that calls the original instead.

kblomqvist:
As long as I don't know how to implement the builtin run command, I would need to be able to call it.

davidism:
Oh, I see what you mean. The commands should still be importable and callable, but not sure if there's a specific way it's supposed to be done with Click.

kblomqvist:
Not sure either, but maybe... http://click.pocoo.org/5/advanced/#invoking-other-commands.

davidism:
@kblomqvist that looks right.

@mrecachinas the test needs to work with `@app.cli.command`. Creating a separate `FlaskGroup` is the existing solution.

davidism:
I've been working on getting this into shape to merge. Overriding the built-in commands in a useful way involves copy-pasting the whole command.

I assume you would override `run` to pass extra arguments to `run_simple`. I can add `**kwargs` to the signature, but you still don't have access to any of the processing that's done to the built-in's options. And you have to remember to copy the built-in's help and options.

Overriding `routes` doesn't really make sense, or I don't know how to refactor it for that to be useful.

Overriding `shell` makes sense, for example to use IPython if it's installed or just forward to the built-in command otherwise, but you need to remember to pop any extra params from the context first, since the built-in takes no arguments.

I guess this is still more convenient than setting up a separate `FlaskGroup`, but it's a little disappointing that it's not more powerful.

---

Here's a custom `run`, assuming the built-in passes `**kwargs` to `run_simple`, for example:

~~~python
@app.cli.command('run', help=builtin_run_command.help, short_help=builtin_run_command.short_help)
@click.option('--ssl', 'ssl_context', flag_value='adhoc')
@click.pass_context
def run_command(ctx, ssl_context=None, **kwargs):
    click.echo(' * SSL is {}'.format(ssl_context or 'disabled')
    ctx.forward(builtin_run_command)

run_command.params[:0] = builtin_run_command.params

davidism:
Actually, shell absolutely requires copying the  whole thing if you want the same behavior.

flask/cli.py
```diff
@@ -341,43 +341,40 @@ def _load_plugin_commands(self):
 
     def get_command(self, ctx, name):
         self._load_plugin_commands()
-
-        # We load built-in commands first as these should always be the
```
untitaker: Is there discussion about this statement? It appears Armin already had that usecase in mind and intentionally wants to discourage overriding builtin commands.

davidism: Custom commands through entry points can already override the built-ins. https://github.com/ei-grad/flask-shell-ipython/blob/3f8a82013a6fd8ef42156470e30e3638604190a1/setup.py#L17

kennethreitz:
This overall seems like a bad idea.

davidism:
Yeah, I'm going to close this. Thanks for the work though, @mrecachinas, we needed to at least see what it would look like.

If someone wants to override a builtin they can use the cli entry points, which already work fine, but I'd rather that stay a semi-undocumented feature. @doobeh is working on a Flask-Gunicorn extension which demonstrates the technique anyway.

It would be nice to show a message "can't override built-in command with `@app.cli.command`" so users aren't confused when their command isn't recognized.

mrecachinas:
@davidism no worries! I agree with your conclusion.

PR #2055, opened 2016-10-11
Updated Celery pattern

geoffreybauduin:
The given pattern caused Celery to lose the current Context.

```
> ~/celery/app/task.py(424)__call__()
-> try:
(Pdb) p self.request
<Context: {}>
```

And with this new pattern:

```
(Pdb) p self.request
<Context: {'called_directly': False, 'retries': 5, 'is_eager': True, 'loglevel': 0, 'delivery_info': {'is_eager': True}, 'args': (), 'callbacks': None, 'errbacks': None, 'kwargs': {}, 'logfile': None, 'id': '9fbd9fa0-1a42-4e70-9c52-558ad3bc9d72', 'headers': None}>
```

Note that I'm using Celery 3.1.24. The current pattern may be outdated (last change was 2013)

### DIFF
--- a/docs/patterns/celery.rst
+++ b/docs/patterns/celery.rst
@@ -44,7 +44,7 @@ This is all that is necessary to properly integrate Celery with Flask::
             abstract = True
             def __call__(self, *args, **kwargs):
                 with app.app_context():
-                    return TaskBase.__call__(self, *args, **kwargs)
+                    return self.run(*args, **kwargs)
         celery.Task = ContextTask
         return celery
 
--- a/docs/patterns/celery.rst
+++ b/docs/patterns/celery.rst
@@ -1,24 +1,27 @@
-Celery Based Background Tasks
-=============================
+Celery Background Tasks
+=======================
 
-Celery is a task queue for Python with batteries included.  It used to
-have a Flask integration but it became unnecessary after some
-restructuring of the internals of Celery with Version 3.  This guide fills
-in the blanks in how to properly use Celery with Flask but assumes that
-you generally already read the `First Steps with Celery
-<http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html>`_
-guide in the official Celery documentation.
+If your application has a long running task, such as processing some uploaded
+data or sending email, you don't want to wait for it to finish during a
+request. Instead, use a task queue to send the necessary data to another
+process that will run the task in the background while the request returns
+immediately.
 
-Installing Celery
------------------
+Celery is a powerful task queue that can be used for simple background tasks
+as well as complex multi-stage programs and schedules. This guide will show you
+how to configure Celery using Flask, but assumes you've already read the
+`First Steps with Celery <http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html>`_
+guide in the Celery documentation.
 
-Celery is on the Python Package Index (PyPI), so it can be installed with
-standard Python tools like :command:`pip` or :command:`easy_install`::
+Install
+-------
+
+Celery is a separate Python package. Install it from PyPI using pip::
 
     $ pip install celery
 
-Configuring Celery
-------------------
+Configure
+---------
 
 The first thing you need is a Celery instance, this is called the celery
 application.  It serves the same purpose as the :class:`~flask.Flask`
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
@lght at first glance this seems highly problematic, as this removes the logic in `__call__` that has to happen around self.run in order to handle bound tasks, context stack interaction and more

lght:
@RonnyPfannschmidt it would seem problematic if it wasn't mentionned in their documentation that doing so would remove the whole request context by creating a new generic one (here, `<Context: {}>`).

In their doc (https://github.com/celery/celery/blob/v3.1.24/celery/app/task.py#L212), they specify the task is, by default, run with the `run` method, unless `__call__` is overriden, but calling directly `__call__` completely masks the context (this function,https://github.com/celery/celery/blob/v3.1.24/celery/app/task.py#L423, creates and pushes a new context, before calling `run` itself). (`run` being the function decorated by the task decorator).

One of the issues faced by such action (calling `Task.__call__`) is that the task won't be retried. Indeed, the default context sets the `request.called_directly` to `True`, and Celery is not willing to retry the task (https://github.com/celery/celery/blob/v3.1.24/celery/app/task.py#L667)

I'm not sure about the intended use of `__call__`, except for calling directly a task.

Most of the answers found on Stackoverflow about such problems, or about executing some lines of code before/after the execution of the task, are suggesting doing an override of `__call__` and then call the `run` method. I'm not saying it's a good practice (and that they are right), just that it's worth mentioning this pattern should be updated.

Try it yourself, you'll see.

untitaker:
Would really like to know how you arrive at the behavior you encountered in the
OP. Are you calling the task directly such that you expect the existing ctx to
be passed down instead of a new one being pushed?

With all we know you might as well file this PR against Celery itself to remove
everything but the invocation of `run` from `__call__`.

lght:
@untitaker I would not go for patching this kind of behavior. It must have an intended use.

I'm calling the task like it is mentionned in celery's doc (`task.apply_async()`).

I'll provide a link to a quick script that shows exactly the disappearence of the task context before the end of the day

ThiefMaster:
Maybe a celery dev could chime in here?

cc @ask

lght:
Speeding things up, you'll find a quick script to see what's actually going on here:

https://gist.github.com/lght/f5fa6e4924739897d9505f14edc5a190

Console output:

```
(venv) > python run.py
 * Restarting with stat
 * Debugger is active!
 * Debugger pin code: 107-501-531
Current context: <Context: {}>
127.0.0.1 - - [12/Oct/2016 19:28:30] "GET /task HTTP/1.1" 200 -
Current context: <Context: {'called_directly': False, 'delivery_info': {'is_eager': True}, 'id': '8d6c3c25-e2c8-4b2e-a39e-736dbf4f973a', 'loglevel': 0, 'kwargs': {}, 'is_eager': True, 'headers': None, 'callbacks': None, 'errbacks': None, 'retries': 0, 'logfile': None, 'args': ()}>
127.0.0.1 - - [12/Oct/2016 19:28:34] "GET /new_task HTTP/1.1" 200 -
^C%  
```

First context is the one with `__call__` pattern, second one is with `run` pattern.

davidism:
I can't reproduce this issue, and the example is gone. The documentation worked for me as-is. It would be nice to make the pattern in the docs more robust

geoffreybauduin:
@davidism sorry, the example is still there, the url just moved, https://gist.github.com/geoffreybauduin/f5fa6e4924739897d9505f14edc5a190

untitaker:
I'd rather tend to close this though just because it appears that nobody else has this problem.

davidism:
I've been investigating this, I really can't tell what the intention of the design was for Celery here. If `__call__` is not overridden, it is skipped and `run` is called directly. But `__call__` has custom logic in it already that seems like it shouldn't be skipped (except for pushing that empty request).

Additionally, it only seems to happen with `task_always_eager = True`, calling it through a worker shows the real request. I haven't dug into remotely stepping through that yet.

So as far as I can tell, this is a correct fix. If you want to override `__call__`, you should not call `super`. But it feels really weird.

davidism:
There's a function, [`_install_stack_protection`](https://github.com/celery/celery/blob/v4.0.2/celery/app/trace.py#L599) that replaces `BaseTask.__call__` with a version that doesn't insert an empty request, to guard against this exact situation, but it's only called when [setting up a `Worker` instance](https://github.com/celery/celery/blob/v4.0.2/celery/apps/worker.py#L97). `task_always_eager` doesn't trigger that because it doesn't go through a worker.

davidism:
Got around to finishing testing this. With the change, things at least behave consistently between `always_eager` enabled or not.

* `my_task()` (task called as function) has no context, which I guess makes sense.
* `my_task.apply()` (task executed eagerly) has a context.
* `my_task.delay()` (task executed remotely) has a context.

It looks like calling `run` instead of `super.__call__` is the intended way to write a task subclass.

woshihaoren:
How to call factory function?   I can't init_app  celery on flask..

davidism:
I'm having trouble understanding your question. If you have a factory, one pattern to create a celery instance is to create a `worker.py` file with the following and point Celery at that instead.

~~~python
app = create_app()
celery = create_celery(app)
~~~

If you have questions about your own code, please use Stack Overflow. If you would like to open an issue about Flask's docs on this subject, please open a *new* issue and be sure to follow the displayed guidelines.

woshihaoren:
OK , I will open a new issue

woshihaoren:
@davidism I use flask-celery-helper solve it

Issue #2058, opened 2016-10-15
The file/path provided (flaskr) does not appear to exist. 


tuhaolam:
The issue tracker is a tool to address bugs.
Please use the #pocoo IRC channel on freenode or Stack Overflow for questions.

~/flask-master/examples/flaskr$ flask run
Usage: flask run [OPTIONS]

Error: The file/path provided (flaskr) does not appear to exist.  Please verify the path is correct.  If app is not on PYTHONPATH, ensure the extension is .py

untitaker:
cd .., if you have an issue with the docs please also quote the sentence that is false.

wgwz:
have you done the export FLASK_APP step?

StefanoGitHub:
I have a similar issues too.
I followed the readme instructions and I get the following error:
```
(venv) ------------------------
@flaskr:$ pip install --editable .
  ...
Successfully installed flaskr
(venv) ------------------------
@flaskr:$ export FLASK_APP=flaskr
(venv) ------------------------
@flaskr:$ flask initdb
Traceback (most recent call last):
  File "/Users/Stefano/GoogleDrive/Python/Flask_examples/flaskr/venv/bin/flask", line 11, in <module>
    sys.exit(main())
  File "/Users/Stefano/GoogleDrive/Python/Flask_examples/flaskr/venv/lib/python2.7/site-packages/flask/cli.py", line 478, in main
    cli.main(args=args, prog_name=name)
  File "/Users/Stefano/GoogleDrive/Python/Flask_examples/flaskr/venv/lib/python2.7/site-packages/flask/cli.py", line 345, in main
    return AppGroup.main(self, *args, **kwargs)
  File "/Users/Stefano/GoogleDrive/Python/Flask_examples/flaskr/venv/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/Stefano/GoogleDrive/Python/Flask_examples/flaskr/venv/lib/python2.7/site-packages/click/core.py", line 1055, in invoke
    cmd_name, cmd, args = self.resolve_command(ctx, args)
  File "/Users/Stefano/GoogleDrive/Python/Flask_examples/flaskr/venv/lib/python2.7/site-packages/click/core.py", line 1094, in resolve_command
    cmd = self.get_command(ctx, cmd_name)
[...]

PR #2068, opened 2016-10-30
Fixes import statement in flaskr

wgwz:
- `from flaskr.flaskr import app` in flaskr/__init__.py
  causes an import error with Python 2
- The relative import now used works for py2 and py3

Ref: https://github.com/pallets/flask/issues/2058 

The issue was that the absolute import `from flaskr.flaskr import app` works in python 3 but not python 2. While there is probably a simple answer why, for now, this relative import works on both.

### DIFF
--- a/docs/tutorial/packaging.rst
+++ b/docs/tutorial/packaging.rst
@@ -55,7 +55,7 @@ into this file, :file:`flaskr/__init__.py`:
 
 .. sourcecode:: python
 
-    from flaskr import app
+    from .flaskr import app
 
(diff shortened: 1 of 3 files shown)


wgwz:
Note: https://github.com/pallets/flask/pull/2052 This was the last fix, `from flaskr.flaskr import app`. At least I thought that was the fix. That was only merged into master, not into 0.11-maintenance as I had been doing.

untitaker:
Thanks!

PR #2071, opened 2016-10-31
Adding coverage generation to tox

alexk307:
Adds the ability to measure code coverage via unit tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,10 +5,10 @@ envlist = {py26,py27,pypy}-{lowest,release,devel}{,-simplejson}, {py33,py34,py35
 
 [testenv]
 commands =
-    py.test []
-
+    py.test --cov=flask --cov-report html {toxinidir}/tests/
 deps=
     pytest
+    pytest-cov
     greenlet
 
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -5,10 +5,10 @@ envlist = {py26,py27,pypy}-{lowest,release,devel}{,-simplejson}, {py33,py34,py35
 
 [testenv]
 commands =
-    py.test []
-
+    py.test --cov=flask --cov-report html {toxinidir}/tests/
```
untitaker: Do we need the last parameter?

alexk307: Doesn't look like we do, I'll get rid of it.

untitaker: Also the `[]` that you removed allows us to pass parameters `a b c` to py.test using `tox -epy27 -- a b c`.

untitaker:
Could you add the `[]` as well? Thanks!

untitaker:
Thanks!

alexk307:
Thanks for merging! Let me know if you want help with the other part of the coverage stuff

PR #2072, opened 2016-11-01
Remove busy-work.

mjpieters:
It is entirely sufficient to walk the MRO of the exception class, no need to check for classes re-appearing later on, no need to add the MRO of any superclass.

* Python refuses point-blank to create a class with a circular MRO.
* All classes in a superclass MRO *already* appear in the MRO of the derived type. Re-adding the contents of a superclass MRO is doing double work.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -14,7 +14,6 @@ from threading import Lock
 from datetime import timedelta
 from itertools import chain
 from functools import update_wrapper
-from collections import deque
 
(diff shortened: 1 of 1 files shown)


davidism:
This reverts 7d506f2408520ecb21025f93d9ece00431004501.

untitaker:
If you want to traverse the superclass of a superclass...

PR #2073, opened 2016-11-01
Parameterizing test

alexk307:


tests/test_helpers.py
```diff
@@ -22,6 +22,9 @@
 from werkzeug.http import parse_cache_control_header, parse_options_header
 from werkzeug.http import http_date
 from flask._compat import StringIO, text_type
+from hypothesis import given
```
davidism: Use `import hypothesis`, `@hypothesis.given`, otherwise this is going to get even harder to keep track of.  Same with below: `import hypothesis.strategies`, `hypothesis.strategies.uuids()`, etc.

davidism:
How does this affect testing time?  Why add it to these methods (did you notice something in particular)?

alexk307:
@davidism 
The testing time should remain relatively the same, as all of these tests are pretty small and have no i/o.

When I spoke with @untitaker yesterday/today they mentioned they were interested in trying it out.

alexk307:
yeah, as I suspected, hypothesis is not capable with python 2.6.

tox.ini
```diff
@@ -11,6 +11,9 @@ deps=
     pytest
     pytest-cov
     greenlet
+    pytest-cov
```
untitaker: This is already in the dependencies.

untitaker:
Oh and Hypothesis tests a few thousand examples, which is extremely slow and probably should only be used in places where it's likely to find bugs, e.g. encoding and decoding routines

tox.ini
alexk307: ah sorry about that.

alexk307:
@untitaker we can limit the number of examples that it runs. I think by default it does ~200. I'll take a look at Werkzeug's test suite as well. I can add back in the tests for 2.6 and skip these on 2.6

alexk307:
I'm going to remove hypothesis for now until I can get it to work on 2.6/can find good enough use cases to put it back in. For now I just parameterized that test.

tests/test_helpers.py
```diff
@@ -113,20 +113,17 @@ def test_json_dump_to_file(self):
...
         app = flask.Flask(__name__)
         c = app.test_client()
-        for i, d in enumerate(test_data):
-            url = '/jsonify_basic_types{0}'.format(i)
-            app.add_url_rule(url, str(i), lambda x=d: flask.jsonify(x))
-            rv = c.get(url)
-            assert rv.mimetype == 'application/json'
-            assert flask.json.loads(rv.data) == d
+
+        url = '/jsonify_basic_types{0}'.format(0)
```
alexk307: good point. Removed.

alexk307:
@untitaker bump

PR #2082, opened 2016-11-09
Error handler support for generic HttpException

gcetusic:
Flask supports setting HttpException as error handler class:

    app.register_error_handler(HTTPException, lambda error: jsonify(error=error.code, text=str(error)))

This fills `error_handler_spec` with something like this for any blueprint `{None: {None: lambda_function}}`

However, when finding the error handler it checks the error code which is e.g. 404 and ignores None, the error handler for any error, previously set by register_error_handler. It doesn't find 404, doesn't check for None in the dict and falls back to the default error handler which shouldn't happen.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1454,7 +1454,7 @@ class Flask(_PackageBoundObject):
             return handler
 
         # fall back to app handlers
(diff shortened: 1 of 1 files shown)


jeffwidman:
Closing in favor of #2304

PR #2091, opened 2016-11-17
small refactoring of  is_json function.

alexpantyukhin:


ThiefMaster:
I don't see any benefit in this. It just reduces readability.

Especially using `foo or bar and foobar` without parentheses. Yes, the order is well-defined, but no, it's not consistent across languages, so it just makes it harder to read the logic since you don't *immediately* see it.

alexpantyukhin:
got it.

PR #2094, opened 2016-11-21
added whitespace before key

VishvajitP:
added whitespace before key


ThiefMaster:
`foo.py:123` is much more common than `foo.py: 123`

untitaker:
@VishvajitP This isn't a key of a dict, `foo.py:23` is a common way to specify line and file.

PR #2095, opened 2016-11-21
added blank line before def

VishvajitP:
added blank line before def


ThiefMaster:
I'm quite sure the lack of empty lines there is intentional to group all the stuff from each property together.

ThiefMaster:
Also, please use separate branches for pull requests and use rebase instead of merge when syncing with upstream. If we merged this commit without squashing we would add quite some noise to the repo's commit history.

PR #2097, opened 2016-11-22
Fix typo: star => start

kumabotz:


davidism:
Star import as in `from x import *`.

kumabotz:
Got it. 👍

PR #2103, opened 2016-12-03
fix unclosed p tag

greyli:


ThiefMaster:
Unclosed p tags are valid HTML 5

PR #2104, opened 2016-12-03
Happy new year 2016

elahmo:
Quickly, before 2017 comes...


davidism:
We are no longer changing copyright years each year. Copyright is established at the first moment a work is created, it does not need to be incremented.

davidism:
If you would like to go the other direction and restore the original dates for all files with copyright statements, that would be appreciated, but it requires a bit of Git forensics.

ThiefMaster:
Sounds a bit like overkill tbh. I'd rather just get rid of the dates altogether...

davidism:
I have no strong opinion on it besides "don't need to update it."

PR #2105, opened 2016-12-04
Warn when cookie domain is set to an IP

JJPO96:
Possible solution for issue #2007, it gives a warning message when an IP is set as SESSION_COOKIE_DOMAIN as user untitaker asked us to do, SERVER_DOMAIN isn't checked because he didn't remembered why it was needed, but the function is_IP in /flask/helpers is prepared to it. If you have any doubt feel free to ask.


flask/helpers.py
```diff
@@ -958,3 +958,33 @@ def total_seconds(td):
     :rtype: int
     """
     return td.days * 60 * 60 * 24 + td.seconds
+
+def is_IP(string, var_name):
```
ThiefMaster: PEP8: should be `is_ip`

Also, the naming is very different from what the function does.

flask/sessions.py
```diff
@@ -332,6 +332,7 @@ def open_session(self, app, request):
 
     def save_session(self, app, session, response):
         domain = self.get_cookie_domain(app)
+        is_IP(domain, "SESSION_COOKIE_DOMAIN")
```
ThiefMaster: ideally you would use `if is_ip(domain): warnings.warn(...)` instead of having the function print something

JJPO96:
What name would you suggest? Regards

ThiefMaster:
I think instead of string operations the proper way to check whether it's an IP is using [`socket.inet_pton()`](https://docs.python.org/2/library/socket.html#socket.inet_pton) and returning `True` if it doesn't raise, otherwise `False`

JJPO96:
@ThiefMaster What should be the category of the warning message?

ThiefMaster:
maybe RuntimeWarning... I'd check if Flask uses any other warnings (that are not deprecation-related) and possibly use the same type

Also, tests are failing since you don't handle the case where the domain is unset, i.e. `None`

JJPO96:
@ThiefMaster  How can I get the socket of the session when I call the function is_ip() in the save_session() function?

ThiefMaster:
It's a function from python's `socket` module, so you simple `import socket` at the top of the file...

flask/sessions.py
```diff
@@ -332,7 +332,8 @@ def open_session(self, app, request):
 
     def save_session(self, app, session, response):
         domain = self.get_cookie_domain(app)
-        is_IP(domain, "SESSION_COOKIE_DOMAIN")
+        if domain != None:
```
ThiefMaster: `if domain is not None:`

flask/sessions.py
```diff
@@ -332,6 +332,8 @@ def open_session(self, app, request):
 
     def save_session(self, app, session, response):
         domain = self.get_cookie_domain(app)
+        if domain != None:
+            is_ip(domain, "SESSION_COOKIE_DOMAIN", self)
```
ThiefMaster: Why are you passing `self` here? This doesn't even work. Please test your code (and flask's unittests) before committing/pushing...

JJPO96:
Sorry for that, in my last commit all the unit tests are passing. I also did the changes you asked for.

flask/sessions.py
```diff
@@ -332,6 +333,9 @@ def open_session(self, app, request):
 
     def save_session(self, app, session, response):
         domain = self.get_cookie_domain(app)
+        if domain is not None:
```
untitaker: can be conjugated with `and`

flask/sessions.py
```diff
@@ -332,6 +333,9 @@ def open_session(self, app, request):
...
+            if is_ip(domain):
+                warnings.warn("IP introduced in SESSION_COOKIE_DOMAIN", RuntimeWarning)
```
untitaker: IMO this should elaborate on why this is a problem

davidism:
@untitaker can you make a decision on this?

untitaker:
I think the code is fine except for my review comments.

davidism:
Setting `SESSION_COOKIE_DOMAIN = '127.0.0.1:5000'` works fine. The issue is when it's detected through `SERVER_NAME` instead, in which case Flask adds a `'.'` before the "domain", which Chrome doesn't like (rightly so, an IP isn't a domain). We can just use the `is_ip` function to decide to remove the `'.'` prefix and things will work as expected.

davidism:
I also don't like the idea of running `is_ip` for essentially every request when ideally `domain` will not be an IP.

davidism:
Continued in #2282.

Issue #1286, opened 2014-12-18
send_file fails when filename contains unicode symbols


georgthegreat:
Hi.

I've detected an issue with supporting unicode filenames in send_file.
If send_file attempts to respond with utf-8 in http headers, the answer is empty, the log contains something like "http-headers should containbe latin-1".
I know that [browser support IS A MESS](http://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http), but it seems, that sending two filenames (`filename=` and `filename*=`) separated by semicolon should work.

I'd like this to be handled by flask or werkzeug. Will you accept such pull request?

keyanp:
Similar discussion has already happened regarding this observation. See the issue from `requests` to read more about the details, but the short of it is that HTTP headers are supposed to only accept `latin-1` characters. Changing this behavior in flask or werkzeug would only cause issues by allowing requests to be sent in an unsupported encoding.

Someone else can chime in if I am mistaken, but I think this can be closed.

Similar discussion:
https://github.com/kennethreitz/requests/issues/1926
https://github.com/jakubroztocil/httpie/issues/212

untitaker:
I suppose the problem here is that it's not clear which encoding send_file takes for the filename. Latin-1 makes sense, but people might assume it's UTF-8.

keyanp:
I updated the API documentation to make that explicit, is that sufficient?

georgthegreat:
BTW, I've solved the problem by sending filename*=utf-8'blah' header field.
Works everywhere except for the IE <= 8.

Here is the snippet:

    response = flask.make_response(flask.send_file(pdf_full_path))
    response.headers["Content-Disposition"] = \
        "attachment; " \
        "filenane={ascii_filename};" \
        "filename*=UTF-8''{utf_filename}".format(
        ascii_filename="book.pdf",
        utf_filename=urlparse.quote(os.path.basename(pdf_full_path))
    )

keyanp:
From the research I did I am surprised that worked for you actually, but either way I'm not sure if we should be suggesting that workaround because the encoding doesn't seem to be universally supported.

georgthegreat:
I'm not really worrying about universal browser support.
Here are the support tables for the feature.
http://greenbytes.de/tech/tc2231/

As far, as I understand, providing both `filename` and `filename*` should be more compatible. A library for transliteration / unaccenting will be required though.

keyanp:
From that link:

[...]

PR #2115, opened 2016-12-18
Feature/ascii

JJPO96:
Implemented the solution purposed by @dsully for issue #1286

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -972,4 +972,4 @@ def is_ascii(string):
     :returns: True if string is fully ASCII encoded, False if not
     :rtype: boolean
     """
-    return all(ord(char) < 128 for char in string):
+    return all(ord(char) < 128 for char in string);
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -504,6 +504,11 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         a :class:`~datetime.datetime` or timestamp.
         If a file was passed, this overrides its mtime.
     """
+    import warnings
+    if type(filename_or_fp) is str:
+        if not is_ascii(filename_or_fp):
+            warnings.warn("Non-ASCII filename is being used in send_file", UnicodeWarning)
+
     mtime = None
     fsize = None
     if isinstance(filename_or_fp, string_types):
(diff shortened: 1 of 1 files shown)


flask/helpers.py
```diff
@@ -504,6 +504,11 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         a :class:`~datetime.datetime` or timestamp.
         If a file was passed, this overrides its mtime.
     """
+    import warnings
```
ThiefMaster: Why not import it at the top level?

flask/helpers.py
```diff
@@ -504,6 +504,11 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
...
+    if type(filename_or_fp) is str:
```
ThiefMaster: This is not the proper way of checking whether something is a string. Use `isinstance(.., string_types)` (assuming you want to handle both unicode and byte strings).

flask/helpers.py
```diff
@@ -958,3 +963,16 @@ def total_seconds(td):
     :rtype: int
     """
     return td.days * 60 * 60 * 24 + td.seconds
+
+def is_ascii(string_to_test):
```
ThiefMaster: I'd just call the argument `string`

flask/helpers.py
```diff
@@ -958,3 +963,16 @@ def total_seconds(td):
...
+    """Returns if a string object is fully ASCII encoded.
+
+    :param string string_to_test: the string to be evaluated
+
+    :returns: True if string_to_test is fully ASCII encoded, False if not
+    :rtype: boolean
+    """
+    if all(ord(char) < 128 for char in string_to_test):
```
ThiefMaster: `return all(ord(char) < 128 for char in string_to_test)`

JJPO96:
@ThiefMaster so is it better to delete this branch and create a new updated one? Or do you have any better suggestion?

ThiefMaster:
No, just rebase/amend and then force-push. That will update this PR without adding even more commits.

JJPO96:
Why did the checks failed?

ThiefMaster:
There's a syntax error. Please test code before committing/pushing it...

JJPO96:
Sorry it was a misunderstood with my colleagues.

davidism:
Closing because #2223 added support for UTF-8 filenames.

PR #2122, opened 2016-12-25
Make flaskr.py command line friendlier

gliptak:
This allows for
```
python flaskr.py initdb
python flaskr.py run
```

### DIFF
--- a/examples/flaskr/flaskr/flaskr.py
+++ b/examples/flaskr/flaskr/flaskr.py
@@ -52,6 +52,11 @@ def initdb_command():
     print('Initialized the database.')
 
 
+@app.cli.command('run')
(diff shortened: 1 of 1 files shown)


untitaker:
You should be able to use `flask run` and `flask initdb` as explained in the (very recently fixed) docs.

gliptak:
Yes, I have read the documentation section. This above allowed me to run "directly" from the source tree. If this is not the right way, I'm happy to close the PR. Thanks

untitaker:
We want to establish `flask` as standard "entry point" for all CLIs, so having anything else in the examples would be detrimental to that agenda. `app.cli` is there for overriding or for testing (I guess).

PR #2123, opened 2016-12-25
Fix config.from_pyfile on Python 3

### LINKED ISSUE #2118: config.from_pyfile crashes on Python 3 when source isn't encoded in default encoding
when I read my instance config file, I get an error. 

> exec(compile(config_file.read(), filename, 'exec'), d.__dict__)
> UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 437: illegal multibyte sequence
Then I modify the code of config.from_pyfile to this

> with open(filename, 'rb') as config_file:
The problem is resolved.

untitaker:
Fix #2118

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -7,11 +7,13 @@
     :license: BSD, see LICENSE for more details.
 """
 
-import pytest
 
-import os
 from datetime import timedelta
+import os
+import textwrap
+
 import flask
+import pytest
 
 
 # config keys used for the TestConfig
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -187,3 +189,15 @@ def test_get_namespace():
     assert 2 == len(bar_options)
     assert 'bar stuff 1' == bar_options['BAR_STUFF_1']
     assert 'bar stuff 2' == bar_options['BAR_STUFF_2']
+
+
+@pytest.mark.parametrize('encoding', ['utf-8', 'iso-8859-15', 'latin-1'])
+def test_from_pyfile_weird_encoding(tmpdir, encoding):
+    f = tmpdir.join('my_config.py')
+    f.write_binary(textwrap.dedent(u'''
+    # -*- coding: {} -*-
```
ThiefMaster: not compatible with 2.6; support for `{}` instead of `{0}` was added in 2.7

untitaker: Oh wow, we still support that old hag? Oh well.

untitaker:
I manually verified that the tests fail without the bugfix applied BTW

PR #2130, opened 2016-12-30
Add a largeapp example under flask/examples

wgwz:
Addresses: https://github.com/pallets/flask/issues/1902#issuecomment-269781735

Confirms that the larger application as outlined in the docs, works as expected. For the sake of less confusion, for those working through this example, having src might be helpful. Not everyone will want to work along with the docs. If there are parts of the docs that are not clear, we should fix that as well. (They seemed clear to me).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,11 @@ Flask Changelog
 
 Here you can see the full list of changes between each Flask release.
 
+Version 0.12.1
+--------------
+
+Bugfix release, unreleased
+
 Version 0.12
 ------------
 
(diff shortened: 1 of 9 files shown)


ThiefMaster:
The weird commits in this PR aside (rebase to upstream/master), I'm not sure how useful this is to have in the repo...

wgwz:
@ThiefMaster Sorry about that. I didn't choose correct base. I was thinking that in order to debug issues that people are having with code examples in the docs, it would be useful to have those examples on-hand.

Issue #2027, opened 2016-09-18
Update app factory docs for Flask's CLI


untitaker:
see https://github.com/pallets/flask/pull/1536

wgwz:
So would best practice be to create a runner/instantiation script? (For using the development server)

untitaker:
Yeah. Also I'm not sure why "multiple instances" is mentioned as usecase, since we already use locks in Flask to ensure that one app object can be used with more than one server in one process.

davidism:
I create a `FlaskGroup` and add an entry point for it, then use that entry point instead of `flask` to call any commands.

`my_app/app.py`

~~~python
from flask import Flask
from flask.cli import FlaskGroup

def create_app(info=None):
    app = Flask('my_app')
    ...
    return app

cli = FlaskGroup(create_app=create_app)
~~~

`setup.py`

~~~python
setup(
    ...,
    entry_points={
        'console_scripts': [
            'my_app=my_app:cli.main',
        ],
    },
)
~~~

wgwz:
@davidism how do you make use of the kwarg "info"?

davidism:
I don't, but it has to be there to make Click happy.  I think it can be used to pass data around within Click.

miguelgrinberg:
I think the factory function should be coded in a way that it is straightforward to call directly. I do not like the idea of adding the `ScriptInfo` argument, just because the Click integration code requires it.

I honestly haven't found a way to use an app factory with Click that matches what you can do with Flask-Script. This is how I typically code my app factory functions:

```python
def create_app(config_name=None):
    if config_name is None:
        config_name = os.environ.get('FLASK_CONFIG', 'development')
    app = Flask(__name__)
    ...
    return app
```

With this structure, Flask-Script can be given the app factory and it will call it without arguments, which runs the `development` configuration unless the `FLASK_CONFIG` environment variable is set to some other configuration. In unit tests, you just call `create_app('testing')` and you get a test ready app. In my opinion this is very handy. Depending on the application, I can add additional arguments to `create_app`, as long as all the arguments have defaults for when Flask-Scripts invokes it.

The closest I have achieved with Click requires the use of a `wsgi.py` file in which `create_app()` is called explicitly. Obviously Flask never sees the factory function, it gets the app created in wsgi.py, but at least I can write my app factory function with the arguments that I need/want.

untitaker:
[...]

PR #2131, opened 2016-12-30
Add larger app ex

wgwz:
I chose the wrong base: https://github.com/pallets/flask/pull/2130

Original PR message:

> Addresses: #1902 (comment)
> 
> Confirms that the larger application as outlined in the docs, works as expected. For the sake of less confusion, for those working through this example, having src might be helpful. Not everyone will want to work along with the docs. If there are parts of the docs that are not clear, we should fix that as well. (They seemed clear to me).

Thiefmaster's response:

> The weird commits in this PR aside (rebase to upstream/master and force-push to fix), I'm not sure how useful this is to have in the repo... Especially since having a "large" app without an app factory is somewhat questionable...

My response:

> @ThiefMaster Sorry about that. I didn't choose correct base. I was thinking that in order to debug issues that people are having with code examples in the docs, it would be useful to have those examples on-hand.
> 
> I agree. But maybe until we can put something better together for #2027, this addition could just be temporary and we could deprecate this "larger applications" example at later time.


untitaker:
This example doesn't work because it contains a circular import and precisely shows why you need app factories.

To separate views from the main app you should use blueprints.

ThiefMaster:
I think it works since the `views` import is at the bottom, after `app` has been defined. Still a very good point though.

untitaker:
Ah, I overread that this is from the docs, and yes, that actually works. I don't think we need to test this in any way though (I'd reserve `examples/` for more "complete" applications, not just patterns).

wgwz:
I'm not by any means saying this is best practice. Just that having the hardcoded example accompanying the docs, might lead to less confusion for those reading it. (And a sanity check that what's in the docs works as it should.)

untitaker:
Fair enough though this doesn't test anything. We should add a test stub as well that just imports and install the package in `tox.ini`.

kblomqvist:
A week ago I got confused by circular imports when adding `models.py` to flaskr example that I was modifying to work as a large application. I think that the large app example should also showcase how to use models.

untitaker:
Would like to see a rewrite of flaskr with flask-sqlalchemy

wgwz:
While this PR isn't for flaskr and neither was [the comment](https://github.com/pallets/flask/issues/1902#issuecomment-269781735) that prompted me to work on this, I do agree. flaskr could be overhauled to show current best practices; i.e. app factories, blueprints, flask-sqlalchemy and the CLI, which also helps with #2027. Maybe at that point this whole larger app portion of the docs can be written with respect to flaskr. 

> Fair enough though this doesn't test anything because pytest doesn't import the package. We should add a test stub as well that just imports and install the package in tox.ini.

@untitaker assuming you want to go forward with this, would *tox.ini* change as below? and what else needs to change/be added? I don't follow what you are saying. (not familiar with tox.)

```
commands =
    # We need to install those after Flask is installed.
    pip install -e examples/flaskr
    pip install -e examples/minitwit
    pip install -e examples/patterns/largerapp
    py.test --cov=flask --cov-report html []
```

untitaker:
@wgwz Yes exactly, it's just about installing the application in the virtualenv (otherwise you can't import it in the tests)

wgwz:
@untitaker OK I understand now. Should I add some basic tests for it under `tests/test_apps`?

untitaker:
The other example apps had their own testfolders iirc

On 31 December 2016 17:53:14 CET, Kyle Lawlor <notifications@github.com> wrote:
>@untitaker OK I understand now. Should I add some basic tests for it
>under `tests/test_apps`? 
>
>-- 
>https://github.com/pallets/flask/pull/2131#issuecomment-269873161

-- 

wgwz:
should this be closed?

untitaker:
I think it's fine as is, thanks :)

PR #2133, opened 2016-12-30
prevent NoAppException when ImportError occurs within imported module

pawl:
Currently this will cause `flask run` to show a `NoAppException` error:
```
from flask import Flask

raise ImportError()

testapp = Flask('testapp')
``` 

The expected behavior is to see the `ImportError` when you visit the page instead of the  `NoAppException` error.

This happens because the code does not differentiate between an `ImportError` when `flask run` tries to import the module and an `ImportError` from the module after it is imported.

The fix came from here: http://stackoverflow.com/a/24262408

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,8 @@ Version 0.12.1
 
 Bugfix release, unreleased
 
+- Prevent `flask run` from showing a NoAppException when an ImportError occurs
+  within the imported application module.
 - Fix encoding behavior of ``app.config.from_pyfile`` for Python 3. Fix
   ``#2118``.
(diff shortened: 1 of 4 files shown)


untitaker:
Please add a changelog entry as well

untitaker:
Applied, thanks!

stephane:
👍

iurisilvio:
This change deserves a bug fix release. Flask gives you no clue about the exception.

nathancahill:
Second bug fix release. This currently swallows all import errors as `NoAppException`

PR #2135, opened 2016-12-30
Inherit Werkzeug docs

### LINKED ISSUE #2132: Flask documentation inconsistent with implementation for form parsing
Since Werkzeug 0.9, request.form is populated for all requests as long as they contain a Content-Type header (https://github.com/pallets/werkzeug/blob/master/werkzeug/wrappers.py#L328).
The Flask documentation however implies that request.form is only available with POST and PUT requests, and that is the case with Werkzeug<0.9 (well, 0.8.3 parses for POST, PUT and PATCH).

untitaker:
Fix #2132

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -30,60 +30,11 @@ Incoming Request Data
 
 .. autoclass:: Request
    :members:
-
-   .. attribute:: form
-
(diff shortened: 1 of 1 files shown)


jeffwidman:
Nice! I never knew Sphinx could do that.

PR #2136, opened 2016-12-31
Added missing html tag in wtforms docs

marcanuy:


ThiefMaster:
The tags/quotes aren't missing. Omitting them is valid HTML5.

PR #2144, opened 2017-01-08
Allow default Errorhandler for HTTPException

sumpfgottheit:
Allow an default errorhandler for all HTTPException.

This will come handy when building a REST API where you want all Errors return JSON instead of the default behaviour.

PR https://github.com/pallets/flask/pull/2082 solves the same problem, but it didn't work in my environment. Tests are included.


jeffwidman:
Thanks for this! Closing in favor of the #2304 which takes this and extends it.

PR #2146, opened 2017-01-08
Added python 3.6

andriisoldatenko:


AUTHORS
```diff
@@ -11,6 +11,7 @@ Patches and Suggestions
 
 - Adam Zapletal
 - Ali Afshar
+- Andrii Soldatenko
```
ThiefMaster: Not sure if that makes lots of sense for such a small change (even though such changes are very welcome of course!)

There are commits from 368 people in this repo, but way less entries in this file (adding every contributor would quickly make the file pretty chaotic).

IMO it'd be better to only add yourself to this file when you contribute e.g. a major fix, new feature, etc

andriisoldatenko: @ThiefMaster Sorry for that. Will remove myself.
Also I'll try to fix unit tests on Python 3.6

untitaker:
In principle you can't give people too much credit.

untitaker:
Also @andriisoldatenko there are a few tests failing. The first one I don't understand, the second one is trivial.

ThiefMaster:
The first failure is probably easier to figure out when replacing `assert not recwarn.list` with this:


    earnings = recwarn.list
    assert not recwarn.list

Hopefully pytest shows the actual list contents then instead of just `<...>.list`

davidism:
Thanks, but #2148 got there first with the test fixes.

PR #2147, opened 2017-01-10
add missing quotes for dict key in views.rst

kamyarg:


ThiefMaster:
The quotes are not missing. The key is taken from the `pk` argument.

kamyarg:
Wow, sorry. 
I missed that. 😅

PR #2148, opened 2017-01-10
Added python3.6 support for tests

drewja:
python 3.6 has deprecated the use of flags not at the start of regular expression,
plans for 3.7 is to rasie an exception: [http://bugs.python.org/issue22493](http://bugs.python.org/issue22493)

new on 3.6  [ModuleNotFoundError](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError)

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -333,7 +333,7 @@ def test_session_expiration():
     client = app.test_client()
     rv = client.get('/')
     assert 'set-cookie' in rv.headers
-    match = re.search(r'\bexpires=([^;]+)(?i)', rv.headers['set-cookie'])
+    match = re.search(r'(?i)\bexpires=([^;]+)', rv.headers['set-cookie'])
     expires = parse_date(match.group())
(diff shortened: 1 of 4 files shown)


tests/test_basic.py
```diff
@@ -333,7 +333,7 @@ def test():
     client = app.test_client()
     rv = client.get('/')
     assert 'set-cookie' in rv.headers
-    match = re.search(r'\bexpires=([^;]+)(?i)', rv.headers['set-cookie'])
+    match = re.search(r'(?i)\bexpires=([^;]+)', rv.headers['set-cookie'])
```
ThiefMaster: Why not use `flags=re.IGNORECASE` instead of the - rather ugly - inline syntax?

drewja: I have no preference, I adapted the inline syntax because that is what was already being used.

davidism: A couple other Pallets projects have merged patches extracting the flags. For example, pallets/werkzeug# 1053. If you're interested in creating a similar one for Flask that would be appreciated.

PR #2151, opened 2017-01-14
Fix Request Reference

barracks510:
Points flask.Request to appropriate place in the documentation.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -538,16 +538,16 @@ The Request Object
 ``````````````````
 
 The request object is documented in the API section and we will not cover
-it here in detail (see :class:`~flask.request`). Here is a broad overview of
+it here in detail (see :class:`~flask.Request`). Here is a broad overview of
 some of the most common operations.  First of all you have to import it from
 the ``flask`` module::
 
     from flask import request
 
 The current request method is available by using the
-:attr:`~flask.request.method` attribute.  To access form data (data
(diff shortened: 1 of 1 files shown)


jeffwidman:
I'm actually a little confused... the current docs link to something called `class flask.request` which is listed under the proper `class flask.Request`... why do the docs distinguish the `request` object as it's a class? Is that correct or should it be changed?

untitaker:
@jeffwidman `Request` is a class, `request` is the current instance of it.

untitaker:
@barracks510 if this is not showing right in the docs I'm happy to learn about improvements, but there's a clear difference between those two paths. hence I'd tend to close this

ThiefMaster:
IMO at least the changes on the tilde links make sense - so e.g. a mention to `method` points to the `Request.method` documentation. Since only the last segment is displayed it would only change (or rather add, currently there is no link) the link target

untitaker:
Ah yes, I oversaw that.

untitaker:
Thanks @barracks510!

jeffwidman:
> Request is a class, request is the current instance of it.

Right, so why is the word ["class" located next to the `request` object](http://flask.pocoo.org/docs/0.12/api/#flask.request)? 
It visually makes it seem like it's a separate class from `Request`...

untitaker:
That's an entirely valid complaint, I guess `attr` would be more appropriate.

Issue #2109, opened 2016-12-14
app.run(host, port) and SERVER_NAME should correspond to where flask is actually running


MartijnRas:
Assuming the SERVER_NAME config variable is configured as '0.0.0.0:5555':

1) Calling app.run(host=None, port=None) results in flask running on http://127.0.0.1:5555/
2) Calling app.run(host='127.0.0.1', port=5000) results in flask running on http://127.0.0.1:5000/

In both cases the SERVER_NAME config variable does not correspond to where flask is actually running.

The host and port parameters in app.run() must of course take precedence, but should be set according to the SERVER_NAME config variable if present.

The current implementation in app.run():

    if host is None:
        host = '127.0.0.1'
    if port is None:
        server_name = self.config['SERVER_NAME']
        if server_name and ':' in server_name:
            port = int(server_name.rsplit(':', 1)[1])
        else:
            port = 5000

Could be replaced by the following:

    _host = '127.0.0.1'
    _port = 5000
    servername = self.config['SERVER_NAME']
    if server_name:
        if server_name and ':' in server_name:
            _host, _port = servername.split(':', 1)
    if host is None:
        host = _host
    if port is None:
        port = _port

The following can be used to make sure the SERVER_NAME config variable corresponds to where flask is actually running:

    if servername:
        self.config['SERVER_NAME'] = host + ':' + port

davidism:
0.0.0.0:5000 is not a valid server name.

davidism:
This won't (a cannot due to lazy loading) apply to the `flask run` command, which is the recommended way to run the dev server now.

jeffwidman:
Seems like this should be closed based on above comments?

davidism:
Yes. Also, server name isn't necessarily the same as the local address the server is bound to.

ralex2304:
How can I make 2 server names?
"example.com" and "www.example.com".

ivanlmj:
Is there any way of putting the `host` and `port` variables in a separate file?

For example, defining a configuration file and having these variables inside of it, instead of defining these exactly on `app.run()`:

    app.config.from_pyfile('settings.cfg') 
    app.run()

mrshelly:
??
Thank you , For read this Email !

________________________________
发件人: ivanleoncz <notifications@github.com>
发送时间: 2018年8月12日 2:09
收件人: pallets/flask
抄送: Subscribed
[...]

PR #2152, opened 2017-01-16
use SERVER_NAME to set host and port in app.run()

r0fls:
Addresses #2109

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -830,7 +830,7 @@ class Flask(_PackageBoundObject):
         _port = 5000
         server_name = self.config.get("SERVER_NAME", None)
         if server_name:
-            _host, _port = server_name.split(':', 1)
+            _host, _, _port = server_name.partition(':')
         host = host or _host
         if _port:
             port = port or int(_port)
--- a/flask/app.py
+++ b/flask/app.py
@@ -825,10 +825,10 @@ class Flask(_PackageBoundObject):
                         information.
         """
         from werkzeug.serving import run_simple
-
         _host = '127.0.0.1'
         _port = 5000
         server_name = self.config.get("SERVER_NAME")
+        sn_host, sn_port = None, None
         if server_name:
             sn_host, _, sn_port = server_name.partition(':')
         host = host or sn_host or _host
--- a/flask/app.py
+++ b/flask/app.py
@@ -828,12 +828,11 @@ class Flask(_PackageBoundObject):
 
         _host = '127.0.0.1'
         _port = 5000
-        server_name = self.config.get("SERVER_NAME", None)
+        server_name = self.config.get("SERVER_NAME")
         if server_name:
-            _host, _, _port = server_name.partition(':')
-        host = host or _host
-        if _port:
-            port = port or int(_port)
+            sn_host, _, sn_port = server_name.partition(':')
+        host = host or sn_host or _host
+        port = int(port or sn_port or _port)
         if debug is not None:
             self.debug = bool(debug)
         options.setdefault('use_reloader', self.debug)
--- a/flask/app.py
+++ b/flask/app.py
@@ -825,14 +825,15 @@ class Flask(_PackageBoundObject):
                         information.
         """
         from werkzeug.serving import run_simple
-        if host is None:
-            host = '127.0.0.1'
-        if port is None:
-            server_name = self.config['SERVER_NAME']
-            if server_name and ':' in server_name:
-                port = int(server_name.rsplit(':', 1)[1])
-            else:
-                port = 5000
+
+        _host = '127.0.0.1'
+        _port = 5000
+        server_name = self.config.get("SERVER_NAME", None)
+        if server_name:
+            _host, _port = servername.split(':', 1)
+        host = host or _host
+        if _port:
+            port = port or int(_port)
         if debug is not None:
             self.debug = bool(debug)
         options.setdefault('use_reloader', self.debug)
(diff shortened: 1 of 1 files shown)


jeffwidman:
LGTM

flask/app.py
```diff
@@ -825,14 +825,13 @@ def run(self, host=None, port=None, debug=None, **options):
...
-                port = int(server_name.rsplit(':', 1)[1])
-            else:
-                port = 5000
+
+        _host = '127.0.0.1'
+        _port = 5000
+        server_name = self.config.get("SERVER_NAME", None)
+        if server_name:
+            _host, _port = servername.split(':', 1)
+        host, port = host or _host, port or int(_port)
```
davidism: Don't use tuple unpacking here.

davidism: This fails if there was no port in `SERVER_NAME`.

flask/app.py
```diff
@@ -825,14 +825,15 @@ def run(self, host=None, port=None, debug=None, **options):
...
-            if server_name and ':' in server_name:
-                port = int(server_name.rsplit(':', 1)[1])
-            else:
-                port = 5000
+
+        _host = '127.0.0.1'
+        _port = 5000
+        server_name = self.config.get("SERVER_NAME", None)
+        if server_name:
+            _host, _port = servername.split(':', 1)
```
jeffwidman: `server_name` rather than `servername`?

r0fls:
Yikes, thanks @jeffwidman. Any pointers on tests we could write to catch some of these types of situations?

flask/app.py
```diff
@@ -825,14 +825,15 @@ def run(self, host=None, port=None, debug=None, **options):
...
-            if server_name and ':' in server_name:
-                port = int(server_name.rsplit(':', 1)[1])
-            else:
-                port = 5000
+
+        _host = '127.0.0.1'
+        _port = 5000
+        server_name = self.config.get("SERVER_NAME", None)
+        if server_name:
+            _host, _port = server_name.split(':', 1)
```
davidism: This still fails if there's no port.

ThiefMaster: https://docs.python.org/2/library/stdtypes.html#str.partition is somewhat useful for cases like this one

flask/app.py
jeffwidman: Good catch. I forgot about that. I like the way this is structured though, reads much cleaner to my eyes than the previous logic.

davidism:
Is there a reason you didn't include the "set `SEVER_NAME` from `run`" part of the issue? It would require some more thought because you can't set the config to `0.0.0.0` (even though you shouldn't be running with that anyway).

davidism:
Posted this in the linked issue, but note that this won't and can't apply to the `flask run` command, which is the recommended way to run the dev server.

jeffwidman:
I haven't looked at the original issue in depth, but you got me curious on how to use `partition` properly... played with it and looks like something like this would properly handle the double fallback from `app.run` to `SERVER_NAME` to the built-in default. And it does it separately for both host and port values so you can specify one or the other or both:

    _host = '127.0.0.1'
    _port = 5000
    server_name = self.config.get("SERVER_NAME", None)
    if server_name:
        _sn_host, _, _sn_port = server_name.partition(':')
    host = host or _sn_host or _host
    port = int(port or _sn_port or _port)

flask/app.py
```diff
@@ -825,14 +825,15 @@ def run(self, host=None, port=None, debug=None, **options):
...
-        if port is None:
-            server_name = self.config['SERVER_NAME']
-            if server_name and ':' in server_name:
-                port = int(server_name.rsplit(':', 1)[1])
-            else:
-                port = 5000
+
+        _host = '127.0.0.1'
+        _port = 5000
+        server_name = self.config.get("SERVER_NAME", None)
```
jeffwidman: `None` is the default fallback, no need to specify it

r0fls: I always forget that.

flask/app.py
```diff
@@ -825,14 +825,14 @@ def run(self, host=None, port=None, debug=None, **options):
...
-                port = int(server_name.rsplit(':', 1)[1])
-            else:
-                port = 5000
+
+        _host = '127.0.0.1'
+        _port = 5000
+        server_name = self.config.get("SERVER_NAME")
+        if server_name:
+            sn_host, _, sn_port = server_name.partition(':')
+        host = host or sn_host or _host
```
davidism: This will fail if `SERVER_NAME` wasn't set, since `sn_host` won't be set.

r0fls:
Should I rebase to 1 commit at this point, or does that not matter?

davidism:
Not sure why the Python 2.6 tests started failing, but it seems unrelated.

r0fls:
Ya, it looks to be a message saying 2.6 is no longer supported or maintained by the core devs.

untitaker:
That is true but Flask still supports it. However, it appears that coverage.py no longer does.

untitaker:
This needs a testcase and changelog.

davidism:
Done.

Re: 2.6, it was fixed in Coverage earlier today, tests pass again. https://coverage.readthedocs.io/en/coverage-4.3.4/changes.html#version-4-3-4-2017-01-17

johnbodley:
@davidism I was wondering whether the docstring for the `host` param should now reflect the updated behavior, i.e. similar to `port`,  

> ... or the host defined in the `SERVER_NAME` config variable if present.

jeffwidman:
That is probably a good idea, do you want to open a PR?

PR #2155, opened 2017-01-17
Disable debug when FLASK_DEBUG=False

kblomqvist:
I got confused why setting FLASK_DEBUG environment variable to `False`, as it's written in Python, didn't disable debug.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -58,7 +58,7 @@ def get_debug_flag(default=None):
     val = os.environ.get('FLASK_DEBUG')
     if not val:
         return default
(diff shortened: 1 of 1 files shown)


flask/helpers.py
```diff
@@ -58,7 +58,7 @@ def get_debug_flag(default=None):
     val = os.environ.get('FLASK_DEBUG')
     if not val:
         return default
-    return val not in ('0', 'false', 'no')
+    return val not in ('0', 'False', 'false', 'No', 'no')
```
davidism: `val.lower() in ...`

PR #2163, opened 2017-01-29
Add support for custom console-scripts

ThiefMaster:
For an installed app there is currently no clean way to provide a custom (e.g. `testapp`) console script instead of using the `flask` script (which needs FLASK_APP to be set) - using any entry point that resides within the app's package would stop the reloader from surviving import-time errors in the app's package.

With this PR it's easy to define a custom console script and also specify its app or app factory, making `FLASK_APP` unnecessary when using that script.

The application's setup.py would simply have to add this to the `setup()` call:

```python
    entry_points={
        'console_scripts': [
            'testapp = flask.cli:main'
        ],
        'flask.cli_apps': [
            'testapp = testapp.app:make_app'
        ]
    }
```

---

### TODO

- [x] Override default FlaskGroup help for custom commands (do not mention `FLASK_APP`)
- [x] Changelog
- [x] Docs
- [ ] Tests?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,8 @@ Major release, unreleased
 - Make `app.run()` into a noop if a Flask application is run from the
   development server on the command line.  This avoids some behavior that
   was confusing to debug for newcomers.
+- Add support for custom cli scripts which automatically create an app without
+  the need for an environment variable.  ``#2163``
 
 Version 0.12.1
 --------------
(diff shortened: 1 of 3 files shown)


kennethreitz:
needs a rebase!

kennethreitz:
this PR is old. this complicates documentation. closing due to inactivity.

ThiefMaster:
Agreed, I was actually going to close it myself a few days ago and then got distracted. In the end it's not too hard to use the "normal" way of adding a custom console script - as long as imports from the app are kept to the minimum the dev server survives restarts in case of almost all syntax errors.

kennethreitz:
✨🍰✨

PR #2173, opened 2017-02-07
py.test => pytest

vojtekb:
Really a minor thing, only experienced by a newcomer.
When following the README or CONTRIBUTE  docs, the first test run results in lots of failures.
This is because a simple `py.test` also discovers the tests in examples/

(My plan is to contribute more, this is just to get started ;)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -38,7 +38,7 @@ Running the testsuite
 You probably want to set up a `virtualenv
 <https://virtualenv.readthedocs.io/en/latest/index.html>`_.
 
-The minimal requirement for running the testsuite is ``py.test``.  You can
+The minimal requirement for running the testsuite is ``pytest``.  You can
 install it with::
 
     pip install pytest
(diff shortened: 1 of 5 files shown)


ThiefMaster:
Recent pytest versions prefer `pytest` over `py.test` (the latter is deprecated and will eventually be removed). Since the guide recommends to install the latest pytest version I think it'd be a good idea to use this opportunity to run pytest `pytest`.

ThiefMaster:
Actually, if pytest goes into a directory it shouldn't go in the best solution would be adding it to [`norecursedirs`](https://github.com/pallets/flask/blob/master/setup.cfg#L11)

untitaker:
It is the plan to also test the examples, such that if you modify them, they
get tested as well.

That's not a common case, but I don't think they take that much time either.

vojtekb:
`norecursedirs` seems like a better idea. 
`pytest` command is indeed preferred over `py.test` (https://github.com/pytest-dev/pytest/issues/1629) so I changed the docs and invocations as well.

untitaker:
Thanks!

untitaker:
However, I reverted your change to add `examples` to `norecursedirs`. As said there are tests in there that do need to be run.

PR #2175, opened 2017-02-09
Update 4.4.3 HTTP Methods Example

swanhtet1992:
I just started learning `flask` and following the documentation pdf. When I reached the topic `4.4.3 HTTP Methods`, I encountered this `ValueError: View function did not return a response` even though I followed the tutorial. It made me confused and had to search on [stackoverflow](http://stackoverflow.com/a/18211982).

This might not be a good fix for an example but the previous example did confuse a beginner, me, who followed the documentation.

Cheers,
swan


untitaker:
Fair enough!

PR #2187, opened 2017-02-19
Fix non-working  setup.cfg in tutorial docs with working copy from sa…

mitchjacksontech:
I worked through the tutorial today, but ended up with code that wouldn't run.
The provided example code in the repo does run.  
This pull matches the code in the docs to the example code

### DIFF
--- a/docs/tutorial/testing.rst
+++ b/docs/tutorial/testing.rst
@@ -77,7 +77,7 @@ update the :file:`setup.py` file to contain::
 Now create :file:`setup.cfg` in the project root (alongside
 :file:`setup.py`)::
 
-    [aliases]
+    [tool:pytest]
     test=pytest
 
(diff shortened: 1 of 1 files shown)


davidism:
An alias so that `test` runs `pytest` is valid.

mitchjacksontech:
valid yet not working

untitaker:
What do you mean with "not working". What exactly crashes?

davidism:
The change you made is invalid, no matter what behaviour you may be seeing. `test=pytest` is not a `pytest` config value. You'll need to describe the issue you're having in order for us to understand the problem.

mitchjacksontech:
My apologizes, I must have misunderstood what I saw... I'll work through it from scratch and come back with full details.

PR #2193, opened 2017-03-01
Don't rely on X-Requested-With for pretty print json response

lepture:
`X-Requested-With` is not reliable nowadays. We shouldn't rely on it any more.

Related issue: https://github.com/pallets/werkzeug/issues/1077

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -178,7 +178,7 @@ The following configuration values are used internally by Flask:
                                   This is not recommended but might give
                                   you a performance improvement on the
                                   cost of cacheability.
-``JSONIFY_PRETTYPRINT_REGULAR``   If this is set to ``True`` or Flask app
+``JSONIFY_PRETTYPRINT_REGULAR``   If this is set to ``True`` or the Flask app
                                   is running in debug mode, jsonify responses
                                   will be pretty printed.
 ``JSONIFY_MIMETYPE``              MIME type used for jsonify responses.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -178,11 +178,9 @@ The following configuration values are used internally by Flask:
                                   This is not recommended but might give
                                   you a performance improvement on the
                                   cost of cacheability.
-``JSONIFY_PRETTYPRINT_REGULAR``   If this is set to ``True`` (the default)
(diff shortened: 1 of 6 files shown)


untitaker:
This makes sense to me.

ThiefMaster:
I wonder if prettyprinting JSON is that useful at all as a Flask feature...

AFAIK most browsers ask you to download a file with a JSON content type so by default you won't even see it in the browser.
And when installing one of the many browser extensions that show JSON nicely instead, they usually take care of prettyprinting it.

docs/config.rst
```diff
@@ -178,11 +178,9 @@ The following configuration values are used internally by Flask:
                                   This is not recommended but might give
                                   you a performance improvement on the
                                   cost of cacheability.
-``JSONIFY_PRETTYPRINT_REGULAR``   If this is set to ``True`` (the default)
-                                  jsonify responses will be pretty printed
-                                  if they are not requested by an
-                                  XMLHttpRequest object (controlled by
-                                  the ``X-Requested-With`` header)
+``JSONIFY_PRETTYPRINT_REGULAR``   If this is set to ``True`` or Flask app
```
ThiefMaster: missing *the*

untitaker:
It might be less useful than it used to be if you think of browsers as your primary client, though you might have different clients which don't have as good devtools for JSON (self-written clients e.g., or just cURL)

ThiefMaster:
ah true, forgot that people might still use curl instead of httpie ;)

docs/config.rst
lepture: fixed.

lepture:
Anyone against this patch?

untitaker:
We need a proper changelog entry. Also there's an upgrading page in the docs
that probably should have a new section

lepture:
I don't think we need an upgrading guide. This patch doesn't break things. A change log is enough.

untitaker:
Fair enough

alanhamlett:
Darn, I really liked the pretty printed JSON when viewed in Chrome. Made linking to example json responses from docs look nice.

davidism:
Use Firefox, it has a built in json view. ;-)

PR #2196, opened 2017-03-03
Add tips for debug config with flask cli

greyli:
`app.debug` and `app.config['DEBUG']` are not compatible with the `flask` script. The app's config is not available in `get_debug_flag()` when starting the server, is there a way to fix that?

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -56,7 +56,7 @@ method::
    (On Windows you need to use ``set`` instead of ``export``).
    
    ``app.debug`` and ``app.config['DEBUG']`` are not compatible with 
-   the :command:`flask` script. They only worked when use ``Flask.run()``
+   the :command:`flask` script. They only worked when using ``Flask.run()``
    method.
    
 Builtin Configuration Values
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -44,6 +44,21 @@ method::
...
+   server, to enable the debug mode, you need to export the ``FLASK_DEBUG``
+   environment variable before running the server::
+
+    $ export FLASK_DEBUG=1
+    $ flask run
+
+   (On Windows you need to use ``set`` instead of ``export``).
+   
+   ``app.debug`` and ``app.config['DEBUG']`` are not compatible with 
+   the :command:`flask` script. They only worked when use ``Flask.run()``
```
untitaker: "using" instead of "use"

greyli: Thanks!

untitaker:
Thanks!

docs/config.rst
```diff
@@ -44,6 +44,21 @@ method::
         SECRET_KEY='...'
     )
 
+.. admonition:: Debug Mode with the ``flask`` Script
```
kevin-brown: This isn't auto-linking the header (for deep linking) because you are using a code block within the header.

PR #2199, opened 2017-03-05
Close <li> tag in lines (16,18) (#1951)

eltonlaw:
Adding a change  in the example code that wasn't added to the docs. https://github.com/pallets/flask/commit/c54d67adee6f99113f525b376da4af27c3001321


ThiefMaster:
The end tag is optional for `<li>`.

ThiefMaster:
Ah, just saw the referenced commit. Makes sense to keep it in sync then.

PR #2207, opened 2017-03-15
Edit debug mode attention for new version Werkzeug

greyli:
Starting from Werkzeug 0.11, the debugger was protected by a PIN.


davidism:
The wording of this is more awkward than the original.

greyli:
Is it better now?

davidism:
Thanks, but I don't find this more useful. The original statement was about how code execution was possible even though the interface itself wasn't useful in production deployments.

PR #2209, opened 2017-03-16
Print a stacktrace on CLI error (closes #2208)

### LINKED ISSUE #2208: Do not swallow exceptions if CLI fails to load
I think the changes introduced in https://github.com/pallets/flask/commit/fa6eded6f572dd4bc23b030f025156cdd1e63305 that swallow ALL exceptions if loading a custom CLI command fails are not good for the general development flow. I often do `flask --help` to see whether my custom command comes up fine and when it doesn't I don't get exceptions and I don't really know what I've done wrong.

I think this `try-except` should be removed. I really think the CLI _should_ fail loudly instead of ignoring errors. Let's discuss this and if you guys agree I'll make a PR.

svenstaro:
I've been bitten by this a few times and I believe silently swallowing errors without any debug output is a bad thing to do. This at least prints the resulting stacktrace.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -11,6 +11,7 @@
 
 import os
 import sys
+import traceback
 from threading import Lock, Thread
 from functools import update_wrapper
 
@@ -368,6 +369,9 @@ class FlaskGroup(AppGroup):
             # want the help page to break if the app does not exist.
             # If someone attempts to use the command we try to create
             # the app again and this will give us the error.
+            # However, we will not do so silently because that would confuse
+            # users.
+            traceback.print_exc()
             pass
         return sorted(rv)
 
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -372,7 +372,6 @@ class FlaskGroup(AppGroup):
             # However, we will not do so silently because that would confuse
             # users.
             traceback.print_exc()
-            pass
         return sorted(rv)
 
     def main(self, *args, **kwargs):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -191,3 +191,20 @@ def test_flaskgroup():
     result = runner.invoke(cli, ['test'])
     assert result.exit_code == 0
     assert result.output == 'flaskgroup\n'
+
+
+def test_print_exceptions():
+    """Print the stacktrace if the CLI."""
+    def create_app(info):
+        raise Exception("oh no")
+        return Flask("flaskgroup")
+
+    @click.group(cls=FlaskGroup, create_app=create_app)
+    def cli(**params):
+        pass
+
+    runner = CliRunner()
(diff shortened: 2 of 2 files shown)


flask/cli.py
```diff
@@ -368,6 +369,9 @@ def list_commands(self, ctx):
             # want the help page to break if the app does not exist.
             # If someone attempts to use the command we try to create
             # the app again and this will give us the error.
+            # However, we will not do so silently because that would confuse
+            # users.
+            traceback.print_exc()
             pass
```
ThiefMaster: That's not needed since there's some actual code in there now.

svenstaro: Oh right.

svenstaro:
@davidism Done.

untitaker:
This requires a changelog

Kamekameha:
I know this wasn't completely introduced here but in conjunction with the commit it is trying to fix, fa6eded6f572dd4bc23b030f025156cdd1e63305, as mentioned in #2208. However, a traceback is now always printed for cases where there's no app in the first place. This is pretty much just relevant to the `--help` option on the main `flask` command, since most likely everything else requires an app to properly run.

In short, a more appropriate fix would be to special-case `NoAppException` to print no traceback at all, and keep the introduced behavior as is for every other exception.

But then we have at least two places where a `NoAppException` is raised because of another exception, of a `TypeError`:

https://github.com/pallets/flask/blob/9291ead32e2fc8b13cef825186c968944e9ff344/flask/cli.py#L77-L85

And:

https://github.com/pallets/flask/blob/9291ead32e2fc8b13cef825186c968944e9ff344/flask/cli.py#L150-L156

Which could (unlikely) very well be errors on the script side rather than the call to the `call_factory` function itself, where in such cases we already get kind of misleading error messages ("kind of" because in the second case the handled exception message is indeed added to the new one), but moreover, such previously mentioned fix will end up swallowing these `NoAppException`s, losing whatever that could be of use from these exceptions (especially in Python 3 with exception chaining).

To handle this more correctly would be to do a stack depth check like in https://github.com/pallets/flask/blob/9291ead32e2fc8b13cef825186c968944e9ff344/flask/cli.py#L207 and re-raise appropriately so that exceptions are handled as expected, if inspecting and validating the whole function signature in `call_factory` is not feasible/desirable. This should actually be done regardless of whether `NoAppException` is special-cased or not in `list_commands` because of the mentioned reasons right above.

Alternatively to special-casing `NoAppException`, revert the error stack-trace printing since it's actually unexpected and slightly intrusive getting so on a help message that didn't exactly fail, and so testing the app's custom commands goes simply by running it or by similar means, given that users must do so either way and these exceptions will definitely be raised in such scenarios. Of course, that's just my opinion on it.

PR #2211, opened 2017-03-18
Fix static route when using host_matching = True

### LINKED ISSUE #1559: url_for('static', _external=True) gives bogus url with app.url_map.host_matching = True

jab:
Change `Flask.__init__` to accept a new `static_host` keyword argument,
defaulting to None. <s>Passing this results in url_map.host_matching to be
set to True implicitly. This is because host_matching must be set by the
time the static route is added, or else it won't match.</s> (Edit: see updates below.)

Fixes #1559.
Subsumes #1560.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -131,8 +131,8 @@ class Flask(_PackageBoundObject):
                           at `static_url_path`.  Defaults to the ``'static'``
                           folder in the root path of the application.
     :param static_host: the host to use when adding the static route.
-                        Defaults to None. Setting this implicitly causes
-                        ``app.url_map.host_matching`` to be set to True.
+                        Defaults to None. Only use if setting
+                        ``app.url_map.host_matching`` to True.
     :param template_folder: the folder that contains the templates that should
                             be used by the application.  Defaults to
                             ``'templates'`` folder in the root path of the
@@ -534,18 +534,26 @@ class Flask(_PackageBoundObject):
         self._got_first_request = False
         self._before_request_lock = Lock()
 
-        # Register the static folder for the application if one is configured.
-        # Note we do this without checking if the configured folder exists.
+        # Create a callback that adds a route for static requests using the
+        # provided static_url_path, static_host, and static_folder,
+        # iff there is a configured static_folder.
+        # Note we do this without checking if static_folder exists.
         # For one, it might be created while the server is running (e.g. during
         # development). Also, Google App Engine stores static files somewhere
         # else when mapped with the .yml file.
-        if self.has_static_folder:
-            if static_host:  # Passing static_host implies host_matching = True.
-                # Must be set before adding the url rule or else it won't match.
-                self.url_map.host_matching = True
+        # This is deferred until right before the first request, because the
+        # user may set app.url_map.host_matching to True after the app has
+        # been initialized, and host_matching needs to be checked for at the
+        # time the route is added for requests to be routed correctly.
+        self._should_add_static_route = self.has_static_folder
+        def add_static_route():
+            assert not static_host or self.url_map.host_matching, (
+                'Expected app.url_map.host_matching to be True when static_host provided')
             self.add_url_rule(self.static_url_path + '/<path:filename>',
                               endpoint='static', host=static_host,
                               view_func=self.send_static_file)
+
+        self._add_static_route = add_static_route
 
         #: The click command line context for this application.  Commands
         #: registered here show up in the :command:`flask` command once the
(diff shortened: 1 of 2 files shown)


jab:
It looks like the build is failing for Python 3 here due to the known, unrelated issue #513

tests/test_basic.py
```diff
@@ -1171,6 +1171,16 @@ def test_static_url_path():
         assert flask.url_for('static', filename='index.html') == '/foo/index.html'
 
 
+def test_static_route_with_host_matching():
+    app = flask.Flask(__name__, static_host='example.com')
+    assert app.url_map.host_matching, 'passing static_host implies host_matching = True'
+    c = app.test_client()
+    assert c.get('http://example.com/static/index.html').status_code == 200
```
jab: This would give 404 without this fix.

tests/test_basic.py
```diff
@@ -1171,6 +1171,16 @@ def test_static_url_path():
...
+    with app.test_request_context():
+       rv = flask.url_for('static', filename='index.html', _external=True)
+       assert rv == 'http://example.com/static/index.html'
```
jab: This would give `http:///static/foo` without this fix.

jab:
> It looks like the build is failing for Python 3 here due to the known, unrelated issue #513

(same thing happened again)

flask/app.py
```diff
@@ -1982,6 +1997,10 @@ def wsgi_app(self, environ, start_response):
                                a list of headers and an optional
                                exception context to start the response
         """
+        with self._before_request_lock:
+            if not self._got_first_request and self._should_add_static_route:
+                self._add_static_route()
+                self._should_add_static_route = False
```
untitaker: Why do we need this variable? We can just check `if self._got_first_request and self.has_static_folder: ...` 

I also don't understand why it's `not self._got_first_request` here.

jab: I was concerned that only using the `_got_first_request` guard inside the new, earlier `with self._before_request_lock` block I added\*, could introduce a race condition:

If the first two requests come in at almost the same time, the first one enters the first `with ...lock` block and adds the static route, then the second one enters that first `with ...lock` block and *re*-adds the static route, *before* the first  request enters the second `with ...lock` block and finally flips the `_got_first_request` guard to True. So I added the additional guard for whether we've added the static route yet.

But after your comment, and upon taking another look, I'm wondering whether this race condition can actually happen, or whether adding the same route twice is not a big deal in any case. If not, I've updated the logic to match your suggestions in 8ba6f35. If that's sound, this should now be ready to merge, otherwise I'm happy to work through this further.

Thanks for taking another look.

\* which comes before the existing `with self._before_request_lock` block in `try_trigger_before_first_request_functions()` (called by `full_dispatch_request()`, which is called by `wsgi_app()` a few lines after the logic I added), and which is the block that finally sets the `_got_first_request` guard to True

jab: (Zooming out for a second, is this kind of confusion a sign that the control flow here could be cleaned up in a future overhaul? If not, please feel free to ignore. /cc @mitsuhiko)

untitaker:
You won't have to worry about any of this if you register your
`add_static_route` as `before_first_request_func`.

jab:
At first that's what I thought too, and that's how I tried to do this initially, but it doesn't work. The problem is the existing `with self._before_request_lock` block runs after the `ctx = self.request_context(environ)` line. By that time, a url adapter has already been created out of the request, and sets a NotFound error on it. So if the first request was for a static resource, Flask would incorrectly serve a 404. Try it yourself.

Unless I'm somehow missing something obvious, since this is tough for even maintainers to follow, it does seem that at some point a cleanup of the control flow is needed.

jab:
Regarding the question of what happens if the static route were added a second time, it looks like it would raise `AssertionError: View function mapping is overwriting an existing endpoint function: static`, so no good.

jab:
@untitaker et al., in light of the above, do you prefer registering the static route at app initialization time, and accepting a new `__init__` parameter for whether to set app.url_map.host_matching to True?

untitaker:
Sorry for the slow response. Considering this makes control flow only more complicated I think I'm for the solution of your previous PR.

untitaker:
Reopened https://github.com/pallets/flask/pull/1560

Issue #2212, opened 2017-03-19
Documentation: Installing and virtual environment under python3


runartrollet:
For installing Flask under a virtual environment with python3, the procedure is somewhat different that what is proposed in the Documentation, under Installation, virtuallenv. See [Answer by _geckon_ to question _Using Python 3 in virtualenv_
](http://stackoverflow.com/a/30233408/3493586)

This is a minor change, just 
`python3 -m venv /path/to/new/virtual/environment`
instead of 
`virtualenv venv`

I would be willing to contribute with this if assigned the task.

ThiefMaster:
Go for it, just make sure it's clear how to do it in both Python2 and Python3

jeffwidman:
@runartrollet it looks like you created a commit for this, but never opened a pull request for this. Are you planning to do so?

davidism:
Oh, that was me rewriting the docs some more. This should be addressed now.

PR #2213, opened 2017-03-19
Issue #2212: documentation: virtualenv and python3

runartrollet:
Please note that for clarity, I changed for folder name of the created virtual environment from `venv` to `virtenv`, seeing the relevant python3-command uses the argument `venv`.


davidism:
Not sure why there's a 👎 on this, seems like a good idea to point this out. I'll make a few cleanup changes first.

davidism:
I think I'd like to assume Python 3 and make notes about Python 2, instead of the reverse.

Issue #2221, opened 2017-03-23
Flask treats SystemExit and other BaseExceptions as indicating success


Diggsey:
See here:
https://github.com/pallets/flask/blob/master/flask/app.py#L1991

When combined with a database layer, this completely breaks transaction guarantees, as the request context "pop" handler will see no error and assume the transaction was successful, resulting in it being committed while in an inconsistent state.

ThiefMaster:
https://github.com/pallets/flask/blob/master/flask/app.py#L1528

I don't see why you would have database logic running outside your flask app (wsgi middleware?) but if you tell Flask to propagate exceptions they'll be re-raised.

ThiefMaster:
Also, see https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L846-L850 - in teardown_appcontext you can check whether there has been an exception or not and commit/rollback accordingly.

The request context callbacks are not the right place for this - a database transaction usually does not need an actual HTTP request.

untitaker:
I'm not sure I understand. Why does your db driver raise `SystemExit`?

Diggsey:
@ThiefMaster `handle_exception` is not called on `BaseException`s so is not useful. Yes, we use retry middleware to automatically retry transactions which are rolled back due to deadlock. Regardless of whether you consider this good practice, flask should not be treating `BaseException`s as though no error occurred.

@untitaker My DB driver doesn't raise `SystemExit`, it is raised externally whenever the orchestration system (in this case kubernetes) decides to shutdown a pod. This should be a clean shutdown, and should not be corrupting the database...

ThiefMaster:
Oh now I see, you want this code to handle non-Exception-subclass exceptions too...

ThiefMaster:
Even though I'm still not sure why you would ever get a SystemExit from application code - where does Kubernetes inject the exception?

Diggsey:
It may be `KeyboardInterrupt` rather than `SystemExit`, I'm not sure. Since flask quits and none of our error handlers trigger we don't get any log output! (This was a huge PITA to debug).

ThiefMaster:
That sounds like you are using the `app.run()` / `flask run` dev server. I'd expect proper wsgi servers to finish processing a request and then shutting down cleanly instead of sending KeyboardInterrupt when shut down.

untitaker:
[...]

PR #2222, opened 2017-03-23
Handle BaseExceptions

Diggsey:
This ensures that `auto_pop` is called with a non-`None` value in the case of a BaseException being thrown.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1991,6 +1991,9 @@ class Flask(_PackageBoundObject):
             except Exception as e:
                 error = e
                 response = self.handle_exception(e)
+            except:
+                error = sys.exc_info()[1]
+                raise
             return response(environ, start_response)
         finally:
             if self.should_ignore_error(error):
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -791,6 +791,26 @@ def test_error_handling_processing():
         assert resp.data == b'internal server error'
 
 
+def test_baseexception_error_handling():
+    app = flask.Flask(__name__)
+    app.config['LOGGER_HANDLER_POLICY'] = 'never'
+
+    @app.route('/')
+    def broken_func():
+        raise KeyboardInterrupt()
+
+    with app.test_client() as c:
+        try:
+            c.get('/')
+            raise AssertionError("KeyboardInterrupt should have been raised")
+        except KeyboardInterrupt:
+            pass
+
+        ctx = flask._request_ctx_stack.top
+        assert ctx.preserved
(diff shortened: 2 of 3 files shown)


Diggsey:
Fixes #2221

flask/app.py
```diff
@@ -1991,6 +1991,9 @@ def wsgi_app(self, environ, start_response):
             except Exception as e:
                 error = e
                 response = self.handle_exception(e)
+            except:
```
untitaker: Instead of `except:` and using `sys.exc_info` you could use `except BaseException as e`

Diggsey: In python2 it's possible to raise exceptions that don't extend even `BaseException`, using `except` guarantees everything will be caught.

Diggsey:
@untitaker Done

tests/test_basic.py
```diff
@@ -791,6 +791,26 @@ def after_request(resp):
...
+def test_baseexception_error_handling():
+    app = flask.Flask(__name__)
+    app.config['LOGGER_HANDLER_POLICY'] = 'never'
+
+    @app.route('/')
+    def broken_func():
+        raise KeyboardInterrupt()
+
+    with app.test_client() as c:
+        try:
```
JordanP:
```python 
with pytest.raises(KeyboardInterrupt)
    c.get('/')
```
would be more pytest-idiomatic.

Diggsey: Fixed

untitaker:
I'm still unconvinced that this is a good idea. As said catching `KeyboardInterrupt` may turn out to be problematic and I'd still like to know which exception your code actually raised.

Diggsey:
@untitaker I don't follow. The exception is being reraised. Functionally, there is zero difference between the `finally` block, which already existed, vs catching the exception and reraising it.

untitaker:
Sorry, yes. I was confused.

untitaker:
@ThiefMaster @davidism @RonnyPfannschmidt please review

Diggsey:
What sort of time-scale should I expect on this? If it's going to be a while we'll just fork flask...

untitaker:
I need reviews from other people. In your position I would try to work around
this in your own code or, yes, fork Flask.

untitaker:
@Diggsey however, Flask will not see a new release for a while, so I recommend either running from master or, again, forking. I don't think this is eligible for a bugfix release, but I might be wrong.

mitsuhiko:
We can put this into a bugfix release.

untitaker:
I'll do it then

untitaker:
0.12.1 is out

PR #2223, opened 2017-03-23
Fix send_file's attachment_filename to work with non-ascii filenames

### LINKED ISSUE #1286: send_file fails when filename contains unicode symbols

antlarr:
This commit implements https://tools.ietf.org/html/rfc2231#section-4 in
order to support sending unicode characters. Tested on both Firefox and
Chromium under Linux.

~~This adds unidecode as a dependency, which might be relaxed by using
.encode('latin-1', 'ignore') but wouldn't be as useful.~~ This pull request no longer adds any new dependency.

Also, added a test for the correct headers to be added.

Previously, using a filename parameter to send_file with unicode characters, it
failed with the next error since HTTP headers don't allow non latin-1 characters.
Error on request:
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 193, in run_wsgi
    execute(self.server.app)
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 186, in execute
    write(b'')
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 152, in write
    self.send_header(key, value)
  File "/usr/lib64/python3.6/http/server.py", line 509, in send_header
    ("%s: %s\r\n" % (keyword, value)).encode('latin-1', 'strict'))
UnicodeEncodeError: 'latin-1' codec can't encode character '\uff0f' in position 58: ordinal not in range(256)

Fixes #1286

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -14,6 +14,7 @@ import sys
 import pkgutil
 import posixpath
 import mimetypes
+import unicodedata
 from time import time
 from zlib import adler32
 from threading import RLock
@@ -41,7 +42,6 @@ from .signals import message_flashed
 from .globals import session, _request_ctx_stack, _app_ctx_stack, \
      current_app, request
 from ._compat import string_types, text_type
-from unidecode import unidecode
 
 
 # sentinel
@@ -536,7 +536,9 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
             raise TypeError('filename unavailable, required for '
                             'sending as attachment')
         filename_dict = {
-            'filename': unidecode(text_type(attachment_filename)),
+            'filename': (unicodedata.normalize('NFKD',
+                         text_type(attachment_filename)).encode('ascii',
+                                                                'ignore')),
             'filename*': "UTF-8''%s" % url_quote(attachment_filename)}
         headers.add('Content-Disposition', 'attachment',
                     **filename_dict)
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -14,7 +14,7 @@ import sys
 import pkgutil
 import posixpath
 import mimetypes
-import unicodedata
+from unicodedata import normalize
 from time import time
 from zlib import adler32
 from threading import RLock
@@ -535,13 +535,22 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         if attachment_filename is None:
             raise TypeError('filename unavailable, required for '
                             'sending as attachment')
-        filename_dict = {
-            'filename': (unicodedata.normalize('NFKD',
-                         text_type(attachment_filename)).encode('ascii',
-                                                                'ignore')),
-            'filename*': "UTF-8''%s" % url_quote(attachment_filename)}
+        normalized = normalize('NFKD', text_type(attachment_filename))
+
+        try:
+            normalized.encode('ascii')
+        except UnicodeEncodeError:
+            filenames = {
+                'filename': normalized.encode('ascii', 'ignore'),
+                'filename*': "UTF-8''%s" % url_quote(attachment_filename),
+            }
+        else:
+            filenames = {
+                'filename': attachment_filename,
+            }
+
         headers.add('Content-Disposition', 'attachment',
-                    **filename_dict)
+                    **filenames)
 
     if current_app.use_x_sendfile and filename:
         if file is not None:
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -41,6 +41,7 @@ from .signals import message_flashed
 from .globals import session, _request_ctx_stack, _app_ctx_stack, \
      current_app, request
 from ._compat import string_types, text_type
+from unidecode import unidecode
 
 
 # sentinel
@@ -534,8 +535,11 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         if attachment_filename is None:
             raise TypeError('filename unavailable, required for '
                             'sending as attachment')
+        filename_dict = {
+            'filename': unidecode(attachment_filename),
+            'filename*': "UTF-8''%s" % url_quote(attachment_filename)}
         headers.add('Content-Disposition', 'attachment',
-                    filename=attachment_filename)
+                    **filename_dict)
 
     if current_app.use_x_sendfile and filename:
         if file is not None:
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -14,10 +14,10 @@ import sys
 import pkgutil
 import posixpath
 import mimetypes
-from unicodedata import normalize
 from time import time
 from zlib import adler32
 from threading import RLock
+import unicodedata
 from werkzeug.routing import BuildError
 from functools import update_wrapper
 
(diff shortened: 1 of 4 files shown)


antlarr:
Btw, this would also fix https://github.com/pallets/werkzeug/issues/742

aaossa:
Can this be done without adding a dependency? I think that adding a dependency for just one use case is unnecessary. Maybe using Python 2 u"Unicode strings" works?

antlarr:
I removed the Unidecode dependency as requested. By using unicodedata which is part of the standard library, no functionality is lost (at least, I couldn't find a case where it works differently to Unidecode) and I didn't even have to change the test.

davidism:
See also #2115

antlarr:
#2115 just adds a warning if non-ascii characters are used. This PR adds support to send unicode filenames on send_file, so I think #2115 can be just closed once this one is accepted (assuming it will be accepted :) )

davidism:
I know it's just a warning, I was just linking the two PRs. I think it should still warn.

davidism:
This PR seems OK, but I would like to know how this header is handled by older browsers. Don't really care about IE6 old, but don't want to break anything that might reasonably be in use.

flask/helpers.py
```diff
@@ -534,8 +535,13 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         if attachment_filename is None:
             raise TypeError('filename unavailable, required for '
                             'sending as attachment')
+        filename_dict = {
+            'filename': (unicodedata.normalize('NFKD',
+                         text_type(attachment_filename)).encode('ascii',
+                                                                'ignore')),
+            'filename*': "UTF-8''%s" % url_quote(attachment_filename)}
```
davidism: We need to ensure that the order here is retained, since IE8 breaks if `filename*` is first in the header. Since we're unpacking the dict, we can't ensure order. One solution is to patch Werkzeug to allow `*args` to `headers.add`.

davidism: If we don't care about supporting IE8, we can ignore this. @untitaker @ThiefMaster thoughts?

davidism: For reference, DISA (US mil) requires IE10 or 11 to be certified, so even big gov deployments can't be on a network without that. I feel fairly confident in not supporting IE8.

flask/helpers.py
```diff
@@ -534,8 +535,13 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         if attachment_filename is None:
             raise TypeError('filename unavailable, required for '
                             'sending as attachment')
+        filename_dict = {
+            'filename': (unicodedata.normalize('NFKD',
```
untitaker: Please put this in a separate variable, since the linewrapping here is awkward.

flask/helpers.py
untitaker: Changing the Headers API to care about ordering of options would require a refactor. I don't think that is warranted for a workaround for older browsers.

We could check if the filename fits into ASCII and omit `filename*` accordingly. Then the user can decide whether they want to support IE8 or not.

antlarr:
Just for the records, the relevant tests in that page you linked (great resource, btw) are https://greenbytes.de/tech/tc2231/#attfnboth and https://greenbytes.de/tech/tc2231/#attfnboth2

davidism:
Moving this out of the review since it was getting a little crowded. Check if the name is only ascii before including the utf8 version.

~~~python
from werkzeug._compat import to_native

normalized = unicodedata.normalize('NFKD', text_type(attachment_filename))

try:
    normalized.encode('ascii')
except UnicodeEncodeError:
    filenames = {
        'filename': normalized.encode('ascii', 'ignore'),
        'filename*': "UTF-8''%s" % url_quote(attachment_filename),
    }
else:
    filenames = {
        'filename': attachment_filename,
    }
~~~

tests/test_helpers.py
```diff
@@ -560,6 +560,18 @@ def test_attachment(self):
             assert options['filename'] == 'index.txt'
             rv.close()
 
+    def test_attachment_with_utf8_filename(self):
```
davidism: Test that `filename*` isn't present here.

tests/test_helpers.py
```diff
@@ -560,6 +560,18 @@ def test_attachment(self):
...
+        app = flask.Flask(__name__)
+        with app.test_request_context():
+            with open(os.path.join(app.root_path, 'static/index.html')) as f:
+                rv = flask.send_file(f, as_attachment=True,
+                                     attachment_filename=u'Ñandú／pingüino.txt')
+                content_disposition = set(rv.headers['Content-Disposition'].split(';'))
```
davidism: Should use `parse_options_header` like above.

flask/helpers.py
davidism: Just realized that even if we enhance `Headers.add`, we still can't use it because we can't guarantee that projects are using the latest Werkzeug version. 😞

davidism:
Looks like `_option_header_piece_re.match(value)` in `parse_options_header` with `filename*=UTF-8''...` results in the option name "filename", which is already present for the ASCII filename. So only one option gets returned, instead of two. However, not sure if there's a reason it omits the `*`, so not sure if we should change the re to parse `filename*` differently. @mitsuhiko @untitaker

davidism:
Then there's an issue that recwarn picks up an unclosed resource warning, even though the test is calling `rv.close()`. Not sure what's going on there.

tests/test_helpers.py
```diff
@@ -560,6 +561,24 @@ def test_attachment(self):
...
+                'static/index.html', as_attachment=True,
+                attachment_filename=u'Ñandú／pingüino.txt'
+            )
+            value, options = parse_options_header(
+                rv.headers['Content-Disposition']
+            )
+            rv.close()
+
+        assert value == 'attachment'
+        assert sorted(options.keys()) == ('filename', 'filename*')
```
untitaker: You could just use `options == { ... }` here.

untitaker:
I don't see a resource warning, but that's secondary. `filename*` doesn't exist because `parse_options_header` has properly decoded it (use `tox -epy -- --pdb` to look at options, see that it is decoded as UTF-8).

If you want to make an assertion that both or only one of `filename/filename*` is given I guess you'll have to use string operations (`assert "filename*" in rv.headers['Content-Disposition']` e.g.)

davidism:
Then again, that parsing relies on the order of the options. ;-) It could just as easily have seen the ASCII version second.

tests/test_helpers.py
```diff
@@ -560,6 +561,24 @@ def test_attachment(self):
...
+        assert options['filename'] == 'Nandu/pinguino.txt'
+        assert options['filename*'] == 'UTF-8''%C3%91and%C3%BA%EF%BC%8Fping%C3%BCino.txt'
```
ThiefMaster: What's the point of the `''` there? `'foo''bar'` and `'foobar'` are identical

davidism: That's my bad, the outer quotes should be `"`. Then again, this part of the test already fails because the option key is 'filename' at this point.

davidism:
Based on #932, it looks like a browser sending the `Content-Disposition` header only sends one value for the key, so I'll drop the parsing question until an actual issue comes up.

PR #2228, opened 2017-04-02
Add blueprint and the app factory pattern into the flaskr example

wgwz:


ThiefMaster:
AFAIK flaskr is supposed to be a simple example - I think it'd make much more sense to have a separate example for a bigger project with packages, blueprints, etc.

wgwz:
Just thought I'd put it up for discussion. I recall @untitaker wanting to improve things a bit.

untitaker:
I am torn on this. If we update Flaskr in the repo the tutorial would have to
explain the added details as well. OTOH I'm unsure if a separate example is as
educational.

I'd still go with a separate example for now.

wgwz:
I'm open to adding it as a separate example, if that's the consensus. I'm also willing to add some documentation into the flaskr tutorial. Maybe a new section called: "Refactoring flaskr for a larger app". so we would have `examples/flaskr/` and `examples/refactored-flaskr` and this new section could explain how to go from a to b.

untitaker:
That sounds like a good idea.

wgwz:
@untitaker Do you want me to put the doc changes in this PR or file separately?

untitaker:
Please in the same PR

davidism:
The wording is fine. It says *could* be a file named that, because there's no standard. The most common is probably "wsgi.py".

I'd like to just improve `FLASK_APP` so it can contain `package.sub:create_app('dev')`, like Gunicorn and uWSGI can do. Then we don't need the file at all.

greyli:
I agree with the last solution, it's much more simple and elegant. Will this feature be included in the next release (0.13)?

kennethreitz:
I kind of like this example, actually, as is. Blueprints should be simple and approachable for small applications.

PR #2234, opened 2017-04-07
Add example for virtualenv integration in cli docs

greyli:


untitaker:
Editing `activate` scripts is not a good idea, it is like editing code in `site-packages`.

greyli:
Thanks. This part [Virtualenv Integration](http://flask.pocoo.org/docs/0.11/cli/#virtualenv-integration) was already in docs, is it better to remove it?

untitaker:
I see, I wasn't aware of that section.

docs/cli.rst
```diff
@@ -56,6 +56,18 @@ If you are constantly working with a virtualenv you can also put the
 bottom of the file.  That way every time you activate your virtualenv you
 automatically also activate the correct application name.
 
+You need to edit the specific activate script for different shell, for example:
```
untitaker: This is not related to the shell.

davidism: There are different scripts depending on the shell you're using. This sentence is awkward though, should read "Edit the activate script for the shell you use. For example:"

greyli: There are different activate script in there (e.g. activate, activate.csh, activate.fish...) for different shell.

greyli: Edited.

untitaker: Right, I was under the impression that `activate` was POSIX shell compatible.

docs/cli.rst
```diff
@@ -56,6 +56,18 @@ If you are constantly working with a virtualenv you can also put the
 bottom of the file.  That way every time you activate your virtualenv you
 automatically also activate the correct application name.
 
+Edit the activate script for the shell you use, for example:
+
+Unix Bash: ``venv/bin/activate``::
+
+    FLASK_APP=foo.py
```
davidism: I'd rather the examples encourage the use of installed package names rather than file names. `foo` instead of `foo.py`. @untitaker thoughts?

untitaker: Agreed. Perhaps replace with `...` or `<see above>`?

davidism: Or `hello`, since that's what's used in the other examples.

greyli: How about `<your app>`?

untitaker: Yes, all good

docs/cli.rst
```diff
@@ -56,6 +56,18 @@ If you are constantly working with a virtualenv you can also put the
 bottom of the file.  That way every time you activate your virtualenv you
 automatically also activate the correct application name.
 
+Edit the activate script for the shell you use. For example:
+
+Unix Bash: ``venv/bin/activate``::
+
+    FLASK_APP=hello
+    export FLASK_APP
```
davidism: Should just be `export FLASK_APP=hello`, no need to split the lines in this case (and most cases).

greyli: The generated code above was in this form, is it better to keep it?

davidism: Yeah, it's fine either way.

untitaker:
Thanks @greyli 🎉

PR #2238, opened 2017-04-12
filename can be latin-1, not just ascii

davidism:
* The filename header, being a header, can include Latin-1 characters, not just ASCII.
* Don't normalize the unicode name until we know the UTF-8 version needs to be sent. Otherwise you create situations where the name *could* be sent as just the single field, but the normalized form doesn't encode to Latin-1.

ref #2223

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -484,7 +484,7 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
         
     .. _RFC 2231: https://tools.ietf.org/html/rfc2231#section-4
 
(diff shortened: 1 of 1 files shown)


untitaker:
Uhh, did you test this with IE? The reason we keep this in ASCII is because of
allowing IE support...

davidism:
The previous version didn't coerce to ASCII and was documented as Latin-1. https://github.com/davidism/flask/blame/6efea346dd54fa9adb65fc044d72e5ec8bdf4018/flask/helpers.py#L538

untitaker:
That's fair, I missed that.

PR #2240, opened 2017-04-13
update send_file cache control timeout=-1 no-cache

mhohai:
update send_file cache control
set timeout <= 0 will no-cache no-store


flask/helpers.py
```diff
@@ -580,9 +581,13 @@ def send_file(filename_or_fp, mimetype=None, as_attachment=False,
     rv.cache_control.public = True
     if cache_timeout is None:
         cache_timeout = current_app.get_send_file_max_age(filename)
-    if cache_timeout is not None:
+    if cache_timeout is not None and cache_timeout > 0:
         rv.cache_control.max_age = cache_timeout
         rv.expires = int(time() + cache_timeout)
+    else:
```
davidism: I don't think this should happen if `cache_timeout` is `None`, I would interpret that as "don't do anything special with the cache headers."

untitaker:
This effectively becomes the default behavior, which I consider wrong. `send_file` is also used for static assets. Also note that `private` is already the default value.

flask/helpers.py
untitaker: I probably agree, but am unsure what the other parameters would actually change if the timeout is already `0`.

untitaker:
Any thoughts on my previous comment? How does browser's behavior change with this PR?

untitaker:
I would hold off of this before we know what this actually accomplishes.

lepture:
It would benefit the development. Chrome is caching way too much now. With no-cache control header, developers can easily refresh and load the new css/js code.

untitaker:
True, but this affects more than dev.

On 12 May 2017 04:33:09 GMT+02:00, Hsiaoming Yang <notifications@github.com> wrote:
>It would benefit the development. Chrome is caching way too much now.
>With no-cache control header, developers can easily refresh and load
>the new css/js code.
>
>-- 
>You are receiving this because your review was requested.
>https://github.com/pallets/flask/pull/2240#issuecomment-300967948

-- 

kennethreitz:
So, are we going to merge this or toss it?

davidism:
I liked the idea at first, but after researching the headers more it's unclear.

* `no-cache` means always ask for the headers even if the cached resource isn't expired
* `no-store` means don't cache anything anything, always get the remote resource
* `must-revalidate` means something like `no-cache`, but is more strict?

Why do we want `-1` to mean `no-store` as opposed to only `no-cache`, which seems more efficient while still accomplishing "don't use stale static files"? Why do we remove the `public` flag, which I think is the default anyway? I think this probably belongs in a function in your own code, rather than as a magic number for this parameter.

~~~python
def send_file_no_cache(*args, **kwargs):
    rv = send_file(*args, **kwargs)
    rv.cache_control.no_cache = True
    rv.cache_control.no_store = True
    return rv
~~~

The docs I found aren't completely clear about what combinations do.

* https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

If you can clarify more and think this should be merged, please reopen discussion.

PR #2242, opened 2017-04-13
get mtime in utc

davidism:
Test `test_helpers.TestSendfile.test_send_file_range_request` was failing on my machine but not on Travis. Turned out it was sending the mtime of the file as local time, but comparing to a UTC time. Use `utcfromtimestamp` instead of `fromtimestamp`.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -517,7 +517,7 @@ class TestSendfile(object):
         assert rv.status_code == 416
         rv.close()
 
-        last_modified = datetime.datetime.fromtimestamp(os.path.getmtime(
+        last_modified = datetime.datetime.utcfromtimestamp(os.path.getmtime(
             os.path.join(app.root_path, 'static/index.html'))).replace(
             microsecond=0)
 


untitaker:
We should add a test matrix entry for this.

Also I think this should've been merged against the maintenance branch with changelog entry.

davidism:
I'll cherry pick it to maintenance, wasn't paying attention. Do we add changelogs for test fixes?. What would the matrix look like, since this is just making sure the timestamp is in UTC? We basically have to assume the timestamp is in UTC since there's no timezone information in the header.

untitaker:
I didn't realize that this was just a testfix. I do think it still could be cherrypicked into maintenance, but changelog is not necessary IMO

untitaker:
Regarding matrix entry, you mentioned that the tests passed either way on Travis. Not sure if the gain is big enough to warrant another matrix entry now, since this is just a testfix. Another option would be to create a test that emulates different timezones by monkeypatching the `datetime` module (freezegun could do this I think).

davidism:
Technically, all the code still works as expected, the conditional check will just have a window where the times don't line up if the client sends a timestamp in localtime. Since it's just a timestamp, there's no way to know that it's not in UTC.

untitaker:
Right. Nevermind then.

davidism:
Cherry-picked to 0.12-maintenance: cfd3e50ab6447a69019fc008c260f8c078c4ac7d

davidism:
Cherry-picked to 0.12-maintenance: 4ff84d537aa386fde36182ed797a79e3b582be75

PR #2244, opened 2017-04-14
Remove empty XML tag.

antosha417:


ThiefMaster:
No offense, but did you just really go over the XML source of a SVG file with something that is not a graphics program? ;)

davidism:
Going to just leave this. It isn't hurting anything to have a metadata tag in there. Thanks for your interest, though!

PR #2249, opened 2017-04-21
Continue #1489: Add kwarg to disable auto OPTIONS on add_url_rule 

davidism:
Merge master into #1489 and do some cleanup.


untitaker:
I'm really curious why the original option was implemented as attribute on the
view function. Perhaps there is a reason it is not part of `add_url_rule`'s
signature.

davidism:
Yeah, I found that a little weird too, but this doesn't seem to hurt anything.

untitaker:
cc @mitsuhiko?

davidism:
The original issue #1488 explains a little more: they're registering more routes by endpoint name, without providing the view function each time. So they can't set the attribute because there's nothing to set it on.

untitaker:
That's fair, though I wonder if we should deprecate the other way to do it.

davidism:
I'll open an issue separate from this.

PR #2092, opened 2016-11-20
Add routes command to flask cli

willsoto:
Running `flask routes` will list all registered routes for a given app, sorted by endpoint

Inspired by [`mix phoenix.routes`](http://www.phoenixframework.org/docs/routing), this will display a list of all registered routes in an application. This is basically #1446, but I have updated it and made it "dumber". 

NB: I know that the other PR is old and there doesn't seem to be much enthusiasm for it, but I do think this is useful to have to get a high level overview of application routes, especially if they are being generated programmatically. 

Example output:

```sh
$ flask routes
api.me                    GET              /api/users/me/
api.users                 GET              /api/users/
                          POST             /api/users/
                          PUT, DELETE, GET /api/users/<pk>
api.users_schema          GET              /api/users/schema/
auth_blueprint.authorized GET              /login/authorized
auth_blueprint.do_login   GET              /do-login
auth_blueprint.login      GET              /login
auth_blueprint.logout     GET              /logout
auth_blueprint.lr_test    GET              /lr
static                    GET              /static/<path:filename>
```

PR #2259, opened 2017-04-26
Add routes command

davidism:
Based on #1446, with some ideas from #2092. The whole command is basically reworked, but the most important change is that there is a `match` sort order to show the order Flask will dispatch in (which is really just not sorting).


davidism:
One simple change that I didn't end up doing was sorting by the length of the methods before  sorting alphabetically, so that single method routes show up before multi-method ones. Wasn't sure if that would be more useful than having *anything* with `GET` sort first.

PR #2262, opened 2017-04-28
Py36 json support added

### LINKED ISSUE #2243: Take advantage of Py3.6 json support for bytes
[Python 3.6 added support for loading json from str or bytes.](https://docs.python.org/3.6/library/json.html#json.loads) It handles detection of the three encodings supported by JSON. Would be nice to take advantage of this if the user is using 3.6+ instead of our basic "assume it's UTF-8" approach.

Would have to make sure simplejson also allows this (I think it does).

Then again, I'm not sure there have been any reports of Flask decoding JSON incorrectly. Maybe no real clients are encoding with anything besides UTF-8?

akohli96:
Resolves #2243 

1. Added variable in _config.py to check if system is 3.6+
2. Added logic in json.py to check if system is 3.6+ to simply decode json 

EDIT : 3.6 to 3.6+

### DIFF
--- a/flask/json.py
+++ b/flask/json.py
@@ -22,6 +22,8 @@ from jinja2 import Markup
 # depend anyways.
 from itsdangerous import json as _json
 
+#Check if itsdangerous json package is simplejson or standard json
+SIMPLEJSON = _json.__name__ == 'simplejson'
 
 # Figure out if simplejson escapes slashes.  This behavior was changed
 # from one version to another without reason.
@@ -154,7 +156,7 @@ def loads(s, **kwargs):
     application on the stack.
     """
     _load_arg_defaults(kwargs)
-    if not PY36PLUS:
+    if not (PY36PLUS and SIMPLEJSON): #If  python 3.6+ and simple json skip
         if isinstance(s, bytes):
             s = s.decode(kwargs.pop('encoding', None) or 'utf-8')
     return _json.loads(s, **kwargs)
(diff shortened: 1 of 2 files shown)


flask/json.py
```diff
@@ -154,8 +154,9 @@ def loads(s, **kwargs):
     application on the stack.
     """
     _load_arg_defaults(kwargs)
-    if isinstance(s, bytes):
-        s = s.decode(kwargs.pop('encoding', None) or 'utf-8')
+    if not PY36:
```
ThiefMaster: This doesn't cover 3.7, ...

ThiefMaster: `if sys.version_info[:2] < (3, 6)` would work

akohli96: Great catch @ThiefMaster.
Fixing it right now!

akohli96: I have kept it naming style consistent with PY2.

akohli96:
So, its failing on REQUIREMENTS with no simplejson (devel and release).

davidism:
This no longer does anything about the issue. It only skips decoding for simplejson, while the point was to skip decoding for the built in JSON. It looks like we just can't do this if we want to support non-standard charsets.

akohli96:
[ The input encoding should be UTF-8, UTF-16 or UTF-32](https://docs.python.org/3.6/library/json.html#json.loads) for 3.6 json.loads to work

PR #2275, opened 2017-05-09
Fix import order to accommodate pocoo style

pigletfly:
Fix some import order

### DIFF
--- a/flask/config.py
+++ b/flask/config.py
@@ -14,6 +14,7 @@ import types
 import errno
 
 from werkzeug.utils import import_string
+
 from ._compat import string_types, iteritems
 from . import json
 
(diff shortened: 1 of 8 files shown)


davidism:
Thank you, but we don't accept style PRs.

pigletfly:
Some imports follow PEP 8 style , but some are not.I think there should be a coherent import style,PEP 8 style or [Pocoo Styleguide](http://flask.pocoo.org/docs/0.12/styleguide/#general-layout) style.

davidism:
https://github.com/pallets/flask/pull/1702#issuecomment-294037694

pigletfly:
got it

PR #2288, opened 2017-05-20
Set "Vary: Cookie" header when session is accessed

davidism:
Based on [Django's behavior](https://github.com/django/django/blob/1ce04bcce0076360623ae164afd3541a5c031af2/django/contrib/sessions/middleware.py#L45), order is delete cookie if empty, append vary header if accessed, set cookie if modified.

continues #1026


ThiefMaster:
`setdefault()` also needs to count as an access (it only sets modified if the key wasn't already in the session)

ThiefMaster:
What happens if someone already sets a vary header manually?

davidism:
@ThiefMaster good point, we need to check if the header exists already, reparse it, and add cookie if it's not there. Not sure where this belongs, probably in Werkzeug, but duplicated here for a while to not require bumping the minimum Werkzeug version?

jeffwidman:
@davidism unless you're able to get to this soon, do you mind filing an issue so we don't lose track of this?

PR #2291, opened 2017-05-21
log a clickable url to home page when listening 0.0.0.0, more beginne…

pengwk:
Hello, I'm a novice.  I use pycharm for development, when I ran my application to listen on 0.0.0.0, I needed to type “http://localhost:5000" to webbroswer to visit my website. It's a bit inconvenient! So I add a some codes to the app.py file to render the url.

```
 * Restarting with stat
 * You can visit http://localhost:5000/
 * Debugger is active!
 * Debugger pin code: 294-762-868
 * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
```

 In my opinion, it's convenient than before. 

Thanks you very much!

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -858,6 +858,9 @@ class Flask(_PackageBoundObject):
         options.setdefault('use_reloader', self.debug)
         options.setdefault('use_debugger', self.debug)
         try:
+            if self.debug is True and host == '0.0.0.0':
+                from werkzeug._internal import _log
+                _log('info', ' * You can visit %s://%s:%d/', "http", 'localhost', port)
             run_simple(host, port, self, **options)
         finally:
             # reset the first request information if the development server


untitaker:
On which operating system is this? I can type `0.0.0.0:5000` for localhost on Linux.

ThiefMaster:
I guess it doesn't make much sense to use `http://0.0.0.0` to access anything in the browser. OTOH, assuming you are either using a VM/container to run it or a separate machine (otherwise 127.0.0.1 would work fine if it's just for you) then why not just bind to that machine's IP address?

pengwk:
Thanks! I used Windows10. It doesn't support it! @untitaker

pengwk:
@ThiefMaster  I must bind to 0.0.0.0 due to I need to run my app in two LANs(WIFI and Ethernet)  when I develop. 

What I want is that  I can click the url to visit my site, when I firstly start my app in Pycharm. A bit convenient.
image

Thank you!

davidism:
This belongs in Werkzeug if anything, but I don't find it particularly useful. You should not be binding to the outside with the debug server. I'm not on board with things that make that easier.

PR #2295, opened 2017-05-22
Add coverage for Blueprint.app_errorhandler

dawran6:
This test case registers an application-wise error handler from
a Blueprint. Verifies the error handler by aborting the flask app
from the application itself as well as from another registered
Blueprint.


dawran6:
Please help me to review this PR. Any comment is welcome. Thanks!

Issue #2265, opened 2017-05-01
Detect `create_app` when FLASK_APP doesn't specify app object


davidism:
`find_best_app` only looks for `app` and `application`. If it can't find those but finds the name `create_app`, it should try to call that with no arguments to get the Flask app.

If we modify `ScriptInfo.load_app` to pass itself to `locate_app` and `find_best_app`, we can make this consistent with #2266.

greyli:
In this case, how to use `@app.cli.command`?  Since blueprint doesn't support `cli.command`, we need to import `app` from the place where the application instance was created (e.g. `wsgi.py`) .

davidism:
`app.cli.add_command` in the factory, just like any of the other non-decorator methods to register things.

greyli:
I have many commands, now I put it in a separate file named `commands.py` (without factory). Is it possible to make it work outside the factory? By the way, what's the best place for the factory, `app/__init__.py` or `app/application.py`? Thanks!

PR #2297, opened 2017-05-22
Auto-detect create_app() in find_best_app() #2265

HndrkMkt:
Resubmit for #2265 (git issues)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -40,6 +40,8 @@ Major release, unreleased
 - Allow IP address as exact session cookie domain. (`#2282`_)
 - ``SESSION_COOKIE_DOMAIN`` is set if it is detected through ``SERVER_NAME``.
   (`#2282`_)
+- Auto-detect 0-argument app factory called ``create_app`` or ``make_app`` from
+  ``FLASK_APP``. (`#2297`_)
 
 .. _#1489: https://github.com/pallets/flask/pull/1489
 .. _#1898: https://github.com/pallets/flask/pull/1898
@@ -50,6 +52,7 @@ Major release, unreleased
 .. _#2256: https://github.com/pallets/flask/pull/2256
 .. _#2259: https://github.com/pallets/flask/pull/2259
 .. _#2282: https://github.com/pallets/flask/pull/2282
+.. _#2297: https://github.com/pallets/flask/pull/2297
 
 Version 0.12.2
 --------------
--- a/CHANGES
+++ b/CHANGES
@@ -40,8 +40,8 @@ Major release, unreleased
 - Allow IP address as exact session cookie domain. (`#2282`_)
 - ``SESSION_COOKIE_DOMAIN`` is set if it is detected through ``SERVER_NAME``.
   (`#2282`_)
-- Auto-detect 0-argument app factory called ``create_app`` or ``make_app`` from
-  ``FLASK_APP``. (`#2297`_)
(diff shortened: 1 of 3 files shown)


CHANGES
```diff
@@ -40,6 +40,8 @@ Major release, unreleased
 - Allow IP address as exact session cookie domain. (`#2282`_)
 - ``SESSION_COOKIE_DOMAIN`` is set if it is detected through ``SERVER_NAME``.
   (`#2282`_)
+- Auto-detect 0-argument app factory called ``create_app`` or ``make_app`` from
```
ThiefMaster: "zero-argument" sounds better IMO

flask/cli.py
```diff
@@ -37,19 +37,46 @@ def find_best_app(module):
...
     for attr_name in 'app', 'application':
         app = getattr(module, attr_name, None)
-        if app is not None and isinstance(app, Flask):
+        if isinstance(app, Flask):
             return app
 
     # Otherwise find the only object that is a Flask instance.
-    matches = [v for k, v in iteritems(module.__dict__)
-               if isinstance(v, Flask)]
+    matches = [
```
ThiefMaster: what's the reason for the reformatting? IMO the previous one was more readable

flask/cli.py
```diff
@@ -37,19 +37,46 @@ def find_best_app(module):
...
-                         'using a factory function.' % module.__name__)
+    elif len(matches) > 1:
+        raise NoAppException(
+            'Auto-detected multiple Flask applications in module "{module}".'
+            ' Use "FLASK_APP={module}:name" to specify the correct'
+            ' one.'.format(module=module.__name__)
+        )
+
+    # Search for app factory callables.
+    for attr_name in 'create_app', 'make_app':
```
ThiefMaster: I'd put `()` around the tuple, IMO it looks quite strange without here

flask/cli.py
```diff
@@ -37,19 +37,46 @@ def find_best_app(module):
...
-                         'you wrapped it in a WSGI middleware or you are '
-                         'using a factory function.' % module.__name__)
+    elif len(matches) > 1:
+        raise NoAppException(
+            'Auto-detected multiple Flask applications in module "{module}".'
+            ' Use "FLASK_APP={module}:name" to specify the correct'
+            ' one.'.format(module=module.__name__)
+        )
+
+    # Search for app factory callables.
```
ThiefMaster: I think that's obvious :)

flask/cli.py
```diff
@@ -37,19 +37,46 @@ def find_best_app(module):
...
+    for attr_name in 'create_app', 'make_app':
+        app_factory = getattr(module, attr_name, None)
+
+        if callable(app_factory):
+            try:
+                app = app_factory()
+
```
ThiefMaster: not needed

HndrkMkt: The intention is that the factory function could be built in a way that it requires positional arguments (compare http://flask.pocoo.org/docs/0.12/patterns/appfactories/) and thus the call could fail with a TypeError. I'd rather throw a custom exception in that case.

ThiefMaster: i meant the empty line ;)

flask/cli.py
davidism: I don't like the "shove everything to the right" indent style.

ThiefMaster: personally i'd just put all in one line and screw the dumb 79-char  line length :p

PR #2303, opened 2017-05-22
Added tests for make_response and get_debug_flag to improve coverage of helpers.py

MikeTheReader:
Got less coverage improvement than I was hoping, but it did get a bit more.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -22,6 +22,7 @@ from werkzeug.exceptions import BadRequest, NotFound
 from werkzeug.http import parse_cache_control_header, parse_options_header
 from werkzeug.http import http_date
 from flask._compat import StringIO, text_type
+from flask.helpers import get_debug_flag, make_response
 
 
 def has_encoding(name):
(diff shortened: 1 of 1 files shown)


tests/test_helpers.py
```diff
@@ -941,3 +942,52 @@ def test_safe_join_exceptions(self):
         for args in failing:
             with pytest.raises(NotFound):
                 print(flask.safe_join(*args))
+
+class TestHelpers(object):
+
+    def test_get_debug_flag(self):
+        original_debug_value = os.environ.get('FLASK_DEBUG') or ''
```
davidism: Use pytest's monkeypatch for this

tests/test_helpers.py
```diff
@@ -941,3 +942,52 @@ def test_safe_join_exceptions(self):
...
+        assert get_debug_flag(default=True) == True
+
+        os.environ['FLASK_DEBUG'] = original_debug_value
+
+    def test_make_response_no_args(self):
+        app = flask.Flask(__name__)
+        app.testing = True
+        @app.route('/')
+        def index():
+            return flask.helpers.make_response()
```
davidism: test the response object directly rather than making a view

tests/test_helpers.py
```diff
@@ -941,3 +942,52 @@ def test_safe_join_exceptions(self):
...
+        c = app.test_client()
+        rv = c.get()
+        assert rv
+
+    def test_make_response_with_args(self):
+        app = flask.Flask(__name__)
+        app.testing = True
+        @app.route('/')
+        def index():
+            response = flask.helpers.make_response(flask.render_template_string('Hello World'))
```
davidism: test the response object directly rather than making a view

tests/test_helpers.py
```diff
@@ -941,3 +942,38 @@ def test_safe_join_exceptions(self):
         for args in failing:
             with pytest.raises(NotFound):
                 print(flask.safe_join(*args))
+
+class TestHelpers(object):
+
+    def test_get_debug_flag(self, monkeypatch):
+        monkeypatch.setenv('FLASK_DEBUG', '')
+        assert get_debug_flag() == None
```
jeffwidman: Shouldn't this be `is None`? Check identity, not just equality for `None`

tests/test_helpers.py
```diff
@@ -941,3 +942,38 @@ def test_safe_join_exceptions(self):
...
+        assert get_debug_flag(default=True) == True
```
jeffwidman: I think you can omit the `== True`? Since the assert automatically checks if the statement evaluates to `True`... This is true for a bunch of tests below as well. 

Similarly you could rewrite:

    assert FalseThing == False

as:

    assert not FalseThing

I find the latter cleaner, but I guess this is a matter of personal preference as it's arguably less explicit.

tests/test_helpers.py
```diff
@@ -941,3 +942,38 @@ def test_safe_join_exceptions(self):
         for args in failing:
             with pytest.raises(NotFound):
                 print(flask.safe_join(*args))
+
+class TestHelpers(object):
+
+    def test_get_debug_flag(self, monkeypatch):
```
jeffwidman: Since this function is a bunch of asserts with different values for X in `monkeypatch.setenv('FLASK_DEBUG', X)`, couldn't you DRY this up using `pytest` parameterization?

Basically you can use a decorator to tell pytest "run this test 4 times with these 4 different sets of inputs and here are the 4 expected outputs".

MikeTheReader: Wasn't familiar with that -- thanks for leading me that way.

tests/test_helpers.py
MikeTheReader: With parameterizing this goes away as the validation target changes per test run.

tests/test_helpers.py
MikeTheReader: Good call. I did add some logic around this after parameterizing (not sure if that was overkill in this case).

Issue #941, opened 2014-01-03
Registering a handler for HTTPException has no effect


mattoufoutu:
When registering a handler for `werkzeug.exceptions.HTTPException`, it has no effect when an HTTP error is raised.

Assume the following handler:

``` python
@app.errorhandler(HTTPException)
def http_err_handler(error):
    response = jsonify({
        "success": False, 
        "message": error.name
    })
    response.status_code = error.code
    return response
```

When requesting a page for which no route exists, a JSON response should be returned by the error handler, but instead, the usual `Flask`-generated HTTP error page is returned.

On the other hand, if the error handler is defined to handle a specific error code (by passing the error code to the `app.errorhandler` decorator), the exception is trapped and the JSON message returned.

As `wekzeug.exceptions.HTTPException` is the class raised internally by the `abort()` function, why isn't it possible to create a "catch-all" handler like this? Am I missing something?

Bekt:
#839 doesn't fix the issue OP posted. #839 only handles subclasses of `HTTPException`.

untitaker:
Could you provide a testcase that fails?

Bekt:
```py
import unittest

from werkzeug.exceptions import HTTPException, NotFound

import flask


class ErrorHandlerTest(unittest.TestCase):

    def setUp(self):
        app = flask.Flask(__name__)

        @app.errorhandler(HTTPException)
        def http_exception(e):
            return 'generic', 500

        @app.errorhandler(NotFound)
        def notfound_exception(e):
            return 'not found', 404

        @app.route('/error')
        def error():
            flask.abort(500)

        self.app = app

    def test_notfound_handler(self):
        rv = self.app.test_client().get('/')
        self.assertEqual(rv.status_code, 404)
        self.assertEqual(rv.data, 'not found')

    def test_http_handler(self):
        rv = self.app.test_client().get('/error')
        self.assertEqual(rv.status_code, 500)
        self.assertEqual(rv.data, 'generic')


if __name__ == '__main__':
    unittest.main()
```

```bash
$ venv/bin/python app.py
F.
======================================================================
FAIL: test_http_handler (__main__.ErrorHandlerTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "app.py", line 35, in test_http_handler
    self.assertEqual(rv.data, 'generic')
AssertionError: '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>500 ...' != 'generic'

[...]

Issue #2197, opened 2017-03-03
Allow error handler to pass on decision to subsequent handlers


tuukkamustonen:
I would like to do error handling conditionally, like:

```python
def internal_error(exc: peewee.InternalError) -> Optional[Response]:
    if exc.args[0] == 1213:
        # Deadlock found when trying to get lock; try restarting transaction
        return CustomErrorResponse(...)

def any_exception(exc: Exception) -> Response:
        return CustomErrorResponse(...)

app.register_error_handler(peewee.InternalError, internal_error)
app.register_error_handler(Exception, any_exception)
```

Could it be so that if `None` is returned from an errorhandler, Flask keeps on iterating and calls the next suitable errorhandler?

Lines: https://github.com/pallets/flask/blob/master/flask/app.py#L1521-L1525 Looks like the code has changed somewhat since 0.10.1 that I'm running (in 0.10.1 this would have been 3-line change I think).

Of course I could call `return any_exception(exc)` from `internal_error()` but that kinda adds direct dependency between the two methods and I would like to avoid it (consider the error handlers coming from different modules/classes etc. and maybe there's even third error handler sitting between that would be skipped here).

davidism:
This is similar to how `before_request` handlers can return a response if they want to stop response handling early. I started to implement this, but I've decided to close it.

Error handler behavior is already somewhat complicated due to the lookup by blueprint, app, code, and exception, all over the exception's MRO. The current order is:

1. blueprint, code, mro
2. app, code, mro
3. blueprint, exception, mro
4. app, exception, mro

When you throw in the idea of falling through handlers, it becomes even worse: what order is something handled in if there are handlers spread across multiple maps? For example, if the following handlers exist:

1. blueprint, Child(Parent(Exception)), doesn't return
2. blueprint, Exception, returns
3. app, Parent, returns

Should the next handler for Child be 2 because it is the next closest base in the current map, or should it be 3 because it is the next most specific base. What if 3 didn't return, how would we get back to 2? What if there were more layers of bases interleaved back and forth between maps?

So while it was an interesting idea and it makes sense why you'd expect that, the implementation becomes way too hard to reason about.

davidism:
Here's the patch if anyone's interested:

~~~patch
From 826fe2f8ea25322d3d4831b4950eb1ef19a50f22 Mon Sep 17 00:00:00 2001
From: David Lord <davidism@gmail.com>
Date: Tue, 30 May 2017 12:26:28 -0700
[...]

PR #2304, opened 2017-05-22
Added support for generic HTTPException handlers on app and blueprints

cerickson:
Resolves issue #941, replaces pr #2082 and #2144, code based on #2144, tangentially related to #1383
Error handlers are now returned in order of blueprint:code, app:code, blueprint:HTTPException, app:HTTPException, None

Corresponding tests also added.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -22,6 +22,7 @@ from werkzeug.exceptions import BadRequest, NotFound
 from werkzeug.http import parse_cache_control_header, parse_options_header
 from werkzeug.http import http_date
 from flask._compat import StringIO, text_type
+from flask.helpers import get_debug_flag, make_response
 
 
 def has_encoding(name):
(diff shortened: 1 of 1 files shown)


cerickson:
This might also close #2197

jeffwidman:
@cerickson I don't know how you feel about this, and it might be beyond your git foo, but what do you think about cherry-picking the initial commit from #2144 as the initial commit in your branch and then making your changes on top of that? The final code would be the same, but that way @sumpfgottheit would get credit for the hard work he did in figuring this out initially and taking the time to write the initial set of tests.

flask/app.py
```diff
@@ -1462,28 +1462,31 @@ def url_defaults(self, f):
         return f
 
     def _find_error_handler(self, e):
-        """Finds a registered error handler for the request’s blueprint.
-        Otherwise falls back to the app, returns None if not a suitable
-        handler is found.
+        """Finds a registered error handlers for a request in this order:
```
jeffwidman: This should be written in the imperative:  `Find` rather than `Finds`

flask/app.py
```diff
@@ -1462,28 +1462,31 @@ def url_defaults(self, f):
...
 
-        # try blueprint handlers
-        handler = find_handler(self.error_handler_spec
-                               .get(request.blueprint, {})
-                               .get(code))
-        if handler is not None:
-            return handler
+        # check for any in blueprint or app
+        for name, c in ((request.blueprint, code), (None, code),
+                        (request.blueprint, None), (None,None)):
```
jeffwidman: The tuple `None,None` is missing a space

cerickson:
I'm not sure how to do that in git, but you are more then welcome to merge the PRs somehow or if you can give me a link or a rough outline of the procedure I'd be happy to try it.  We should absolutely give the original PR credit.

jeffwidman:
Why did you close this?

cerickson:
We don't know, something went weird with git. Can't reopen it so resubmitting it. -davidism

PR #2305, opened 2017-05-23
Updating contributing rst

sorech02:


CONTRIBUTING.rst
```diff
@@ -24,12 +24,31 @@ Reporting issues
...
 ==================
 
+
+- `Make sure you have a github account <https://github.com/join>`_
+- `Download and install latest version of git on your
+  machine <https://git-scm.com/downloads>`_
+- `Set up your username in git 
+  <https://help.github.com/articles/setting-your-username-in-git/>`_
+- `Set up your email in git 
+  <https://help.github.com/articles/setting-your-email-in-git/>`_
```
a-wakeel: Do we really need these 4 points here? I mean these are some common things to consider and these are understood in my point of view.
What do you think?

PR #2306, opened 2017-05-23
DRYing up the test suite using pytest fixtures

### LINKED ISSUE #2287: DRY up the test suite using pytest fixtures
The test suite is a bit of a mess as far as how each test sets up an app context, request context, etc.

It'd be nice to DRY this up using pytest fixtures.

Benefits:
1. More readable tests.
2. Tests are easier to write.
3. Tests are easier to refactor.
3. Tests are more robust because it's easier to run a test with multiple parameters.

This is a large project, but easy to tackle incrementally. Perhaps it should be broken into multiple issues, as each section of tests requires some specific fixtures, for example `jsonify()` tests require some specific `json` setup.

Related: #2286

Tafkas:
Using pytest fixtures for test cases. Resolves #2287

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,41 @@ import sys
 import pkgutil
 import pytest
 import textwrap
+from flask import Flask as _Flask
+
+
+class Flask(_Flask):
+    testing = True
+    secret_key = __name__
+
+    def make_response(self, rv):
+        if rv is None:
+            rv = ''
+
+        return super(Flask, self).make_response(rv)
+
+
+@pytest.fixture
+def app():
+    app = Flask(__name__)
+    return app
+
+
+@pytest.yield_fixture
+def app_ctx(app):
+    with app.app_context() as ctx:
+        yield ctx
+
+
+@pytest.yield_fixture
+def req_ctx(app):
+    with app.test_request_context() as ctx:
+        yield ctx
+
+
+@pytest.fixture
+def client(app):
+    return app.test_client()
 
 
 @pytest.fixture
@@ -21,6 +56,7 @@ def test_apps(monkeypatch):
         os.path.abspath(os.path.join(
             os.path.dirname(__file__), 'test_apps'))
     )
+
 
 @pytest.fixture(autouse=True)
 def leak_detector(request):
@@ -62,12 +98,13 @@ def limit_loader(request, monkeypatch):
 
     def get_loader(*args, **kwargs):
         return LimitedLoader(old_get_loader(*args, **kwargs))
+
     monkeypatch.setattr(pkgutil, 'get_loader', get_loader)
 
 
 @pytest.fixture
 def modules_tmpdir(tmpdir, monkeypatch):
-    '''A tmpdir added to sys.path'''
+    """A tmpdir added to sys.path"""
     rv = tmpdir.mkdir('modules_tmpdir')
     monkeypatch.syspath_prepend(str(rv))
     return rv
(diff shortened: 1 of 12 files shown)


tests/test_helpers.py
```diff
@@ -34,240 +34,230 @@ def has_encoding(name):
...
-            assert rv == '"\\u2603"'
+        rv = flask.json.dumps(u'\N{SNOWMAN}')
+        assert rv == '"\\u2603"'
 
         app.config['JSON_AS_ASCII'] = False
-        with app.app_context():
-            rv = flask.json.dumps(u'\N{SNOWMAN}')
-            assert rv == u'"\u2603"'
+        rv = flask.json.dumps(u'\N{SNOWMAN}')
+        assert rv == u'"\u2603"'
```
jeffwidman: Note that your refactoring here runs both tests within a single app_context as the fixture teardown isn't executed until you exist the test function. Versus the previous way used separate app contexts. I don't think it matters in this case, although I'm uncertain as it's been a while since I've looked at what `app_context()` is actually doing.

You might consider refactoring this using `pytest` parameterization. Pass the test function `True`/`False` for `JSON_AS_ASCII` as well as the expected output. This would also stay within the original behavior of creating a new app context each run through.

tests/test_helpers.py
```diff
@@ -34,240 +34,230 @@ def has_encoding(name):
...
         assert flask.json.loads(rv.data) == test_value
 
-    def test_jsonify_dicts(self):
+    def test_jsonify_dicts(self, app, client):
         """Test jsonify with dicts and kwargs unpacking."""
         d = dict(
             a=0, b=23, c=3.14, d='t', e='Hi', f=True, g=False,
             h=['test list', 10, False],
-            i={'test':'dict'}
+            i={'test': 'dict'}
```
jeffwidman: What do you think about switching this from the `dict(key=value)` format to the more common `{key: value}` format? It's not a requirement at all, just a thought...

tests/test_regression.py
```diff
@@ -9,15 +9,14 @@
...
-import os
 import gc
 import sys
-import flask
 import threading
+
+import pytest
 from werkzeug.exceptions import NotFound
 
+import flask
```
jeffwidman: 💯 for properly ordering/spacing imports

tests/test_reqctx.py
```diff
@@ -162,34 +166,37 @@ def g():
...
 
-    rv = app.test_client().get('/?foo=bar')
+    rv = client.get('/?foo=bar')
     assert rv.data == b'Hello World!'
 
     result = greenlets[0].run()
     assert result == 42
 
+
 @pytest.mark.skipif(greenlet is None, reason='greenlet not installed')
```
jeffwidman: If this and the previous test were organized into a `class` specifically for tests about greenlets, then pytest supports conditionally skipping the entire class if `greenlet`s are missing rather than having to mark each test separately.

tests/conftest.py
```diff
@@ -62,12 +99,13 @@ def __getattr__(self, name):
...
     monkeypatch.setattr(pkgutil, 'get_loader', get_loader)
 
 
 @pytest.fixture
 def modules_tmpdir(tmpdir, monkeypatch):
-    '''A tmpdir added to sys.path'''
+    """A tmpdir added to sys.path"""
     rv = tmpdir.mkdir('modules_tmpdir')
     monkeypatch.syspath_prepend(str(rv))
     return rv
```
Tafkas: Good idea to make it another issue. I could work on it then.

tests/test_reqctx.py
Tafkas: Should we also put this as a separate issue?

tests/conftest.py
jeffwidman: https://github.com/pallets/flask/issues/2312

tests/test_reqctx.py
jeffwidman: https://github.com/pallets/flask/issues/2313

jeffwidman:
@Tafkas can you resolve the merge conflicts?

Is this otherwise ready for final review/merge or are you still working on it?

Tafkas:
I am finished. I keep resolving merge conflicts after upstream keeps changing. @davidism

jeffwidman:
Thanks again @Tafkas! Really, really appreciated.

Issue #2251, opened 2017-04-21
Class-based View should set provide_automatic_options attr in as_view


davidism:
~Deprecate in favor of #2249, #1489, #1488, which adds the arg to `add_url_rule`. We couldn't figure out if there was a reason for doing it with an attribute originally.~

davidism:
Ah, maybe it was for supporting `View` classes, like `View.methods`, although the `View` class doesn't set this in `as_view` right now.

arunjohnkuruvilla:
Any pointers on how to get started on this? I am interested in giving it a try.

PR #2316, opened 2017-05-23
Adds provide_automatic_options to Class-based Views

ka7eh:


ka7eh:
Should resolve https://github.com/pallets/flask/issues/2251

PR #2322, opened 2017-05-23
Don't overwrite Vary header when setting for cookie access

### LINKED ISSUE #2317: Don't overwrite Vary header when setting for cookie access
If the `Vary` header is already set, the current session code just adds another header line. What it should do is update the existing header by splitting, adding the new option to the set, joining, and setting the new value. See Django's [`patch_vary_headers`](https://github.com/django/django/blob/1ce04bcce0076360623ae164afd3541a5c031af2/django/contrib/sessions/middleware.py#L45).

Related to #2288

nnja:
Closes #2317

### DIFF
--- a/flask/sessions.py
+++ b/flask/sessions.py
@@ -9,18 +9,20 @@
     :license: BSD, see LICENSE for more details.
 """
 
-import uuid
 import hashlib
+import uuid
 import warnings
-from base64 import b64encode, b64decode
+from base64 import b64decode, b64encode
 from datetime import datetime
-from werkzeug.http import http_date, parse_date
+
+from itsdangerous import BadSignature, URLSafeTimedSerializer
 from werkzeug.datastructures import CallbackDict
+from werkzeug.http import http_date, parse_date
+
+from flask.helpers import patch_vary_header
 from . import Markup, json
 from ._compat import iteritems, text_type
-from .helpers import total_seconds, is_ip
-
-from itsdangerous import URLSafeTimedSerializer, BadSignature
+from .helpers import is_ip, total_seconds
 
 
 class SessionMixin(object):
@@ -405,7 +407,7 @@ class SecureCookieSessionInterface(SessionInterface):
 
         # Add a "Vary: Cookie" header if the session was accessed at all.
         if session.accessed:
-            self._patch_vary_cookie_header(response)
+            patch_vary_header(response, 'Cookie')
 
         if not self.should_set_cookie(app, session):
             return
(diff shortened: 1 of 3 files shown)


flask/sessions.py
```diff
@@ -423,3 +426,16 @@ def save_session(self, app, session, response):
...
+        Add a "Vary: Cookie" header, preserving the original header if one is present.
+
+        The header should retain it's original ordering.
+        """
+        vary_headers = self.comma_delim_header_re.split(response.headers.get('Vary', ''))
+        if 'Cookie' not in vary_headers:
+            vary_headers.append('Cookie')
+
+        updated_header = ', '.join(header for header in vary_headers if header)
+        response.headers.set('Vary', updated_header)
```
ThiefMaster: I would use the `[]` syntax to set it

flask/sessions.py
```diff
@@ -348,6 +349,8 @@ class SecureCookieSessionInterface(SessionInterface):
     """The default session interface that stores sessions in signed cookies
     through the :mod:`itsdangerous` module.
     """
+    #: the regular expression used to split the value of the 'Vary' header
+    comma_delim_header_re = re.compile(r'\s*,\s*')
```
ThiefMaster: Doesn't belong here IMO, it's not meant to be overridable. I think a somewhat common place to store regexes like this is an underscored function arg since they are evaluated only once and do not involve a lookup on the module scope, i.e.

    def _patch_vary_cookie_header(self, response, _comma_re=re.compile(r'\s*,\s*')):

flask/sessions.py
```diff
@@ -423,3 +426,16 @@ def save_session(self, app, session, response):
...
             secure=secure
         )
+
+    def _patch_vary_cookie_header(self, response):
+        """
+        Add a "Vary: Cookie" header, preserving the original header if one is present.
+
+        The header should retain it's original ordering.
+        """
+        vary_headers = self.comma_delim_header_re.split(response.headers.get('Vary', ''))
```
davidism: Don't need a regex:

~~~python
[h.strip() for h in header.split(',')]
~~~

flask/sessions.py
```diff
@@ -423,3 +426,16 @@ def save_session(self, app, session, response):
...
+        if 'Cookie' not in vary_headers:
```
davidism: Need to be case insensitive:

~~~python
if not any(h.lower() == 'cookie' for h in headers)
~~~

flask/sessions.py
davidism: Can just do away with the regex, see below.

flask/sessions.py
ThiefMaster: true, even better.

flask/sessions.py
ThiefMaster: `werkzeug.datastructures.Headers` is case-insensitive

nnja:
Requested changes submitted, ready for re-review.

davidism:
I extracted and renamed the method to `flask.helpers.patch_vary_header` since it could be useful to others modifying the `Vary` header. I'd also like to see this in `werkzeug.http` so that we can eventually import it from there instead, if you'd like to make another PR to Werkzeug.

davidism:
😞 turns out there's already support for this in Werkzeug. #2345 

~~~python
response.vary.add('Cookie')
~~~

PR #2325, opened 2017-05-24
More DRYing up the test suite

Tafkas:
I have found more spots in the test suite that could have been "DRYed" up.

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -153,9 +153,9 @@ def test_session_transaction_needs_cookies(app):
     assert 'cookies' in str(e.value)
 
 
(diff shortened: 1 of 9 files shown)


tests/test_testing.py
```diff
@@ -266,7 +258,7 @@ def view(company_id):
     with app.test_request_context():
         url = flask.url_for('view', company_id='xxx')
 
-    with app.test_client() as c:
+    with client as c:
```
davidism: All these can be `with client:` and continue to use `client` in the block. No need to rebind the name in this case.

Tafkas:
@davidism fixed it :)

PR #2326, opened 2017-05-24
Handle app factory with arguments in FLASK_APP

HndrkMkt:
#2264

PR #2338, opened 2017-05-26
cleanup #2326

kennethreitz:


kennethreitz:
Additional info may be needed here, but this is an improvement.

Issue #2341, opened 2017-05-27
If FLASK_APP accept one argument and I do not send it, the default value is not correct, instead I get a ScriptInfo object.


jrbaez01:
An example is in the Flickr app [link here](https://github.com/pallets/flask/blob/master/examples/flaskr/flaskr/factory.py#L19)

As a newcomer it is confusing when I follow the [README.md](https://github.com/pallets/flask/blob/master/examples/flaskr/README#L24), but the when running the app it results in the following error. 
![image](https://cloud.githubusercontent.com/assets/11146130/26524504/51b31274-4302-11e7-8dd6-21426b08bbbf.png)

jrbaez01:
Hello @davidism, 

Suggested solution to this issue: https://github.com/pallets/flask/pull/2342, hope it is useful.

Cheer.

davidism:
Closed by PR.

PR #2342, opened 2017-05-27
#2341: Accept default argument value when args lenght equal 1

jrbaez01:
Fix issue https://github.com/pallets/flask/issues/2341


davidism:
Good catch, this makes sense. Thanks!

Issue #1511, opened 2015-06-20
TaggedJSONSerializer loads(dumps(..)) breaks on certain values


ThiefMaster:
``` diff
diff --git a/tests/test_basic.py b/tests/test_basic.py
index 695e346..1e064ff 100644
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -387,6 +387,7 @@ def test_session_special_types():
         flask.session['dt'] = now
         flask.session['b'] = b'\xff'
         flask.session['t'] = (1, 2, 3)
+        flask.session['spacefirst'] = {' t': 'not-a-tuple'}
         return response

     @app.route('/')
@@ -403,6 +404,7 @@ def test_session_special_types():
     assert rv['b'] == b'\xff'
     assert type(rv['b']) == bytes
     assert rv['t'] == (1, 2, 3)
+    assert rv['spacefirst'] == {' t': 'not-a-tuple'}


 def test_session_cookie_setting():
```

While I don't see any cases where this would be likely to result in security problems (it would require custom types that are not safe to recreate with arbitrary arguments AND code that serializes a dict with both the key and value being provided by the user), I don't like that there's any input where `value != loads(dumps(value))` without an exception.

Two ways to fix this:
- raise an exception when trying to encode a dict that has a single item with a key starting with a space.
- tag such dictionaries so they are decoded properly instead of being decoded as if they are a valid tagged it em

PR #2352, opened 2017-06-02
Make session serializer extensible

### LINKED ISSUE #1438: Make TaggedJSONSerializer extensible
My suggestion would be to add a method to the class that lets you register custom types, e.g. like this:

```
session_json_serializer.register_type('md', MultiDict)
```

This would use the `' md'` prefix when encountering a `MultiDict` object (`type(x) is MultiDict` as subclasses probably shouldn't be handled). To serialize/deserialize I'd simply use `__getstate__` and `__setstate__`.

davidism:
Continues #1452.

Refactors `flask.session.TaggedJSONSerializer` to a new `flask.json.tag` module. Creates a `JSONTag` interface and a `TaggedJSONSerializer.register` method to register new tags.

~~~python
from flask.json.tag import JSONTag

class TagOrderedDict(JSONTag):
    __slots__ = ('serializer',)
    key = ' od'

    def check(self, value):
        return isinstance(value, OrderedDict)

    def to_json(self, value):
        return [[k, self.serializer.tag(v)] for k, v in iteritems(value)]

    def to_python(self, value):
        return OrderedDict(value)

app.session_interface.serializer.register(TagOrderedDict, 0)
~~~

Since `register` takes an index, it can be used to override existing tags (although it would be slightly more efficient to subclass `TaggedJSONSerializer` instead). This can address #1908 by adding a higher-priority `dict` serializer that uses a different format to support non-string keys.

Closes #1438, #1511, #1908.

~Still needs:~

- [x] refactor to move the `serializer` argument to `self.serializer` instead. It was an attempt at optimization but didn't change anything.
- [x] refactor `TagDict` to split the logic between addressing #1511 and regular dict passthrough, as #1452 originally did. This was another optimization attempt that didn't help, it just made the code more confusing.
- [x] more tests
- [x] documentation


flask/json/tag.py
```diff
@@ -0,0 +1,253 @@
...
+from flask._compat import iteritems, text_type
+from flask.json import dumps, loads
+
+
+class JSONTag(object):
+    """Base class for defining type tags for :class:`TaggedJSONSerializer`."""
+
+    __slots__ = ('serializer',)
+    key = None
+    """The tag to mark the serialized object with. If ``None``, this tag is
```
ThiefMaster: AFAIK the most common sphinx way of documenting a class-level variable is this:

```python
#: lorem ipsum
#: blah blah blah
foo = None
```

At least I find this "dangling" docstring quite ugly and I think some editors even highlight it as a no-op statement  and/or a docstring that's not where it should be/

flask/json/tag.py
```diff
@@ -0,0 +1,253 @@
...
+    only used as an intermediate step during tagging."""
+
+    def __init__(self, serializer):
+        """Create a tagger for the given serializer."""
+
```
ThiefMaster: only class-level docstrings are followed by an empty line

flask/json/tag.py
davidism: Sphinx and PyCharm both recognize it. Not a big deal though, I can switch to `#:` and ignore PyCharm.

davidism:
@ThiefMaster from https://github.com/pallets/flask/pull/1452#issuecomment-305609953:

Re: performance: new method is ~ 30 µs slower.

~~~
old: 70.7 µs ± 164 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
new: 96.3 µs ± 678 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
~~~

You'd need to handle 2,000,000 requests before you wasted 1 extra minute. There's always Flask-Session, etc. where server-side sessions are serialized with `pickle`, which is *much* faster.

flask/json/tag.py
ThiefMaster: ah ok, never noticed pycharm handling it nicely. probably never opened a file using this notation ;)

I just checked and `#:` is quite widespread in the flask codebase so I'd go for that

ThiefMaster:
ok, that seems fine

PR #2353, opened 2017-06-02
Clarify documentation for json parsing

grantwwu:
Documentation does not currently mention that is_json accepts mimetypes that are not strictly application/json.

### DIFF
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -97,8 +97,9 @@ class Request(RequestBase):
 
     @property
     def json(self):
-        """If self.is_json would return true, this will contain the
-        parsed JSON data.  Otherwise this will be ``None``.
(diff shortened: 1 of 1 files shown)


davidism:
I think this actually loses clarity. The number one bug with people trying to read JSON is that they didn't set `Content-Type: application/json`, and now it's not mentioned next to the methods that require it. Mentioning `is_json` is good, but removing the other info is not.

davidism:
Please use the following syntax for methods:

    :meth:`is_json`

davidism:
Thanks, I adjusted this to mention both the mimetype and the method.

grantwwu:
Oh, sorry.  I thought about mentioning both application/json and is_json, but wasn't sure and decided to make a PR for feedback.

I see you got this done while I was at lunch, thanks!

PR #2356, opened 2017-06-03
doc: following web standards

DingGGu:
Add double quotation marks on documents for following web standards.

https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2


davidism:
That's not a standard, either works. Armin's style choice is to leave them off when unnecessary.

ThiefMaster:
Just FYI, you linked to a rather ancient standard (HTML4). Read the HTML5 spec instead ;)

PR #2369, opened 2017-06-12
Add docutils label for templates page

tony:
For consistency with rest of pages. Linkable via intersphinx

### DIFF
--- a/docs/templating.rst
+++ b/docs/templating.rst
@@ -1,3 +1,5 @@
+.. _templates:
+
 Templates
 =========
 


davidism:
Can't you already link to the page with

    :ref:`templating`

?

tony:
This isn't the only project where I'm having similar issues. [Django has this problem too in areas where labels are explicit](https://code.djangoproject.com/ticket/28290).

```rst
:ref:`Template system <templating>`
```

Works.

```rst
:ref:`Template system <flask:templating>`
```

(which I need to avoid collisions) Won't. 

In my situation, I'm using multiple intersphinx sources, so I use [prefixes](http://www.sphinx-doc.org/en/stable/ext/intersphinx.html). I'm writing an expansive article on Flask/Django ATM at https://github.com/git-pull/website/issues/20 90% of intersphinx links work flawlessly.

> The unique identifier can be used to prefix cross-reference targets, so that it is clear which intersphinx set the target belongs to. A link like
> ```rst 
>:ref:`comparison manual <python:comparisons>`
> ```
> will link to the label “comparisons” in the doc set “python”, if it exists.

It seems that not explicitly labelling causes prefixed targets not to work.

```rst
:ref:`URL routing <flask:url-route-registrations>`
```

Works.

It uses [an explicit label](https://raw.githubusercontent.com/pallets/flask/master/docs/api.rst) (ctrl-f to ``.. _url-route-registrations:``)

Also,

```rst
:ref:`API <flask:api>`
```

Works as well ([explicit label at top](https://raw.githubusercontent.com/pallets/flask/master/docs/api.rst))

```rst
:doc:`Template system <flask:templating>`
```

Works.

I do think it's best to be consistent w/ the rest of the docs, in any event.

Sphinx 1.6.2 + Python 3.6.1 

Related ticket: https://github.com/sphinx-doc/sphinx/issues/3856

I may consider a modified ticket where prefixed ``:ref:``'s don't work with implicit document labels.

davidism:
Right, it's an external ref. That's fine then.

tony:
What do you mean by fine?

PR #2370, opened 2017-06-12
Add docutils label for class-based views

tony:
Make section available via intersphinx


tony:
Can this and #2369 be backported to 0.12 too?

Related to https://github.com/git-pull/website/pull/20

PR #2374, opened 2017-06-14
Correctly JSON-encode datetimes aware of non-UTC timezones

### LINKED ISSUE #2372: JSONEncoder encodes aware datetime objects in non-UTC timezone incorrectly as being in GMT.
``flask.json.JSONEncoder()`` encodes aware datetime objects in non-UTC timezone incorrectly using local time tuple with GMT.

### Expected Behavior

The following code segment should print ``"Sun, 01 Jan 2017 20:34:56 GMT"``  (12:34:56 PST8PDT is 20:34:56 GMT):

```python
from datetime import datetime
from pytz import timezone
dt = timezone('PST8PDT').localize(datetime(2017, 1, 1, 12, 34, 56))

from flask.json import JSONEncoder
encoder = JSONEncoder()
print(encoder.encode(dt))
```

### Actual Behavior

The preceding code segment prints ``"Sun, 01 Jan 2017 12:34:56 GMT"`` instead, that is, local timetuple() appended with GMT.

### Environment

* Python version: 3.5.3
* Flask version: 0.12.2
* Werkzeug version: 0.12.2

astralblue:
http_date() requires timetuple in UTC, but JSONEncoder.default() was
passing a local timetuple instead.

This fixes #2372.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -79,6 +79,7 @@ Major release, unreleased
 - Removed error handler caching because it caused unexpected results for some
   exception inheritance hierarchies. Register handlers explicitly for each
   exception if you don't want to traverse the MRO. (`#2362`_)
+- Fix incorrect JSON encoding of aware, non-UTC datetimes. (`#2374`_)
 
 .. _#1489: https://github.com/pallets/flask/pull/1489
 .. _#1621: https://github.com/pallets/flask/pull/1621
@@ -102,6 +103,7 @@ Major release, unreleased
 .. _#2354: https://github.com/pallets/flask/pull/2354
 .. _#2358: https://github.com/pallets/flask/pull/2358
 .. _#2362: https://github.com/pallets/flask/pull/2362
+.. _#2374: https://github.com/pallets/flask/pull/2374
 
 Version 0.12.2
 --------------
--- a/flask/json/__init__.py
+++ b/flask/json/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import io
 import uuid
-from datetime import date
+from datetime import date, datetime
 from flask.globals import current_app, request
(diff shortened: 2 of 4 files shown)


davidism:
Thanks! Nice use of parametrize in the test.

Is it possible to test timezones without depending on pytz?

astralblue:
`datetime.tzinfo` should theoretically be usable for simple cases (as we do not need DST logic of pytz to test JSONEncoder).  Will look into that.

astralblue:
The new method fails because datetime.timezone is Python 3 only.  x_x  The only alternative would be to hand-spin a timezone class, which I feel is an overkill.  x_x

davidism:
I'm fine with a test dependency on pytz then.

davidism:
Be sure to add the dependency to the extra requirements in `setup.py` as well as `tox.ini`.

davidism:
I take it back, that would mean a simple call to `pytest` would fail unless we made the test optional.

astralblue:
Oops, I missed the edit to your comment and misunderstood.  Will revert these changes.

davidism:
😓 sorry for the confusion

zsiegel92:
Hey all, Flask 0.12.2 was released in May, so this can't be part of that. Will this be part of 0.13? For now, I will send timezone info separately.

Thank you!!

davidism:
Jinja releases aren't tied to Flask releases.

davidism:
This will be part of the next release.

spacemunkey77:
Using  Flask 0.13-dev and jsonify() is still setting times in UTC. I have set the local time zone using the following code snippet:

    tz = TimezoneFinder()
    ltz = tz.timezone_at(lng=longitude, lat=latitude)
    os.environ['TZ'] = ltz
    time.tzset()

PR #2378, opened 2017-06-15
Import app from wsgi.py or app.py if FLASK_APP is not defined

### LINKED ISSUE #2376: flask command should auto-detect module
Even when installed as a package, it can be useful in complex situations to have a `wsgi.py` file for setting the app up to run. The `flask` command should detect this file if no `FLASK_APP` is specified at all.

Additionally, it's common to name quick demo apps `app.py`, so that should be detected too.

From discussion in #2375.

miguelgrinberg:
Fixes #2376

This patch adds support for looking for the Flask app instance in `wsgi.py` or `app.py` modules (in that order) when `FLASK_APP` is not defined.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -25,12 +25,12 @@ your Flask application's :attr:`Flask.cli` instance as well as some
 built-in commands that are always there.  Flask extensions can also
 register more commands there if they desire so.
 
-For the :command:`flask` script to work, an application needs to be
-discovered. Flask looks for a module named wsgi.py or app.py by default, and
-if it finds one it assumes the application is defined in it.
+For the :command:`flask` script to work, an application needs to be discovered.
+Flask looks for a module named :file:`wsgi.py` or :file:`app.py` by default,
+and if it finds one it assumes the application is defined in it.
 
 You can instruct Flask to look for the application in a different module by
-exporting the ``FLASK_APP`` environment variable.  It can be either set to an
+exporting the ``FLASK_APP`` environment variable. It can be either set to an
 import path or to a filename of a Python module that contains a Flask
 application.
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -26,9 +26,13 @@ built-in commands that are always there.  Flask extensions can also
 register more commands there if they desire so.
 
 For the :command:`flask` script to work, an application needs to be
-discovered.  This is achieved by exporting the ``FLASK_APP`` environment
-variable.  It can be either set to an import path or to a filename of a
-Python module that contains a Flask application.
+discovered. Flask looks for a module named wsgi.py or app.py by default, and
+if it finds one it assumes the application is defined in it.
+
+You can instruct Flask to look for the application in a different module by
+exporting the ``FLASK_APP`` environment variable.  It can be either set to an
+import path or to a filename of a Python module that contains a Flask
+application.
 
 In that imported file the name of the app needs to be called ``app`` or
 optionally be specified after a colon.  For instance
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -174,12 +174,14 @@ def prepare_exec_for_file(filename):
 
     if sys.path[0] != dirpath:
         sys.path.insert(0, dirpath)
+
     return '.'.join(module[::-1])
 
 
 def locate_app(script_info, app_id, raise_if_not_found=True):
     """Attempts to locate the application."""
     __traceback_hide__ = True
+
     if ':' in app_id:
         module, app_obj = app_id.split(':', 1)
     else:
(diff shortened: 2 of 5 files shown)


lepture:
I'm wondering why `wsgi.py` is placed before `app.py`?

`wsgi.py` is usually used for production environment. In this case, we don't have to use flask cli to create the server. On the other hand, `app.py` is what we used for development, it is better to put `app.py` before `wsgi.py`.

ThiefMaster:
I guess many custom CLI functions are also used in production. Either way, if it's a `wsgi.py` I'd expect it to be used only when running the WSGI app - not necessarily for other CLI functionality.

Anyway, if people do use CLI commands in production I think they should strongly consider adding their own CLI that knows how to instantiate the app instead of using the generic `flask` command...

miguelgrinberg:
@lepture I don't see the use of `wsgi.py` and `app.py` as being different for production or development. Regardless of the one you use, it's going to be part of your project, so it will be present in both environments. I used very simple logic to decide that `wsgi.py` should be first:

- If you have neither `app.py` nor `wsgi.py` in your project, it does not matter which one you check first.
- If you have only one of `app.py` and `wsgi.py` in your project, it also does not matter.
- If you have both `app.py` and `wsgi.py` in your project, I think it would be wrong to ignore `wsgi.py` in favor of `app.py`, since I see no other function for a `wsgi.py` file than to be the module that defines the app.

@ThiefMaster I don't follow some of your statements:

> Either way, if it's a `wsgi.py` I'd expect it to be used only when running the WSGI app - not necessarily for other CLI functionality.

The app is going to be needed when running most CLI commands. Maybe there are some cases in which you want to define custom commands that do not require the app, but that is not going to be the general case. In fact, the CLI defaults to creating an app context for your command, so right there you can see that you have to have the app.

> Anyway, if people do use CLI commands in production

It's not just people. Extensions define their own commands as well. My Flask-Migrate extension adds a bunch of commands under `flask db` that are used in production and development.

>  I think they should strongly consider adding their own CLI

That would be tricky for commands added by extensions. But even for the app's own custom commands, what's the benefit of using a separate CLI? There is nothing wrong with that idea if that's what you prefer, but I don't see how that would be a recommendation to be made in the general case.

>  that knows how to instantiate the app instead of using the generic flask command...

Well, the flask command has to know how to instantiate the app anyway, so why invent a different method for custom commands? It guess you are thinking that the flask command does not have a place in production, but in any case, why would you want to create the app in a different way between dev and prod? Normally the app is the same, and you just change the configuration.

ThiefMaster:
> The app is going to be needed when running most CLI commands.

Sure, I just meant that you don't use the WSGI part of Flask in the CLI, so `wsgi.py` is not *that* obvious there

> My Flask-Migrate extension adds a bunch of commands under flask db that are used in production and development.
> That would be tricky for commands added by extensions

When using `FlaskGroup` for the CLI command defined via entry points show up in a custom CLI too

> but in any case, why would you want to create the app in a different way between dev and prod

I wouldn't - I'd use the custom command for both dev and prod. My point was just that if your app does use more advanced CLI commands than just `run` and `shell` it might be a good point to add a custom CLI, especially considering that it doesn't take more than a few lines of code.

lepture:
> If you have both app.py and wsgi.py in your project, I think it would be wrong to ignore wsgi.py in favor of app.py, since I see no other function for a wsgi.py file than to be the module that defines the app.

Yes. I always have both `app.py` and `wsgi.py` in my project. The `wsgi.py` as I understand, is designed to be a pure wsgi applcation, which would contain just a few lines of code. Something as simple as:

```python
from your_project import create_app
app = create_app()
```

The `wsgi.py` file could be used by gunicorn/uwsgi etc. The idea comes from Django.

Like Django, I takes `app.py` as the `manage.py`, which could contain a lot of custom commands, e.g. `initdb`, `createuser`.

miguelgrinberg:
> Sure, I just meant that you don't use the WSGI part of Flask in the CLI, so wsgi.py is not that obvious there

Fair enough. I see what you meant now. I don't know how to separate the WSGI and non-WSGI uses of the app instance. Using `wsgi.py` is totally clear in my mind as a way to tell a web server what is the application, and you need to do this whether you use `flask run`, gunicorn or uwsgi. The `flask shell` or `flask something-else` cases are strange, I agree. But given that you need to define the app for the web server no matter what, you'd probably want the same method to be used for the other commands. I guess you could use `app.py` and then WSGI is out of the picture, at least the name is.

miguelgrinberg:
> When using FlaskGroup for the CLI command defined via entry points show up in a custom CLI too

@ThiefMaster Yeah, you are correct. I just verified that at least for Flask-Migrate the commands do make it to the custom CLI, so that's great. Not only that, also `run` and `shell` are exposed in the custom script. So effectively, a custom CLI looks pretty much the same as the `flask` command, with the difference that in a custom CLI you write the code that loads your app in the script itself, and with the `flask` command you point to it to a separate module that does it. The difference is minimal, in my opinion. My `wsgi.py` module has the exact same content I would put in the factory function referenced in the `create_app` argument to the click group. As I see it, which one you use is a matter of preference, I don't really see a clear benefit to use one or the other.

greyli:
Does it mean `app.py` is the best name to store application instance (i.e. `app`) or application factory (i.e. `create_app`)? 
In this case, I got two supposes:
- Naming the main script as `app.py` in the small project.
- Put the application factory `create_app` in `app.py` module instead of `__init__.py` in the big project.

Is that right?

davidism:
Right on point one, wrong on point two.

greyli:
Oh, sorry, I misunderstood it. 

In the large project, we often need a file to create application instance outside of the application package. For example:
```python
from myapp import create_app

app = create_app()
```
We often call it `devapp.py` , `autoapp.py` or `wsgi.py` when deploy. So now the better name is `wsgi.py`.

davidism:
You don't need to do that. `FLASK_APP=myapp` will pick up the factory automatically.

greyli:
Thanks, you are very helpful!

Issue #2377, opened 2017-06-15
flask command should not require .py extension


davidism:
If you don't install your app in the env, you have to append `.py` in the `FLASK_APP` var.

Haven't dug into why it's like this yet, but it might be more straightforward to add the cwd to PYTHONPATH so that `FLASK_APP=app.py` becomes `FLASK_APP=app` even if the app isn't installed in the env.

We can still support paths to set the cwd. `FLASK_APP=/path/to/app` would set the current directory to `/path/to` then try to import `app`.

PR #2383, opened 2017-06-18
Don't require FLASK_APP to have .py extension

bcongdon:
Addresses #2377. The simplest way that I found to do this was to just patch the `find_default_import_path` method to check if there's a file with the `.py` extension.

Worked in all of my testing. 🙂


davidism:
Thanks! This is sort of the opposite direction of what I was imagining though. For example, a package that's not installed still requires `FLASK_APP=myapp/__init__` when it should be importable as `myapp` because the current directory is on the `PYTHONPATH`.

bcongdon:
Ah, I see. I updated the method to append the sys path if the referenced `FLASK_APP` is a directory. That should cover both the case where `FLASK_APP=myapp` and `FLASK_APP=/path/to/myapp` (where `myapp` is a module that contains a Flask app)

CHANGES
```diff
@@ -82,6 +82,8 @@ Major release, unreleased
 - Fix incorrect JSON encoding of aware, non-UTC datetimes. (`#2374`_)
 - Template auto reloading will honor the ``run`` command's ``debug`` flag even
   if ``app.jinja_env`` was already accessed. (`#2373`_)
+- The ``flask`` command no longer requires that the ``FLASK_APP`` environment
+  variable have a ``.py`` extension. (`#2383`_)
```
pachewise: @bcongdon - you need to setup a target for this one:

`_#2383: https://github.com/pallets/flask/pull/2383`

flask/cli.py
```diff
@@ -222,8 +230,11 @@ def find_default_import_path():
     app = os.environ.get('FLASK_APP')
     if app is None:
         return
-    if os.path.isfile(app):
-        return prepare_exec_for_file(app)
+    for path in [app, app + '.py']:
+        if os.path.isfile(path):
+            return prepare_exec_for_file(path)
+    if os.path.isdir(app):
```
pachewise: I would put the `isdir(app)` block above L233.

CHANGES
bcongdon: Ah, missed that. Thanks!

davidism:
A friend pointed out that `FLASK_APP=module flask run` already works without the `.py`. So now I'm trying to figure out why that's importable but `FLASK_APP=package` isn't.

pachewise:
@davidism wouldn't it hinge on os.path.isfile() behavior? In that case, don't see how `FLASK_APP=module` would work unless os.path doesn't care about file extensions...

davidism:
Because the current directory is already inserted into the path.

jmsdvl:
@davidism are you sure `FLASK_APP=module flask run` already works?  This is the behavior I'm seeing (without installing either `mod` or `pack` into the current environment):

* `FLASK_APP=(mod|package) flask run`: immediate error; file/path does not seem to exist, etc. etc.
* `FLASK_APP=(mod|package) FLASK_DEBUG=1 flask run`: the server _starts_ but as soon as you request a page it throws a `flask.cli.NoAppException`

However (not surprisingly):

* `FLASK_APP=(module|pack) FLASK_DEBUG=(0|1) python -m flask run`: always works

In no case does it seem to matter if `FLASK_APP` is pointed at a module or a package.

davidism:
Thanks for working on this. I ended up fixing this in #2414 by dropping the `.py` requirement rather than handling directories differently.

PR #2386, opened 2017-06-21
Add command to CLI to generate secure random strings

xMamo:
I wrote command-line utility to generate a secure random string. It can be used, for example, to set Flask's `SECRET_KEY`.
The command is called with `flask generate-secret [--length <length>]` (length defaults to 30).

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -442,6 +442,7 @@ class FlaskGroup(AppGroup):
             self.add_command(run_command)
             self.add_command(shell_command)
             self.add_command(routes_command)
+            self.add_command(generate_secret_command)
 
         self._loaded_plugin_commands = False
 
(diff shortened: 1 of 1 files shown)


davidism:
Needs a test and docs. Should mention something like "Use the `flask generate-secret` command to get a random value." in the `SECRET_KEY` docs.

I'm curious where you got the pattern you're using here. Seems weird to work backwards to get a string that's close to the desired length. I'd rather input the number of random bytes to start with and get a string of whatever length is necessary.

    -b, --bytes byte length of initial random data
    b64encode(os.urandom(n)).decode('UTF-8')

16 bytes (128 bits) is probably a sufficient default.

ThiefMaster:
I don't see the point of using a base64 string considering that most flask config files can contain python code and thus backslash-escaped strings.

Also, I'm not sure whether this is really within the scope of flask.

davidism:
Hmm, true, and we already mention how to generate a key as raw bytes in the docs. Although we should probably make it more clear outside the quickstart, in the config and session docs.

untitaker:
I would use hexlify instead of base64 to ensure a safer subset of ASCII. Whether asciifying is necessary.. I just wouldn't take unnecessary risks.

davidism:
Actually, looks like I already added an example of how to generate it to the SECRET_KEY docs. I think I'm fine with keeping this a manual process.

davidism:
Closing this because I don't think we need a built-in command for it. I'd rather the user understand how the secret is generated, which is already documented.

Issue #2385, opened 2017-06-21
Remove deprecated code


davidism:
Remove old code that has deprecation warnings:

- [x] **`Flask.ext`**
- [x] `Flask.init_jinja_globals`
- [x] `Flask.error_handlers`
- [x] `Flask.request_globals_class`
- [x] `Flask.static_path`
- [x] ~`send_file` mimetype guessing~ removed in 0.12 #1849
- [x] `Request.module`
- [x] `scripts/flask-07-upgrade.py`

davidism:
The only one I'm slightly worried about is `flask.ext`, but it's been deprecated almost a year now and there's plenty of information out there about what to change. We had a push to correct the documentation in a lot of extensions and docs, and I haven't seen it pop up for a long time on Stack Overflow. I'd rather not have it in 1.0.

PR #2394, opened 2017-06-26
Remove deprecated code

davidism:
* `Flask.error_handlers`
* `Flask.static_path`
* `Request.module`

#2385

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -268,40 +268,40 @@ def url_for(endpoint, **values):
     """
     appctx = _app_ctx_stack.top
     reqctx = _request_ctx_stack.top
+
     if appctx is None:
-        raise RuntimeError('Attempted to generate a URL without the '
-                           'application context being pushed. This has to be '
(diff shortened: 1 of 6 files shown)


flask/helpers.py
```diff
@@ -268,40 +268,40 @@ def external_url_handler(error, endpoint, values):
     """
     appctx = _app_ctx_stack.top
     reqctx = _request_ctx_stack.top
+
     if appctx is None:
-        raise RuntimeError('Attempted to generate a URL without the '
-                           'application context being pushed. This has to be '
-                           'executed when application context is available.')
+        raise RuntimeError(
+            'Attempted to generate a URL without the application context being'
```
ThiefMaster: Mostly just a matter of taste but I'd keep the space at the end of the line instead of the beginning of the next one

davidism: I started doing it this way because I removed the space by mistake a lot more often when it appeared at the end.

PR #2398, opened 2017-06-27
fix variable substitution in error message

miguelgrinberg:
@davidism looks like you've made a small mistake in how the module variable is replaced in the error message when you edited my commit from a few days ago. This should address it.


davidism:
Oops, good catch.

PR #2403, opened 2017-06-30
Remove unused import from test_basic.

wdhorton:
Removed the unused import of `pickle` in `tests/test_basic`.

Pickle isn't used explicitly in the file, and I ran the tests and they all still pass.


ThiefMaster:
Thanks for the PR, but usually it's not really worth to open PRs to remove unused imports - better to simply remove them when changing something else in the file.

PR #2417, opened 2017-07-15
[0.12.x] Remove outdated text on flask.ext redirect package

Tethik:
...from 0.12.x documentation. i.e. here: http://flask.pocoo.org/docs/0.12/extensiondev/#ext-import-transition

I noticed that the live documentation for 0.12.x has this paragraph on the current stable version saying that the flask.ext. namespace should be used. This is deprecated since 0.11.x right? The development version has that paragraph removed, so I just did the same for this branch.

### DIFF
--- a/docs/extensiondev.rst
+++ b/docs/extensiondev.rst
@@ -29,12 +29,6 @@ be something like "Flask-SimpleXML".  Make sure to include the name
 This is how users can then register dependencies to your extension in
 their :file:`setup.py` files.
 
-Flask sets up a redirect package called :data:`flask.ext` where users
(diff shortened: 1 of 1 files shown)


davidism:
Would be forward-merged from the 0.11 pr.

davidism:
Unfortunately, we can't roll out new docs without a new release, and I don't plan on making any more 0.12 releases.

If you'd like to look over the current docs to make sure we're not mentioning `flask.ext` any more, that would be helpful. Thanks for your interest!

Tethik:
Alright, thanks for letting me know.

PR #2418, opened 2017-07-15
[0.11.x] Remove outdated text on flask.ext redirect package

Tethik:
See #2417. Same thing except for 0.11.x instead of 0.12.x


davidism:
Unfortunately, we can't roll out new docs without a new release, and I don't plan on making any more 0.11 releases.

PR #2422, opened 2017-07-18
Add more information to doc regarding PERMANENT_SESSION_LIFETIME config.

asmith26:
I was struggling to identify how to reduce the time possible for a replay attack. I came across [this stackoverflow answer](https://stackoverflow.com/a/25137976/5179470) which provided me the necessary information - hence updating here.


davidism:
Thanks for bringing this up. Markdown link syntax isn't valid in ReStructuredText. I also felt that the config docs should remain brief, while the security docs should be updated to talk about cookie expiration and the session signature. I pushed a commit with those changes.

PR #2425, opened 2017-07-20
give context about purpose of this section, #2421

### LINKED ISSUE #2421: Tutorial: step 3 (Installing as package) needs rationale
Step 3 of the tutorial ([installing flaskr as a package](http://flask.pocoo.org/docs/0.12/tutorial/packaging/)) needs rationale at the top. As it stands, that section of the tutorial tells the user out of the blue that they should install their webapp as a package, but doesn't say why. The user can already run their webapp via `flask run`, so it's not clear why they should at this point set up and install their package into their venv.

uvtc:
Otherwise it's unclear why you should be installing at all (the quickstart describes a hello.py file that doesn't require any installation). Resolves #2421.

### DIFF
--- a/docs/tutorial/packaging.rst
+++ b/docs/tutorial/packaging.rst
@@ -3,6 +3,11 @@
 Step 3: Installing flaskr as a Package
 ======================================
 
+It's customary to install your Flask application package into your
+project's venv right from the start. As described below, you'll use
+pip to install it such that code changes will automatically propagate
+to the installed package.
+
 Flask is now shipped with built-in support for `Click`_.  Click provides
 Flask with enhanced and extensible command line utilities.  Later in this
 tutorial you will see exactly how to extend the ``flask`` command line


davidism:
But this doesn't clarify why you would install it. It just says it's customary, which was basically the same issue you had with the original.

uvtc:
Sorry, maybe I poorly explained what my original problem was, or used poor choice of words.

I had already read the quickstart, and was following along with the tutorial, and then hit a chapter that started talking about how to install my webapp. I had no idea why this would be necessary, since in the quickstart all I had to do was set some env vars and do `flask run`. Further, I had no idea that it was even possible to install a package into my venv and have auto-reload automatically work (`pip --editable .`). I just wanted to set some env vars and run `flask run` and have it work. What I was shooting for in the paragraph I added in the PR is: "don't panic, when your webapp is a package (instead of a top-level hello.py script), you must install it and that's just part of the dev process". I'm fine with that explanation, and I think the reader will be as well (at least, where they are for now, in chapter 3 of the tut).

I'm guessing that later, during deployment, when the instructions say something like, "so, install your webapp package into...", then the reader will be pleased to know that they already have their webapp set up as a package, and that this is one of the reasons it was useful to set that up in chapter 3 of the tut.

jeffwidman:
This intent of this PR resonates with me, as I had pretty much the same reaction when I first started working with flask and wasn't very familiar with python packaging of "why would I do this, `flask run` does what I need..." 

but I also agree with @davidism that the "customary" wording isn't helpful to n00bs--it would be much better to actually explain the benefits of creating your own python package.

davidism:
Closing in favor of a more thorough rewrite. Thanks for bringing this up.

greyli:
@jeffwidman I am confusing about it for a long time... Could you give me some reasons about why we need to install it? Thanks!

PR #2430, opened 2017-07-27
make_test_environ_builder: use url_scheme from path if provided

erebus1:
When providing https url in path ("https://example.com/")
we hope that we will get https scheme in environment

### DIFF
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -40,10 +40,10 @@ def make_test_environ_builder(
         if subdomain:
             http_host = '{0}.{1}'.format(subdomain, http_host)
 
+	url = url_parse(path)
         if url_scheme is None:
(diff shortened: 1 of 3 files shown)


flask/testing.py
```diff
@@ -40,10 +40,10 @@ def make_test_environ_builder(
         if subdomain:
             http_host = '{0}.{1}'.format(subdomain, http_host)
 
+	url = url_parse(path)
```
davidism: style: should have empty line before block

jeffwidman: I'm curious why? I don't recall that in pep8, am I forgetting something?

ThiefMaster: AFAIK there's anything about it in PEP8, so I think it's a matter of taste.

davidism: More importantly, this should be indented with spaces, not tabs.

Issue #2023, opened 2016-09-12
How should logging in Flask look like?


untitaker:
Flask started to ship with a default, hardcoded logging handler. Unfortunately this setup makes it harder to install custom logging setups, because then you'll have to undo all the things Flask did to the app logger, or replace the `app.logger` entirely. A symptom of this is #1993, where Flask's own logger had to be tweaked yet again such that messages didn't get logged twice (once via Flask's setup, once via the custom one).

My question is: **Do we even want Flask to do any logging setup?** It appears that this sort of default logging is only useful during development, so maybe it makes sense to set up a default logging handler in the new Flask CLI instead of from within the application.

jaapz:
What would happen to the log messages in production, when the current flask logger would be moved to development only?

davidism:
The same thing that already happens: you need to add your own logging handler in order to see messages and exceptions generated in production.

untitaker:
Why does that already happen?

On 13 September 2016 10:56:52 CEST, David Lord <notifications@github.com> wrote:
>The same thing that already happens: you need to add your own logging
>handler in order to see messages and exceptions generated in
>production.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/pallets/flask/issues/2023#issuecomment-246618501

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

davidism:
@untitaker I'm not sure.  It's weird because `flask.logging` *seems* to configure a log handler for outside debug mode.  I and [many others](http://stackoverflow.com/questions/32722143/flask-application-traceback-doesnt-show-up-in-server-log) don't see logs unless a handler is explicitly set though.

Rdbaker:
Python takes a modular approach to its logging by separating the concerns into [loggers, handlers, filters, and formatters](https://docs.python.org/2/howto/logging.html#advanced-logging-tutorial). Maybe it would be good to have a very simple, yet explicit, default setup for each of these classes with some documentation on how to easily override it, just as flask does for the request and response classes.

untitaker:
That is actually the current state. The problem arises when people use `logging.defaultConfig()` and messages get logged twice, once with Flask's handler, once with the just configured one.

The question is whether it's a good idea to even have any logging setup by default as it may interfere with the plans of the app developer.

On 14 September 2016 03:44:50 CEST, Ryan Baker <notifications@github.com> wrote:
[...]

Issue #641, opened 2012-11-29
Design issue with flask.logger.create_logger().


brownhead:
[`flask.logger.create_logger()`](https://github.com/mitsuhiko/flask/blob/master/flask/logging.py#L17) destroys any log handlers that have been added previously. Because the creation of the logger (and the calling of `create_logger()` that this entails) is [deferred until the first time `flask.app.logger` is accessed](https://github.com/mitsuhiko/flask/blob/master/flask/app.py#L548), this can lead to the destruction of log handlers that were purposefully added by the application.

I just tracked down a bug due to this that took me a very long time to debug because I had no idea this went on. To my eyes, it looked like the standard logging module was breaking at a seemingly random moment (it wasn't even my code that was accessing `app.logger`, it was [Flask's internal error handling code](https://github.com/mitsuhiko/flask/blob/master/flask/app.py#L1347)). Therefore, I think one of two courses of action should be taken:
- Because this behavior of `create_logger()` seems very strange, perhaps it should be removed. As in don't delete log handlers that are already in place when `create_logger()` gets called.
- If there is a good reason for this behavior to remain, it should be documented clearly in `app.logger`'s docstring.

I am happy to carry out either (or some other) change and send a pull request.

crazygit:
@bhafichuk-redfall  I completely agree with you. I wasted a good morning figuring out why config logger from configure not work for Flask app instance

kerma:
So what's wrong with the #1311 that it has been hanging here for a year and more? Another soul here who spent hours to figure this out. Now I have to go with dropping `app.logging` usage and adding the ugly hack :(

tinproject:
For the record, @jonblack:

> Using the `LOGGER_NAME` setting helped me to create the correct "top-level" logger, but the default level for subsequent child loggers is being mysteriously set to `WARNING` instead of `NOTSET` (which means, use the parent's level). Here's an example output:
> ```
> myapp -> DEBUG
> myapp.factory -> WARNING
> myapp.measurements -> WARNING
> myapp.measurements.views -> WARNING
> ```

this happen because flask don't set explicitely the level for it's logger, instead _'subclass'_ the logger to return a effective level of `DEBUG` in the case of the level of the logger is `NOSET` (https://github.com/pallets/flask/blob/master/flask/logging.py#L61). 
This allows debug messages to be logged whichever the effective level is inherited from it's parent.  
[...]

PR #2436, opened 2017-07-28
Simplify logging

davidism:
Based on discussion in #2023 and #641, Flask interferes too much with logging. This patch greatly reduces the logging configuration that Flask does.

* Remove `LOGGER_NAME` and `LOGGER_HANDLER_POLICY` configuration.
* `app.logger` is always named `'flask.app'`. Use the `'flask'` namespace to allow other logging from Flask in the future.
* Don't use a `Logger` subclass to override `getEffectiveLevel`. Call `setLevel` when the logger is first accessed, based on `app.debug`, instead of checking `debug` for every message.
* Only one handler with one format is added, instead of different formats for production and development.
* A handler is only added if there are no handlers configured that would handle the logger's `getEffectiveLevel`.
* Handlers are never removed.
* The `default_handler` is accessible so that it can be passed to `removeHandler` in case the user wants to get rid of the default handler after it is configured.
* The proxy stream to `environ['wsgi.errors']` is exposed as `wsgi_errors_stream`, in case the user wants to use it in their own configuration.
* Logging documentation is split from error handling documentation and reorganized to address the key concepts.
* Added documentation for injecting request information into messages.
* Logging tests are split into a separate module.
* Some previously miscategorized tests are correctly organized.


docs/logging.rst
```diff
@@ -0,0 +1,169 @@
...
+
+
+Default Configuration
+`````````````````````
+
+If you do not configure logging yourself, Flask will add a
+:class:`~logging.StreamHandler` automatically. During requests, it will write
+to the stream specified by the WSGI server in ``environ['wsgi.errors']``
+(which is usually :data:`sys.stderr`). Outside a request, it will log to
+:data:`sys.stderr`.
```
brownhead: This is really clear!

docs/logging.rst
```diff
@@ -0,0 +1,169 @@
...
+Removing the Default Handler
+````````````````````````````
+
+If you configured logging after accessing
+:meth:`app.logger <flask.Flask.logger>`, and need to remove the default
+handler, you can import and remove it::
+
+    from flask.logging import default_handler
+
+    app.logger.removeHandler(default_handler)
```
brownhead: Will this code work if `app.logger` hasn't been accessed yet? In other words, can I run this code at any time and know that afterwards I've successfully opted out of Flask's logging?

docs/logging.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    from flask import request
+
+    class RequestFormatter(logging.Formatter):
+        def format(self, record):
+            record.url = request.url
+            record.remote_addr = request.remote_addr
+            return super().format(record)
+
+    mail_handler.setFormatter(RequestFormatter(
```
brownhead: nit: this is referring to the `mail_handler` instance created above, but that doesn't seem obvious if you find yourself reading this section by itself (I skipped the "Email Errors to Admin" section as I was reading through and was confused when I read this). Is it possible to have this snippet affect the default handler flask now exposes instead?

docs/logging.rst
```diff
@@ -0,0 +1,169 @@
...
+---------------
+
+Other libraries may use logging extensively, and you want to see relevant
+messages from those logs too. The simplest way to do this is to add handlers
+to the root logger instead of only the app logger. ::
+
+    from flask.logging import default_handler
+
+    root = logging.getLogger()
+    root.addHandler(default_handler)
```
brownhead: This seems like a really useful section! If `app.logger` is accessed before this code gets called, will `default_handler` be on the root and on `flask.app`, and therefore log messages twice?

docs/logging.rst
```diff
@@ -0,0 +1,169 @@
...
+        logger.addHandler(default_handler)
+        logger.addHandler(mail_handler)
+
+
+Werkzeug
+````````
+
+Werkzeug logs basic request/response information to the ``'werkzeug'`` logger.
+If the root logger has no handlers configured, a
+:class:`~logging.StreamHandler` is added.
```
brownhead: This section seems pretty opaque to me, and I'm not sure what to take away from it :/. What are the implications of this additional root logger, and when will this check for "no handlers configured" be done?

flask/logging.py
```diff
@@ -1,94 +1,69 @@
...
-    WSGI request is in flight we log to wsgi.errors, otherwise this resolves
-    to sys.stderr.
+def wsgi_errors_stream():
+    """Find the most appropriate error stream for the application. If a request
+    is active, log to ``wsgi.errors``, otherwise use ``sys.stderr``.
+
+    If you configure your own :class:`logging.StreamHandler`, you may want to
+    use this for the stream. If you are using file or dict configuration and
+    can't import this directly, you can refer to it as
+    ``ext://flask.logging.wsgi_errors_stream``.
```
brownhead: Naive reader feedback: I don't know what this format is, and it might be good to elaborate on how to use this seemingly-a-URI. Can this be passed to `logging.getLogger` and it'll evaluate it lazily?

flask/logging.py
```diff
@@ -1,94 +1,69 @@
...
+
+def has_level_handler(logger):
+    """Check if there is a handler in the logging chain that will handle the
+    given logger's :meth:`effective level <~logging.Logger.getEffectiveLevel>`.
+    """
+    level = logger.getEffectiveLevel()
+    current = logger
+
+    while current:
+        if any(handler.level <= level for handler in current.handlers):
```
brownhead: These effective levels are tricky to think about 😅 . Had to go back and forth with the docs for awhile. `handler.level` is the _lowest_ level that will be handled by the logger, so this seems right.

flask/logging.py
```diff
@@ -1,94 +1,69 @@
...
-    logger.addHandler(prod_handler)
-
-    # Disable propagation by default
-    logger.propagate = False
+    logger = logging.getLogger('flask.app')
+
+    if app.debug and logger.level == logging.NOTSET:
+        logger.setLevel(logging.DEBUG)
+
+    if not has_level_handler(logger):
```
brownhead: ❤️

flask/logging.py
davidism: https://docs.python.org/3/library/logging.config.html#access-to-external-objects

Note that this module isn't part of the public documentation.

docs/logging.rst
davidism: It says that Werkzeug will add a handler to it's logger, `'werkzeug'`, if there are no handlers configured for the root logger.

docs/logging.rst
davidism: Yes, that's why the docs say to configure your own logging first, above in the Basic Configuration section.

docs/logging.rst
davidism: Yes, `removeHandler` skips handlers that are already removed.

docs/logging.rst
brownhead: Aha, I see. So the second sentence could be "If the root logger has no handlers configured, StreamHandler is added to the werkzeug logger"?

docs/logging.rst
```diff
@@ -0,0 +1,175 @@
...
+care about separately, instead of configuring only the root logger. ::
+
+    for logger in (
+        app.logger,
+        logging.getLogger('sqlalchemy'),
+        logging.getLogger('other_package'),
+    ):
+        logger.addHandler(default_handler)
+        logger.addHandler(mail_handler)
+
```
idjaw: This is fantastic for all the other logging implementations that are bundled in most Flask-based apps.

flask/logging.py
```diff
@@ -1,94 +1,69 @@
...
+    If you configure your own :class:`logging.StreamHandler`, you may want to
+    use this for the stream. If you are using file or dict configuration and
+    can't import this directly, you can refer to it as
+    ``ext://flask.logging.wsgi_errors_stream``.
     """
-    ctx = _request_ctx_stack.top
-    if ctx is not None:
-        return ctx.request.environ['wsgi.errors']
-    return sys.stderr
+    return request.environ['wsgi.errors'] if request else sys.stderr
```
idjaw: Is the expectation here that `wsgi.errors` will be an environment variable that is guaranteed to be available? 

Also, is [this](https://github.com/pallets/flask/pull/2436/files#diff-934bb9540808833c60c87a3f6f62dfe0R47) the area where this logic is tested.

davidism: `wsgi.errors` is part of the WSGI standard. Flask was already using it, I just documented it more visibly.

idjaw:
Great job with this!

davidism:
100% coverage on `logging.py`: https://codecov.io/gh/pallets/flask/pull/2436/src/flask/logging.py?before=flask/logging.py

behrtam:
Am I understanding the new `logging docs` wrong? I was expecting a different log output with the injected information. https://gist.github.com/behrtam/03b484db328bd1e4c95b201871fe6c15

davidism:
You didn't log anything with the app logger.

oribaldi:
Hello guys! This is just what I need :) However, the changes are not in the latest release v0.12.2. Do you know when there will be a new release?

davidism:
This is part of the 1.0 milestone. It will be released once that is complete.

Light-Gao:
So glad many people work on this...
I just experienced Flask for weeks, and I really like it...
I hope I could do more help in this community in the future days...

untitaker:
I am not really sure if the rename of the logger from `__name__` to `flask.app` made sense. previously you were able to have multiple loggers for submodules (`myapp.foo`, `myapp.bar`) and their records would all propagate to `myapp`, which would be the name of the app logger.

davidism:
I know I had a good reason at the time for changing that, but I can't remember what it was. :-/ I guess it should be ok to add the `LOGGER_NAME` config back in.

PR #2438, opened 2017-07-29
Adds support DEBUG variable from config.

chilic:
It's just small fix. 
Which should revert back support DEBUG variable from config instead of OS env.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -335,7 +335,9 @@ class ScriptInfo(object):
                 '"app.py" module was not found in the current directory.'
             )
 
-        debug = get_debug_flag()
+        debug = get_debug_flag(
+            app.config.get('DEBUG')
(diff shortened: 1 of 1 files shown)


flask/cli.py
```diff
@@ -590,7 +592,10 @@ def run_command(info, host, port, reload, debugger, eager_loading,
     """
     from werkzeug.serving import run_simple
 
-    debug = get_debug_flag()
+    debug = get_debug_flag(
+        info.load_app().config.get('DEBUG')
```
ThiefMaster: No. This will break the reloader when loading the app fails e.g. due to a syntax error.

chilic: Hm. Thank you. Will check this again.

chilic:
Need work.

PR #2439, opened 2017-08-01
Add clean command to flask cli

greyli:
The added command `flask clean` can be used to remove all the `.pyc` and `.pyo` file recursively starting at the path to the root of the application. Similar to Flask-Script's `clean` command.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -697,6 +697,28 @@ def routes_command(sort, all_methods):
         click.echo(row.format(rule.endpoint, methods, rule.rule).rstrip())
 
 
+@click.command('clean', short_help='Clean python compile file.')
+@with_appcontext
+def clean_command():
+    """Clean all the .pyc and .pyo files recursively starting at 
+    the path to the root of the application.
+    """
+    from flask.globals import _app_ctx_stack
+    app = _app_ctx_stack.top.app
+    
+    root_path = app.root_path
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Is it really that useful? IMO this is nothing specific to Flask, and if you use a decent shell (like zsh) you can just `rm -r **/*.py[co] **/__pycache__/` to clean up, and e.g. if you tend to change branches a lot something like this could even go into a git post-checkout hook

flask/cli.py
```diff
@@ -697,6 +698,28 @@ def routes_command(sort, all_methods):
...
+    app = _app_ctx_stack.top.app
+    
+    root_path = app.root_path
+    if not root_path:
+        return click.echo('The path to the root of the application was not found, '
+                          'please define it through app.root_path manually.')
+
+    for dirpath, dirnames, filenames in os.walk(app.root_path):
+        for filename in filenames:
+            if filename.endswith('.pyc') or filename.endswith('.pyo'):
```
ThiefMaster: python3 puts these files in a `__pycache__` subdirectory

greyli:
I just thought it's a simple way to clear all these files in different python version or OS. It may help when you rename some files in a large project or reconstruct the application.

greyli:
Ok, may be it's better to keep it as a user-defined command.

PR #2454, opened 2017-08-25
Changes documentation screenshots for a more recent UI

alemangui:
Small cosmetic change for the docs. 

The current screenshots in Flask's docs are 8 years old. These new ones have the same info but with a recent UI (the CSS for the debugger and the browser's UI has changed since).


ThiefMaster:
Thanks!

Just a FYI for future commits; Ideally the first line is shorter than 51 characters, and it should for sure not be so long that even GitHub truncates it. https://chris.beams.io/posts/git-commit/ is quite useful in terms of how to format Git commits.

alemangui:
Sounds good, thanks for the info

PR #2463, opened 2017-09-05
parse cli command's args with the application context

### LINKED ISSUE #2410: Flask CLI doesn't support application context in argument callbacks
I recently converted an old managing script to use Flask's built in CLI instead. I have a table for clients that are identified using a shortname/slug. I thought about using Click's argument callback for this. Given a shortname my function would get an instance of the client from my ORM. The problem is that the application context is not active when the callback is run.

### Expected Behavior

Since many scripts will rely on a database connection it makes sense to have the application context be available in arguments.

```python
import click
from flask import current_app, Flask

app = Flask("demo")

def example_callback(ctx, param, value):
    # Access the application context to trigger an exception
    click.echo(current_app.config)
    return {"id": 1, "shortname": value}

@app.cli.command()
@click.argument("arg", callback=example_callback)
def view(arg):
    click.echo(arg)
```

I run this using `flask view foobar`.

### Actual Behavior

```pytb
  File "/dev/flask-cli-demo/venv/bin/flask", line 11, in <module>
    sys.exit(main())
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/flask/cli.py", line 513, in main
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/flask/cli.py", line 380, in main
    return AppGroup.main(self, *args, **kwargs)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/click/core.py", line 1064, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/click/core.py", line 621, in make_context
    self.parse_args(ctx, args)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/click/core.py", line 880, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/click/core.py", line 1404, in handle_parse_result
    self.callback, ctx, self, value)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/click/core.py", line 78, in invoke_param_callback
    return callback(ctx, param, value)
  File "/dev/flask-cli-demo/app.py", line 8, in example_callback
    click.echo(current_app.config)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/werkzeug/local.py", line 306, in _get_current_object
    return self.__local()
  File "/dev/flask-cli-demo/venv/local/lib/python2.7/site-packages/flask/globals.py", line 51, in _find_app
    raise RuntimeError(_app_ctx_err_msg)
RuntimeError: Working outside of application context.
```

### Environment
```
click==6.7
Flask==0.12.2
itsdangerous==0.24
Jinja2==2.9.6
MarkupSafe==1.0
pkg-resources==0.0.0
Werkzeug==0.12.2
```

### Workaround
The only workaround I've found is doing the same thing `@with_appcontext` does which is:

```python
with ctx.ensure_object(ScriptInfo).load_app().app_context():
```

Using `@with_appcontext` on `example_callback` doesn't work since that gives a warning about using a deprecated function signature (from before Click 2.0).

### Why is this a problem?
This results in two separate instantiations of my ORM. Some ORM's (like the one I'm using) doesn't allow sharing objects across instances (for good reason) which makes the client object I produce unusable anyway.

r-m-n:
this allows to use application object in a callbacks, resolves #2410

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -377,7 +377,9 @@ class AppGroup(click.Group):
         def decorator(f):
             if wrap_for_ctx:
                 f = with_appcontext(f)
-            return click.Group.command(self, *args, **kwargs)(f)
+                cmd = click.Group.command(self, *args, **kwargs)(f)
+                cmd.parse_args = with_appcontext(cmd.parse_args)
+            return cmd
         return decorator
 
     def group(self, *args, **kwargs):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -317,6 +317,13 @@ def test_appgroup(runner):
     def test():
         click.echo(current_app.name)
 
+    @cli.command(with_appcontext=True)
+    @click.argument(
+        "arg", callback=lambda ctx, param, value: current_app.name
+    )
+    def test_callback(arg):
+        click.echo(arg)
+
     @cli.group()
     def subgroup():
         pass
@@ -328,6 +335,10 @@ def test_appgroup(runner):
     obj = ScriptInfo(create_app=lambda info: Flask("testappgroup"))
 
     result = runner.invoke(cli, ['test'], obj=obj)
+    assert result.exit_code == 0
+    assert result.output == 'testappgroup\n'
+
+    result = runner.invoke(cli, ['test_callback', 'arg'], obj=obj)
     assert result.exit_code == 0
     assert result.output == 'testappgroup\n'
 


ThiefMaster:
This looks a lot like it causes an app to be created for every command, which would break commands that need to work without having an app.

davidism:
It's creating an app context for the command, but before that it separately creates one for the args, so this doesn't actually solve the linked issue. The issue is that they want to share the same context for both.

PR #2465, opened 2017-09-06
Mention Flask instead of Jinja in the issue template

jaapz:
For people not really comfortable with the flask ecosystem this can be really confusing.


ThiefMaster:
Thanks! Looks like the mention of Jinja there was a copy&paste screwup ;)

PR #2467, opened 2017-09-07
Add additional information in the url_rule docs

henrycjc:
Small doc change because it is not documented anywhere else and may be confusing. 

If a user is making use of the `@errorhandler(405)` decorator, and they wish to include an Allow header (as per HTTP spec), they may be confused to find that url_rule is None (when they were attempting to get the allowed methods). While an experienced Flask user knows why this won't work, it may be unclear to a new user who is attempting to conform to HTTP. 

This doc change aims to clarify that: because the request was never successfully matched, it was never bound to a URL rule.  In order to access the valid methods, they must examine `routing_exception` instead.

Doc change only, no uncommented code is affected.

### DIFF
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -119,6 +119,10 @@ class Request(RequestBase, JSONMixin):
     #: The internal URL rule that matched the request.  This can be
     #: useful to inspect which methods are allowed for the URL from
     #: a before/after handler (``request.url_rule.methods``) etc.
+    #: Though if the request's method was invalid for the URL rule,
+    #: the valid list is available in ``routing_exception.valid_methods``
+    #: instead (an attribute of the Werkzeug exception :exc:`~werkzeug.exceptions.MethodNotAllowed`)
+    #: because the request was never internally bound. 
     #:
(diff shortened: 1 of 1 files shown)


brad-alexander:
Looks good to me.  👍 

I've seen discussion around this, and though I have never run into this problem myself, it makes sense to me and I feel this is ultimately a great change. Thank you for taking the time to share the fruits of your troubleshooting. 

As an aside coincidentally, I just read this article yesterday: [Matplotlib Lead Developer Explains Why He Can’t Fix the Docs—But You Can](https://www.numfocus.org/blog/matplotlib-lead-developer-explains-why-he-cant-fix-the-docs-but-you-can/) wherein the lead dev provides a call to action to do just this.

>"What I’ve started trying to push as the threshold on docs is, “Did it make it worse?” If it didn’t make it worse, merge it. Frequently, it takes more time to leave a GitHub comment than to fix it."

I wonder if there are some tests we can put around this behavior more for reference and confirmation that it **should** work this way and as a minor benefit - catch it in the unlikely event that the behavior changes in the future. Definitely should not block merging this in.

PR #2477, opened 2017-09-27
Add a concept of application teardown

lalinsky:
The main purpose of this is to have a single place where extensions
can register their cleanup functions. If you are creating many app
instances for testing purposes, you need to keep track of all
individual extensions and release the resources from all of them.
Allowing them to register a cleanup function inside init_app would
hide the complexity and app.close() would be the only thing tests
need to care about.


ThiefMaster:
Any reason why you create many app instances in your tests? Usually you'd use a session-scoped pytest fixture for the app (so it runs only once) and then a function-scoped appcontext fixture so your tests are separated from each other.

lalinsky:
We wanted complete isolation of the app instance, since there are many extensions with external connections registered on it.

davidism:
I understand the use case, and appreciate you making a PR to discuss it. However, I think it's better to keep this in your application.

In most use cases, holding a pool of connections shouldn't matter (not saying it doesn't in your specific case, this is just speaking from my experiences testing). If it does matter, perhaps the extension should be modified instead so that connections are cleaned up on `teardown_appcontext`. You'd still need to be aware of any extensions that didn't use this new hook, so you wouldn't really reduce the work anyway.

Additionally, the WSGI spec does not have any concept of "app closed". So this would never execute when running the server, which will probably confuse developers who try to use it for resource cleanup outside testing.

Issue #2218, opened 2017-03-22
Werkzeug 0.12.0 changed how abort works


aluttik:
[This line in flask.\_\_init\_\_.py](https://github.com/pallets/flask/blob/master/flask/__init__.py#L17) breaks everything now that `werkzeug==0.12.0` was released.

`werkzeug.exceptions.abort` is now a function, so `abort.mapping` throws an error.

[This line in setup.py](https://github.com/pallets/flask/blob/master/setup.py#L74) definitely needs to be updated - preferably with a frozen Werkzeug version so this doesn't happen again.

untitaker:
`abort` was always documented to be a function in the Werkzeug docs. We will definetly not pin Werkzeug to a specific version.

PR #2484, opened 2017-10-06
Allow users to customize the behavior of flask.abort

mkomitee:
Prior to werkzeug 0.12.2, it was possible to manipulate the status code to exception type maps flask used by invoking

    flask.abort.mapping.update(mapping)

This brings that back by creating a flask specific werkzeug.exceptions.Aborter object and exposing it as flask.aborter. It also implements a flask.abort function so that the correct docstring is exposed:

    flask.aborter.mapping.update(mapping)

I've included tests for the new feature which include 100% coverage of the new code branches, and all tests pass (in python 2.7).

This also helps address #2218 without having to lock to the specific version of werkzeug.

### DIFF
--- a/flask/__init__.py
+++ b/flask/__init__.py
@@ -14,7 +14,6 @@ __version__ = '0.13-dev'
 
 # utilities we import from Werkzeug and Jinja2 that are unused
 # in the module but are exported as public interface.
-from werkzeug.exceptions import abort
 from werkzeug.utils import redirect
 from jinja2 import Markup, escape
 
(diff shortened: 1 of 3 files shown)


davidism:
We already had this discussion a while ago, but it was harder to find than I expected. https://github.com/pallets/flask/issues/1837#issuecomment-294289517 I initially had the same idea as you, but reached the conclusion that we should not change this after playing around with it for a while. You can always change the existing aborter the same way, it just lives in Werkzeug instead of Flask.

davidism:
#2350 documents this

davidism:
I think I was going to document this specific pattern in #2350 but then decided not to for some reason. Probably because I would rather people create error subclasses and raise them directly instead.

davidism:
We came back to this years later, #4576 has added an `app.aborter` instance, and `flask.abort` will call it if `current_app` is active.

PR #2486, opened 2017-10-08
Drop unsupported Python 2.6 and 3.3

hugovk:
### 2.6

* Unsupported since [2013-10-29](https://en.wikipedia.org/wiki/CPython#Version_history)
* https://snarky.ca/stop-using-python-2-6/
* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html
* http://www.python3statement.org
* Current pip 9 deprecates Python 2.6 support, pip 10 won't support it (https://github.com/pypa/pip/issues/3955)
* Not much PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

### 3.2

* (Already dropped 👍 )

### 3.3

* Unsupported since [2017-09-29](https://en.wikipedia.org/wiki/CPython#Version_history)
* pip 10 will deprecate Python 3.3 support, pip 11 won't support it (https://github.com/pypa/pip/issues/3796)
* Very little PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

### Also

This prevents this project's dependencies, such as MarkupSafe, from moving forward. See https://github.com/pallets/markupsafe/pull/79#issuecomment-335014965.
* [ ] Changelog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ matrix:
       env: TOXENV=py,codecov
     - python: 3.4
       env: TOXENV=py,codecov
-    - python: 3.3
-      env: TOXENV=py,codecov
     - python: 2.7
       env: TOXENV=py,codecov
     - python: pypy
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ matrix:
       env: TOXENV=py,codecov
     - python: 2.7
       env: TOXENV=py,codecov
-    - python: 2.6
-      env: TOXENV=py,codecov
     - python: pypy
       env: TOXENV=py,codecov
     - python: nightly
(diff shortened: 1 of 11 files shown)


davidism:
Hi, I appreciate your interest, but please stop opening these on so many projects. Maintainers are typically aware of Python 2.6 and 3.3 status and have their own plans for this.

hugovk:
Sure. What are the plans for Flask, Jinja2 and MarkupSafe?

davidism:
They will drop support when I make significant releases. 1.0 will definitely still support them. Opening PRs is premature because there may be merge conflicts due to other commits in the mean time. This is the same reason we don't accept PRs that fix code style everywhere.

PR #2493, opened 2017-10-13
Fix misleading documentation

AMDmi3:
IMO, you would normally interpret "creating and closing database connections all the time" as "creating and closing database connection on each request" as opposed to keeping persistent database connection for a lifetime of an application. However, the documentation is either misleading or "creating and closing database connection on each query" was implied. This, and the term "**application** context" itself makes it look like data placed into flask.g will be available for the following requests, which is not true at all. Let's make the documentation more explicit on this.

### DIFF
--- a/docs/tutorial/dbcon.rst
+++ b/docs/tutorial/dbcon.rst
@@ -8,11 +8,11 @@ Let's continue building our code in the ``flaskr.py`` file.
 
 You currently have a function for establishing a database connection with
 `connect_db`, but by itself, it is not particularly useful.  Creating and
-closing database connections all the time is very inefficient, so you will
-need to keep it around for longer.  Because database connections
-encapsulate a transaction, you will need to make sure that only one
-request at a time uses the connection. An elegant way to do this is by
-utilizing the *application context*.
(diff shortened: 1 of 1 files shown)


docs/tutorial/dbcon.rst
```diff
@@ -8,11 +8,11 @@ Let's continue building our code in the ``flaskr.py`` file.
 
 You currently have a function for establishing a database connection with
 `connect_db`, but by itself, it is not particularly useful.  Creating and
-closing database connections all the time is very inefficient, so you will
-need to keep it around for longer.  Because database connections
-encapsulate a transaction, you will need to make sure that only one
-request at a time uses the connection. An elegant way to do this is by
-utilizing the *application context*.
+closing database connections for each query is very inefficient, so you
+will need to keep it around for the duratiob of a request.  Because database
```
ThiefMaster: *duration

davidism:
Thanks for your feedback. However, I'm in the process of completely rewriting the tutorial, so this PR won't be relevant once that's merged. I'll keep it in mind though.

AMDmi3:
Thank you!

Issue #2495, opened 2017-10-17
Support for Config(None) initialization


dmulter:
### Expected Behavior

It may be an unusual case, but I initialize Config objects just to load remote secure environment variables in addition to my normal app config loading. I've always just initialized with Config(None) under Python 2, but have just switched over to Python 3 and it throws an exception as described below. Note that the file is loaded into a local temp file.

I can work around the problem by putting anything in for root_path, but it would be cleaner if I could pass None for the root_path. I would expect Flask to assume fully qualified paths for files if root_path is not set.

```python
config = Config(None)
with tempfile.NamedTemporaryFile() as s3_file:
    if get_config_file(stage, s3_file.name):
        config.from_pyfile(s3_file.name)
```

### Actual Behavior

The os.path.join() call fails since the root_path is None.

```pytb
  File "/app/config.py", line 28, in _from_s3
    config.from_pyfile(s3_file.name)
  File "/usr/lib/python3.6/site-packages/flask/config.py", line 125, in from_pyfile
    filename = os.path.join(self.root_path, filename)
  File "/usr/lib/python3.6/posixpath.py", line 78, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

### Environment

* Python version: 3.6
* Flask version: 0.12.2
* Werkzeug version: 0.12.2

davidism:
Just pass an empty string.

dmulter:
Yes, I understand passing an empty string for root_path fixes the problem. I'm just suggesting that the code could be backward-compatible with Python2 behavior by continuing to handle the Config(None) style. Thanks for considering!

PR #2506, opened 2017-10-27
add option to use None config for later evaluation

AdamGS:
Allows to create "empty" configuration object in order to initialize it in later during runtime. #2495


davidism:
Thanks for doing this, but I'd rather users explicitly pass an empty string if that's what they want.

PR #2522, opened 2017-11-15
doc: add a pattern doc for Dramatiq

Bogdanp:
This adds an entry to the patterns section for [Dramatiq]. Let me know if there's anything I should change! 

[Dramatiq]: https://dramatiq.io

### DIFF
--- a/docs/patterns/dramatiq.rst
+++ b/docs/patterns/dramatiq.rst
@@ -1,0 +1,83 @@
+Dramatiq Background Tasks
+=========================
+
+Often when processing web requests there are things that can be moved
+out of the request-response cylce in order to conserve time.  An
+example of such a workload is sending an e-mail: not only can it take
+a while, but it can fail so it makes sense to move that sort of
+operation to a background queue where it can be processed
+asynchronously and retried if necessary.
+
+Dramatiq_ is a powerful distributed task processing library for
+Python 3.  This guide will show you how to use Dramatiq and Flask
+together.  We assume you've already read the Dramatiq `user guide`_.
+
+
+Installation
+------------
+
+Dramatiq is a separate Python package and it comes in two flavors::
+
+    $ pip install dramatiq[rabbitmq,watch]
+
+installs it with support for RabbitMQ as a task broker, and::
+
+    $ pip install dramatiq[redis,watch]
+
+installs it with support for Redis as a task broker.
+
+
+Configuration
+-------------
+
+There's nothing Flask-specific to set up, but you do need to tell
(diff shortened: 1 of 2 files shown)


ThiefMaster:
This doesn't seem to be running the tasks inside an app context.

Bogdanp:
You're right, it isn't. I didn't occur to me to do it for such a simple example. I'll add a section on how to automatically wrap tasks inside an app context in a bit.

davidism:
This seems like the exact same as the celery pattern. Why do we need another doc?

wgwz:
There's also the issue of maintenance. Keeping this up to date will fall on the Flask community. Maybe it's better to make a mention of Dramatiq and link to the project documentation specific to Flask?

Bogdanp:
I've added info about how to set up the app context.

> This seems like the exact same as the celery pattern. Why do we need another doc?

They are indeed very close. I figured this'd be useful to someone trying to decide what to use, but if it isn't that's OK too. :)

> There's also the issue of maintenance. Keeping this up to date will fall on the Flask community. Maybe it's better to make a mention of Dramatiq and link to the project documentation specific to Flask?

@wgwz The APIs documented here are unlikely to change, but you make a valid point. I'd be happy to maintain it though I realise that doesn't mean much from a first-time contributor. What would be a good place to put that mention if we go that route?

I also forgot to leave a disclaimer in my OP, sorry about that! I am the creator of Dramatiq.

wgwz:
The current [pattern](http://flask.pocoo.org/docs/0.12/patterns/celery/) could be made somewhat tool agnostic. Then it could mention "here are different ways to handle background tasks". It could even start from threading like is done [here](https://blog.miguelgrinberg.com/post/the-flask-mega-tutorial-part-xi-email-support) for emails, and extending into more advanced cases. I think the Celery snippet is important to keep, because it is widely used and the Celery docs don't have anything specific to Flask. So in this case you could add a link potentially as "Here are more tools for task handling". 

But I'm not really a regular contributor either, so we will have to see what others think. :)

lepture:
It is a burden for Flask community to maintain this piece of documentation.

lepture:
@mitsuhiko @ThiefMaster @untitaker @davidism 

Close it for now. If you find it is useful, you can reopen it.

davidism:
I just want to say, I do like that there are more queue solutions out there. Celery definitely needs competition!

lepture:
I'd like @Bogdanp to keep the documentation on his own project. And Flask adds a link ref to his documentation.

Issue #2535, opened 2017-11-30
Update PyCharm instructions


davidism:
2017.3 added the ability to run modules instead of needing to find the `flask` executable in the virtualenv. The instructions in `cli.rst` should be updated to show selecting "Module name:" and then typing "flask".

cc @doobeh

PR #2541, opened 2017-12-03
update pycharm instructions

nsiregar:
Update Pycharm instructions to run via module name #2535


Tafkas:
I ran the `pycharm-runconfig.png` through https://tinypng.com/ and got the size down to 18KB (-67%)without any noticeable decrease in quality. It's probably for the best to keep images as small as possible.

davidism:
Thanks for making this. PyCharm changed their configuration again to be able to specify modules instead of files. I'm going to close this and write the instructions for the new version.

davidism:
I created a smaller (dimensions) image then optimized it with the tinypng link from @Tafkas. I also reworded some of the docs to reflect the current CLI.

PR #2542, opened 2017-12-05
Documentation fix - quickstart "HTTP Methods" code example

vkroz:
Minor documentation fix: 
`return` is required to make 'http methods' example work


ThiefMaster:
I think that's more pseudocode-like. Anyway, can't hurt to be explicit.

PR #2553, opened 2017-12-14
Docs: Add missing backslash

greyli:


davidism:
Paths only have one backslash, you're confusing that with a Python string.

greyli:
@davidism The backslashes added are used to escape the origin one.

Check [here](http://flask.pocoo.org/docs/dev/cli/#environment-variables-from-virtualenv).

davidism:
OK, please explain why you're making a change like that in the future.

greyli:
Okay, sorry about that.

Issue #2550, opened 2017-12-13
Some of tests fail on running pytest


djmgit:
### Expected Behavior

On running ```pytest``` all tests should pass.

To reproduce please execute ``` pytest ```
### Actual Behavior

5 tests fail 

Here is the full output of pytest: https://paste.ubuntu.com/26177622/

### Environment

* Python version: 3.5.2

djmgit:
I would like to work on this, any pointers regarding from where to start would be of great help.
I have build Flask from master branch.

davidism:
Right, I noticed this too. A recent PyTest update added automatic capture of log messages. To do this, I think they add a logging handler. Flask only adds a logging handler if there are no loggers already registered, so it fails to add its default logger.

The ideal fix here is to disable logging capture for these tests, although I haven't looked into how PyTest handles this yet.

camilojimenez:
I tracked the issue to this release of pytest

https://github.com/pytest-dev/pytest/blob/6cdd85122705393cfcfe08570bae581bf9f2d157/CHANGELOG.rst

According to the changelog:

- Pytest now captures and displays output from the standard logging module. The user can control the logging level to be captured by specifying options in pytest.ini, the command line and also during individual tests using markers. Also, a caplog fixture is available that enables users to test the captured log during specific tests (similar to capsys for example). For more information, please see the logging docs. This feature was introduced by merging the popular pytest-catchlog plugin, thanks to Thomas Hisch. Be advised that during the merging the backward compatibility interface with the defunct pytest-capturelog has been dropped. (#2794)

The actual logging code is here

https://github.com/pytest-dev/pytest/blob/6cdd85122705393cfcfe08570bae581bf9f2d157/_pytest/logging.py#L259

I skimmed over the logging code and could not find any way to disable logging capture in a documented way.

What are the thoughts about

- Pinning the pytest version to 3.2.5
- Updating the test to support this
- Write a setup/teardown method to remove any existent loggers

davidism:
I would go with option three. It should work out similar to some other tests in that module that modify the handlers.

camilojimenez:
Does this seem like an appropriate solution?

~~~~
@pytest.fixture
def no_logging_handlers(app):
    logging.root.handlers[:] = []
    app.logger

def test_logger(app, no_logging_handlers):
    assert app.logger.name == 'flask.app'
    assert app.logger.level == logging.NOTSET
    assert app.logger.handlers == [default_handler]
~~~~

[...]

PR #2554, opened 2017-12-14
fix non-passing tests for logging on pytest >= 3.3.0

camilojimenez:
This fixes issue #2550, where some tests fail when using newer pytest (after version 3.2.5).

The reason for this issue is that pytest added a plugin that instantiates a handler for the logging module, while the tests were expecting no handlers.

The PR solves the issue by temporarily disabling the logging module and removing any handlers for all tests on the file

### DIFF
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -17,6 +17,10 @@ def reset_logging(monkeypatch):
     logger.handlers = []
     logger.setLevel(logging.NOTSET)
 
+    logging_plugin = pytest.config.pluginmanager.getplugin('logging-plugin')
+    pytest.config.pluginmanager.unregister(name='logging-plugin')
+    logging.root.handlers = []
+
     yield
 
     logging.root.handlers[:] = root_handlers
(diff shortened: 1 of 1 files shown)


davidism:
Since it's also possible to disable the plug-in in other ways, it would be better to only restore it if it was actually enabled to begin with.

davidism:
You can ignore the failures on 2.6 and 3.3.

lepture:
@davidism are we going to remove the support for 2.6 and 3.3?

ThiefMaster:
Yes

davidism:
Yes. First step is removing the tests from CI, I just haven't gotten around to it for the other Pallets projects yet.

camilojimenez:
@davidism done, now it only adds it back if it was set to begin with

davidism:
Thanks for figuring this out. I did a bit of refactoring: use the `pytestconfig` fixture instead of accessing `pytest.config`, and `unregister` returns the plugin.

PR #2555, opened 2017-12-14
remove tests and docs for python 2.6 and 3.3

davidism:
pallets/meta#24

I'll test 2.6 and 3.3 before I release 1.0, which will be the last version to support these.


ThiefMaster:
Do we really want to include 2.6/3.3 in the 1.0 release?

davidism:
I'm OK with removing them completely. When I said "1.0 will be the last version" I meant even for maintenance releases after that, so it's mostly just a last concession.

davidism:
If you think it's OK, I'd rather just completely stop supporting 2.6 and 3.3 right now.

ThiefMaster:
I'd do one last release before 1.0 and then drop it with 1.0

davidism:
I'm not going to make a 0.13 release. The only thing blocking me from releasing 1.0 is finishing the tutorial. If we make a new release now, 1.0 will have nothing in it.

ThiefMaster:
maybe the last release is fine for forever-yesterdays who still use 2.6 ;)

davidism:
Yeah, I'm fine with that. That's basically what this would be, 1.0 just "secretly" still supports 2.6 if someone just happens to be able to upgrade Flask without upgrading Python.

lepture:
I'm fine with that. If people want to use Python 2.6, they can still use the old version of Flask.

PR #2557, opened 2017-12-15
cli.rst: added flask run cli options help

javabrett:
When browsing the CLI doc, I found myself wanting to know how to set host bind interface and port.  These are not currently discussed, and are revealed with `flask run --help`.

Added that output to the page.


davidism:
Thanks, I didn't want to include the output of every command's help because that page is already long enough. I already mention using `--help` at the top.

> The `--help` option will give more information about any commands and options.

javabrett:
@davidism agreed, but sometimes we read docs offline and away from the runtime.  I tried again with a minimal mention of these important options that doesn't just include the output of `--help`.

PR #2558, opened 2017-12-16
add appveyor configuration

davidism:


davidism:
Need to get my Windows box back up and fix the errors before merging.

davidism:
So apparently no one is using Flask sessions on Python 2.7 on Windows, otherwise someone would have reported that Python 2.7 doesn't provide `inet_pton` on Windows.

davidism:
Given that we'd have to add a dependency on `win_inet_pton` just to support it, I'm inclined to just omit the one check that uses it (session cookie domain) on Py27.

davidism:
Changed `is_ip` to only check IPv4 with `inet_aton` on Python 2 on Windows.

PR #2560, opened 2017-12-17
Add conftest to flaskr

jakerobers:
In practice, it is often important to have more than one test file. Though this is in the pytest domain, I thought that it would be important to illustrate the capability of fixture sharing.

I updated the docs and flaskr code example to use a `conftest.py`file.

### DIFF
--- a/examples/flaskr/tests/conftest.py
+++ b/examples/flaskr/tests/conftest.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""
+    Flaskr conftest
+    ~~~~~~~~~~~~
+
+    Defines fixtures for the Flaskr test suite.
+
+    :copyright: (c) 2015 by Armin Ronacher.
+    :license: BSD, see LICENSE for more details.
+"""
+import os
+import tempfile
+import pytest
+from flaskr.factory import create_app
+from flaskr.blueprints.flaskr import init_db
+
+
+@pytest.fixture
+def app(request):
+
+    db_fd, temp_db_location = tempfile.mkstemp()
+    config = {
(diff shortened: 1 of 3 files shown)


examples/flaskr/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+
+@pytest.fixture
+def client(request, app):
+
+    client = app.test_client()
+
+    def teardown():
+        os.close(app.config['DB_FD'])
+        os.unlink(app.config['DATABASE'])
+    request.addfinalizer(teardown)
```
ThiefMaster: this is deprecated in favor of yielding between init and cleanup code

jakerobers: Okay. I will fix this.

davidism:
The more we add to this thing the less it's a simple example that beginners can look at and understand. 😞

davidism:
Not anything against your pr, just a general observation.

jakerobers:
I understand.

Would there be a better location for this? Maybe we can display a link instead.

jakerobers:
I am going to close this for now and give it some more thought. Maybe we can explain conftest in a simpler manner.:

At the very least, I will submit another PR request to fix the issue that @ThiefMaster pointed out with the existing Flaskr example:
https://github.com/pallets/flask/blob/master/examples/flaskr/tests/test_flaskr.py#L44

davidism:
I think this can go in the `testing` section outside the tutorial, but explaining all the options in PyTest ultimately belongs in PyTest's docs. I am working on creating a "tutorial" example and "full" example that takes the tutorial farther.

PR #2561, opened 2017-12-18
Updated cli.rst with a minimal mention of --host and --port options.

javabrett:
Added a (minimal) mention of `--host` and `--port` `flask run` options.  Replaces #2557.


davidism:
Given the dev server shouldn't be used externally, I want to emphasize these options the least.

javabrett:
Right, that's why I suggested it right above the don't-use-in-production-warning.  I find there are enough development situations where you need to bind to a specific adapter to get everything working.  But given I eventually got-back to my desk after doc-reading and found the options, I guess others can too.

davidism:
The docs do need work, however, if you want to contribute. See the [1.0 milestone](https://github.com/pallets/flask/milestone/2).

PR #2588, opened 2018-01-08
Remove the imported app module from sys.modules on SyntaxError

### LINKED ISSUE #2423: Reloader fails to reload after SyntaxError when the app is in an __init__.py file
### Expected Behavior

The reloader should work correctly even for an app in a package's `__init__.py` file.

In `hello/__init__.py`:
```python
from flask import Flask
app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello, World!'
```

```cmd
set FLASK_APP=hello
set FLASK_DEBUG=1
flask run
```

Than edit `__init__.py` and introduce a syntax error.

### Actual Behavior

I get the SyntaxError the first time I try to browse and afterwards get:

```pytb
  File "flask\cli.py", line 48, in find_best_app
    'using a factory function.' % modules.__name__)
NoAppException: Failed to find application in module "hello". Are you sure it contains a Flask application? Maybe you wrapped it in WSGI middleware or you are using a factory function.
```

### Cause

This is caused by weird CPython behavior which might be a bug:
```py
>>> __import__('syntax_error')
...
SyntaxError: ...
>>> a = __import__('syntax_error')  # woot it works?!
>>> a.__file__
'<path to the directory containing __init__.py>'  # Instead of path to __init__.py
```

After the `SyntaxError` in the app, the Wekzeug reloader won't know about the `__init__.py` file any more since it iterates `sys.modules` and the entry there contains the path to the directory instead of the path to the `__init__.py` file, and Flask will call `__import__` repeatedly but will get the weird empty module and will fail to find an app in it, leading to the described exception.

### Environment

* Python version: 2.7.13
* Flask version: 0.12.2
* Werkzeug version: 0.12.2

Not sure if this happens in Python 3.

segevfiner:
When a SyntaxError occurrs in a package's `__init__.py` file, Python seems to leave an empty module in `sys.modules`. This will result in any request after the first to give you a `NoAppException` instead of the `SyntaxError` again.

I'm not 100% sure this doesn't have any side effects... It's an hack... So please review and tell me what you think about this. (Maybe more conditions for triggering this?)

Fixes #2423 

### Sample
In `hello/__init__.py`:
```python
from flask import Flask
app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello, World!'
```

```cmd
set FLASK_APP=hello
set FLASK_DEBUG=1
flask run
```

Than edit `__init__.py` and introduce a syntax error.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -215,6 +215,11 @@ def locate_app(script_info, module_name, app_name, raise_if_not_found=True):
             )
         else:
             return
+    except SyntaxError:
+        if module_name in sys.modules:
+            del sys.modules[module_name]
+        raise
+
 
     module = sys.modules[module_name]
 
--- a/tests/test_apps/cliapp/syntaxerror/__init__.py
+++ b/tests/test_apps/cliapp/syntaxerror/__init__.py
@@ -1,0 +1,1 @@
+spam()spam()
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -243,6 +243,15 @@ def test_locate_app_suppress_raise():
         )
 
 
+def test_locate_app_init_syntax_error(test_apps):
+    info = ScriptInfo()
+    with pytest.raises(SyntaxError):
+        locate_app(info, 'cliapp.syntaxerror', None)
+
+    with pytest.raises(SyntaxError):
+        locate_app(info, 'cliapp.syntaxerror', None)
+
+
 def test_get_version(test_apps, capsys):
     """Test of get_version."""
     from flask import __version__ as flask_ver


davidism:
Is there an easy way to write a test for this? Otherwise, seems fine to me.

mitsuhiko:
This does not just apply for packages, it affects a boatload of situations and this only catches on case. I'm not sure what the best approach here is.

mitsuhiko:
One way would be to patch `__import__` when the reloader is active and catch down `ImportError`s and clean up recursively.

segevfiner:
Hmm. I see... This won't catch sub-packages that fail to import on `SyntaxError`, only the main app package. As such, they will be left empty, and on the next request the app will get them as an empty module. Flask won't fail to find the app, but the app will fail to find stuff in that package...

Seriously that Python behavior is so off...

davidism:
Maybe modify the Werkzeug reloader to record the loaded modules before starting the reloader, then reset them before triggering a reload.

davidism:
I can't reproduce this on Python 2.7 or Python 3.6. I tried starting with valid syntax then editing in an error, and vice versa.

davidism:
The test you added passes without your patch.

segevfiner:
Weird. It didn't pass without the patch for me... I wonder what is
different...

I tested on Python 2.7.14 32-bit, running on Windows 10 x64 (Latest ATM),
inside a virtualenv that contains only Flask & Werkzeug from Git and their
dependencies.

Another idea is that maybe that weird behavior isn't in core Python and is
triggered by something else... At that point the bug will be in that code
and not in Flask. (Some import machinery hook...? That fails to delete the
module from sys.modules on SyntaxError)

I will try to reproduce and figure this out later when I have the time. (I
will try to also post more exact environment details)

בתאריך 9 בינו' 2018 17:21,‏ "David Lord" <notifications@github.com> כתב:

davidism:
I've had developers try to reproduce this on Windows, Mac, and Linux with and without the Watchdog reloader. It works for all of them.

davidism:
The normal Python importer is careful about cleaning up after failed imports. It's likely that something else in your stack is affecting this. I'm going to close this for now, but if you can figure out what's going on I can reopen it later.

mitsuhiko:
@davidism there are long running bugs in the interpreter where cleanup does not correctly happen. It's a limitation of the python importer.

segevfiner:
I went with addressing the underlying issue in CPython https://github.com/python/cpython/pull/5142.

davidism:
I'm not doubting that the error happens for you, but I'm curious why no one I asked was able to reproduce this in a variety of environments. I think getting cpython to fix it is the right course of action.

segevfiner:
@davidism Yeah I got curious about that too. Which is why I went ahead and debugged it. The bug is obviously in 2.7 but *not* in 3.6. I literally went in with a debugger and saw the exact flow of code that causes this. It's hard to tell why they weren't able to reproduce it though.

PR #2603, opened 2018-01-22
Update cli help for FLASK_ENV

greyli:
This PR replace FLASK_DEBUG with FLASK_ENV in command and change the cli help string.


davidism:
Thanks. I changed it back to only mentioning `wsgi.py`, as that was deliberate.

Issue #2587, opened 2018-01-07
Document SameSite cookie and add session support


davidism:
Werkzeug added support, need to update the security docs to reflect this. Add a new config to set the value for the session cookie, possibly default it to true. ref #2030

PR #2607, opened 2018-01-23
Added support for cookie SameSite attribute

FadhelC:
New Feature: Support for the cookie's SameSite attribute.

Support for cookie's SameSite attribute was lately added to Werkzeug 0.14.0., so implementing this feature in Flask.
simply set the SESSION_COOKIE_SAMESITE to "Strict" or "Lax", and your cookie will have the SameSite attribute set.
Please check https://www.owasp.org/index.php/SameSite for more info.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -210,11 +210,13 @@ The following configuration values are used internally by Flask:
 
 .. py:data:: SESSION_COOKIE_SAMESITE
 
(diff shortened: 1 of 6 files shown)


davidism:
Thanks! closes #2587

davidism:
We're ok with bumping all the dependency versions, so Werkzeug 0.14 is ok here.

Issue #2604, opened 2018-01-22
Flask shell: Give readline a completer that is aware of shell context


CookiesBestDinner:
### Problem
`cli.py:shell_command` (flask shell) creates a context and passes it to `code.interact`
This leaves tab completion broken:

```
>>> stuff = 5
>>> stuTABTABTABTAB <- no match
```
### Suggestion
This can be remedied by supplying a completer to readline:

```
>>> import readline
>>> import rlcompleter
>>> readline.set_completer(rlcompleter.Completer(globals()).complete)
>>> stuTAB <- stuff! yay!
```

Or perhaps `flask/__main__.py` could be used as the context:

*file: flask/cli.py*
```
...
def shell_command():
...
    import __main__
    ctx = __main__.__dict__
...
```

davidism:
This requires setting up an `InteractiveConsole` instead of using the `code.interact` shortcut. Then we can call `runcode` on the interpreter before calling `interact` on it. If possible, check if the completer is already set up.

CookiesBestDinner:
Pardon me, what's wrong/missing with

```
readline.set_completer(rlcompleter.Completer(globals()).complete)
```
?
It "works for me", but what am I missing?

davidism:
I was under the impression it had to be called from within the new shell. It looks like you've already done all the work, why not make a PR?

CookiesBestDinner:
Is this sufficient indication of that it is the full context?


```
rlcompleter.holdontothisplease = ctx
```
```
$ flask shell
>>> import rlcompleter
>>> rlcompleter.holdontothisplease is globals()
True
```

davidism:
I've been looking into this, this is not enough to match how the standard shell behaves. It looks like `venv` messes with `site.py` and doesn't set up the completer the same way as the default `site.py`.

davidism:
Essentially need to copy Python's `site.enablerlcompleter`.

https://github.com/python/cpython/blob/22feeb88b473b288950cdb2f6c5d28692274b5f9/Lib/site.py#L401-L449

davidism:
Looks like `venv` makes `sys.__interactivehook__` available and `code.interact` just doesn't call it like the standard shell. But when using `virtualenv` or `mkvirtualenv` it's undefined.

davidism:
Unfortunately, `sys.__interactivehook__` doesn't pick up the added context, so it's not available for completion.

CookiesBestDinner:
[...]

PR #2608, opened 2018-01-24
2604 flask shell tab completion

CookiesBestDinner:
This pull request makes tab completion available in the flask shell, and sets `__name__` and `__doc__` so that those do not fall back to their counterparts in the builtins module (`code.InteractiveInterpreter` sets those for its default context, so I figure flask shell should also do this.)

I've set `__name__` to `'__flaskshell__'`, I don't really know if that's what it "should" be, it does provide a way for PYTHONSTARTUP to detect what's going on though. My next suggestion would be `'__console__'` which is what `code.InteractiveInterpreter` uses by default.

I've got roughly half a clue what I'm doing here, do tell if something should be different.

Link to issue:
https://github.com/pallets/flask/issues/2604

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -752,7 +752,7 @@ def shell_command():
         app.env,
         app.instance_path,
     )
-    ctx = {}
+    ctx = {'__doc__': None, '__name__': '__flaskshell__'}
 
     # Support the regular Python interpreter startup script if someone
     # is using it.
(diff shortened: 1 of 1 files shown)


davidism:
You also need to call `readline.parse_and_bind('tab: complete')` otherwise it still doesn't complete on tab.

davidism:
This fails if `readline` is not available.

davidism:
Thanks for working on this. Closing after discussion in #2604.

PR #2609, opened 2018-01-28
allow nightly to fail on travis

davidism:
For some reason nightly just fails with "invocation error" on Travis. May be related to pytest-dev/pytest#3011.


davidism:
Going to check on my local machine just to make sure I can't reproduce this with a current alpha build.

davidism:
Works fine on my machine. Guess there's something wrong with Travis's build.

PR #2610, opened 2018-01-28
add note about syntax errors in custom scripts

### LINKED ISSUE #2424: The reloader fails on SyntaxError when using a custom cli entry point
### Expected Behavior

The reloader should keep working even on SyntaxError when using a custom cli entry point. Obviously it might be diffcult to make it work as is, so it might require the cli entry point to be written in a certain way for this to work. For example it should be in a separate file that is not modified often during app development and should import the app only from the `create_app` function. This might just require documenting this more clearly.

```python
import click
from flask import Flask
from flask.cli import FlaskGroup

app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello, World!'

@click.group(cls=FlaskGroup, create_app=lambda info: app)
def cli():
    pass

if __name__ == '__main__':
    cli()
```

```cmd
set FLASK_DEBUG=1
python app.py run
```

### Actual Behavior

Introucing a SyntaxError into the file will cause the reloader to exit.

### Environment

* Python version: 2.7.13
* Flask version: 0.12.2
* Werkzeug version: 0.12.2

davidism:
closes #2424 

If you use a custom entry point and introduce a module-level error, the reloader can't recover. Mention this as a reason to use the `flask` command.

Unrelated, mention which version of PyCharm adds Flask support.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -355,15 +355,25 @@ script is available. Note that you don't need to set ``FLASK_APP``. ::
     $ pip install -e .
     $ wiki run
 
+.. admonition:: Errors in Custom Scripts
+
+    When using a custom script, if you introduce an error in your
+    module-level code, the reloader will fail because it can no longer
+    load the entry point.
+
+    The ``flask`` command, being separate from your code, does not have
+    this issue and is recommended in most cases.
+
 .. _console script: https://packaging.python.org/tutorials/distributing-packages/#console-scripts
 
 
 PyCharm Integration
 -------------------
 
-The Flask CLI features aren't yet fully integrated into the PyCharm IDE,
-so we have to do a few tweaks to get them working smoothly. These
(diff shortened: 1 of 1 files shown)


ThiefMaster:
For custom entry points you could mention that there is one good way to avoid unrecoverable reloader crashes:

- have imports for the cli entrypoint without side effects (e.g. `yourpackage.cli`, with `yourpackage/__init__.py` being empty-ish and `yourpackage.cli` having no global imports of other application stuff or anything else not absolutely necessary)
- use another package/module like `yourpackage_cli` as the entry point for the CLI, so any error when importing in `yourpackage` will not happen at CLI import time

davidism:
I thought about it, but felt like it was getting too complicated. I figured "flask, being separate from your code" was enough of a hint. Maybe adding a custom entry point example to `examples` would be easier.

PR #2614, opened 2018-01-31
2613 avoid code repetition

VishvajitP:
This code change removes the code repetition to calculate the list length.

It finds the list length at the start of the code and assigns the length to a variable so that this variable can be used in the subsequent requirements.


ThiefMaster:
Loading the config usually happens once, caching the value provides no real benefit.

Also, the commits in the PR are quite messed up. A change like this should be one single commit, rebased to master. Anyway, in this particular case I don't think it's worth changing due to the reason above.

PR #2622, opened 2018-02-06
bump dependency versions

davidism:
closes #2586

PR #2623, opened 2018-02-07
Remove tojson filter

greyli:
Jinja2 added the `tojson` filter in [version 2.9](https://github.com/pallets/jinja/blob/31f92b5947e9dba8ba997dd93921cbf1538bb79f/CHANGES#L163). Now that we [bumped the dependency's versions](https://github.com/pallets/flask/pull/2622) to the latest, maybe we should delete it in Flask.


ThiefMaster:
I somehow doubt that this logic still runs after your change: https://github.com/greyli/flask/blob/master/flask/json/__init__.py#L84-L100

ThiefMaster:
No, what I mean is that your PR changes the behavior:

- `tojson_filter` calls `htmlsafe_dumps`
- `htmlsafe_dumps` calls `htmlsafe_dumps`
- `htmlsafe_dumps` calls `dumps`
- `dumps` calls `_dump_arg_defaults`, which is the function I mentioned in my last comment.

But the [Jinja version](https://github.com/pallets/jinja/blob/master/jinja2/filters.py#L1047) of the filter does not do this! I see two options here:

- Close this PR and leave it as-is - i.e. let flask replace the builtin Jinja filter as is the case right now, which works fine
- Populate the `json.dumps_function` and `json.dumps_kwargs` policies of Jinja when setting up the Jinja environment in Flask.

But since `_dump_arg_defaults` depends on application state (e.g. the blueprint handling the current request) the second option is not really feasible without very nasty hacks, so I think we'll just keep the current implementation.

PR #2629, opened 2018-02-12
prevent double slash /a//b when blueprint prefix ends with slash

zcchen:
Simple example file as below:

```
from __future__ import print_function

from flask import Flask, Blueprint, redirect

bp = Blueprint('bp', __name__)
app = Flask(__name__)

@bp.route('/b/')
def tmp():
    return "URI should be '/a/b/"


@app.route('/')
def index():
    return redirect("/a/b/")

app.register_blueprint(bp, url_prefix='/a/')

if __name__ == "__main__":
    print("app.url_map: ", app.url_map)
    app.run(port=5000, host='127.0.0.1', debug=True)
```

With the old code, there is one rule like this `<Rule '/a//b/' (HEAD, OPTIONS, GET) ->`.
With the new code, that rule would be change to `<Rule '/a/b/' (HEAD, OPTIONS, GET) ->`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -143,6 +143,8 @@ unreleased
   :data:`SERVER_NAME` does not implicily enable it. It can be enabled by
   passing ``subdomain_matching=True`` to the ``Flask`` constructor.
   (`#2635`_)
+- A single trailing slash is stripped from the blueprint ``url_prefix``
+  when it is registered with the app. (`#2629`_)
 
 .. _pallets/meta#24: https://github.com/pallets/meta/issues/24
 .. _#1421: https://github.com/pallets/flask/issues/1421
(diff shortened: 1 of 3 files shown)


ThiefMaster:
I think people should rather not add trailing slashes to their `url_prefix` ;)

Maybe showing a warning would be better instead of removing consecutive slashes so people can actually fix their code?

PS: The commit message is a bit too long (even GitHub truncates it), first line should be 51 chars max.

zcchen:
Maybe someone would use `url_prefix = '/'`?
At that time, bp.route would also have a rule like '/' .

ThiefMaster:
`url_prefix='/'` seems to be quite pointless, isn't it?

zcchen:
Yes.
But should we clear the extra and useless slashes in the URI?

davidism:
Related to pallets/werkzeug#1132. I'd rather show a warning for prefixes that end with `/` in Flask, and focus on the Werkzeug issue.

davidism:
This should probably just strip exactly one trailing slash off the prefix. I don't want to mess with route definitions in general, just the prefixing. If pallets/werkzeug#1132 is implemented, it will handle normalizing routes in general.

anowlcalledjosh:
I agree that some kind of feature to either strip a trailing slash or to warn that it's not necessary should be added – I recently was trying to work out why some things were not working properly when using `url_prefix="/"`, since that seemed like the obvious thing to do to explicitly register a blueprint at the root of the app.

PR #2633, opened 2018-02-18
Fix the README to reflect locate_app() in flask/cli.py

paulrigor:
Fix the README to reflect locate_app() in flask/cli.py


davidism:
Thanks, but this doesn't change anything, both are valid. Also, I'm currently rewriting the entire tutorial and example, so this will be overwritten.

PR #2641, opened 2018-02-26
Refactor test codes to PEP8 recommendation

JoMingyu:
- '== cond' to 'is cond'
ex) app.templates_auto_reload == False to app.templates_auto_reload is False
- Change about blank lines
- Change field variable names intuitively
- Change about blank lines
- Change about continuation line's visual indent
- Change about importation

### DIFF
--- a/tests/test_apps/blueprintapp/__init__.py
+++ b/tests/test_apps/blueprintapp/__init__.py
@@ -1,8 +1,8 @@
 from flask import Flask
+from blueprintapp.apps.admin import admin
+from blueprintapp.apps.frontend import frontend
 
 app = Flask(__name__)
 app.config['DEBUG'] = True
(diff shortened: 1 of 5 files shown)


ThiefMaster:
- PEP8 actually discourages using `is` (or `==`) for boolean checks; these operators are only useful there if you really need to check the type as well. But FWIW, I think that if in this particular place we really care if it's literal False/True (ie if it could be None as well), using `is` is nicer even though it's not what PEP8 recommends ;)
- Your rename from `d` to `data` results in mis-indented continuation lines. Also, I think there is absolutely nothing wrong with using short names in unit tests, especially if their purpose is obvious.

While probably having good intentions, PRs like this not are not very helpful - they mostly add noise and this kind of refactoring is usually done in one go by one of the developers.

PR #2648, opened 2018-02-28
Change PROPAGATE_EXCEPTIONS config default value to False

decaz:
I suggest to change `PROPAGATE_EXCEPTIONS` config default value to `False` for more appropriate behavior at production environment: user should implement error handler for `500` and that's all.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -129,7 +129,10 @@ The following configuration values are used internally by Flask:
     handlers. If not set, this is implicitly true if ``TESTING`` or ``DEBUG``
     is enabled.
 
-    Default: ``None``
+    Default: ``False``
+
+    .. versionchanged:: 1.0
+        Default value is ``False`` now.
 
 .. py:data:: PRESERVE_CONTEXT_ON_EXCEPTION
 
--- a/flask/app.py
+++ b/flask/app.py
@@ -280,7 +280,7 @@ class Flask(_PackageBoundObject):
         'ENV':                                  None,
         'DEBUG':                                None,
         'TESTING':                              False,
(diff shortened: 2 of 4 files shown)


ThiefMaster:
http://flask.pocoo.org/docs/0.12/config/#builtin-configuration-values

> explicitly enable or disable the propagation of exceptions. If not set or explicitly set to None this is implicitly true if either TESTING or DEBUG is true.

Nothing to change here ;)

decaz:
The idea was to get rid of `PROPAGATE_EXCEPTIONS` change requirement (it's the only configuration option which requires to be assigned explicitly).

davidism:
If you're in a production environment, neither `DEBUG` or `TESTING` is true, so you don't have to do anything.

decaz:
@davidism If I'm in a production I should change it to `False` otherwise my `500` exception handler won't catch any exception.

ThiefMaster:
If you need to set it to False in production it means you have DEBUG or TESTING enabled, both of which should always be False in production.

decaz:
@ThiefMaster I understand why it returns True, but I don't understand why I should explicitly set DEBUG and TESTING to False in a production. Is not should be their default behavior?

ThiefMaster:
You don't need to set them, they are False by default.

decaz:
@ThiefMaster that's true for `TESTING` but `DEBUG` is `None` by default: https://github.com/pallets/flask/blob/master/flask/app.py#L281

ThiefMaster:
None is falsy as well.

decaz:
Of course. Had a look at `Flask.propagate_exceptions` again - now it's clear. Thanks!

PR #2650, opened 2018-03-05
Docs: Add example for 'any' route converter

greyli:


davidism:
Thanks, but I don't think this is necessary here. I thought we had rewritten this section so the converters linked to their documentation.

PR #2662, opened 2018-03-18
Add tips for 413 connection reset issue

greyli:
When uploading files exceed the limit set by `MAX_CONTENT_LENGTH`, the flask development server will reset the connection instead of return a 413 response. This behavior confused a lot of people:
- https://stackoverflow.com/questions/7371660/flask-resets-the-connection-instead-of-returning-a-413-when-the-file-is-to-big
- https://stackoverflow.com/questions/19911106/flask-file-upload-limit
- https://stackoverflow.com/questions/30414696/upload-file-larger-than-max-content-length-in-flask-results-connection-reset

### DIFF
--- a/docs/patterns/fileuploads.rst
+++ b/docs/patterns/fileuploads.rst
@@ -153,6 +153,12 @@ The code above will limit the maximum allowed payload to 16 megabytes.
 If a larger file is transmitted, Flask will raise a
 :exc:`~werkzeug.exceptions.RequestEntityTooLarge` exception.
 
+.. admonition:: Connection Reset Issue
+
+    When using the local development server, you may get a connection
+    reset error instead of a 413 response. You will get the correct
+    status response when running the app with a production WSGI server.
+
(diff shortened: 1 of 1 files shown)


davidism:
Sounds like this should be fixed in Werkzeug. Also, avoid using the term "Flask development server", as Flask is not the thing providing the server.

greyli:
It would be great if it can be fixed. It seems this issue was quite old, in [this snippet](http://flask.pocoo.org/snippets/47/) that posted by Armin Ronacher in 2011, he said:
> Some WSGI servers solve that problem for you, others do not. For instance the builtin Flask webserver is pretty dumb and will not attempt to fix this problem.

davidism:
See #2188, which needs to be reported against Werkzeug.

davidism:
Related to https://github.com/pallets/werkzeug/pull/2620, which changes the development server so the 413 error will come through in most cases. In general though it's up to the client how they handle their input stream being closed before all input is sent.

PR #2663, opened 2018-03-18
Set separate Session Interface for blueprint

Tishka17:
**Idea:** Have separate session interface for each blueprint and app. 
It can be usefull when having one blueprint for site api and other one for pages. 

Added: 
* `session_interface` attribute to Blueprint object
* checking blueprint while creating/saving/etc. session in app
* one test to check this behavior

Usage example:
Assign SessionInterface object to Blueprint just the save way as with Flask application:
```python
blueprint = flask.Blueprint("bp", __name__)
blueprint.session_interface = MySessionInterface()
```

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -901,6 +901,12 @@ class Flask(_PackageBoundObject):
             from flask.testing import FlaskClient as cls
         return cls(self, self.response_class, use_cookies=use_cookies, **kwargs)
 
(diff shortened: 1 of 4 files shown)


davidism:
I don't understand the use case here, but if it really is necessary it can be done by writing a custom `SessionInterface` that does the dispatch, rather than needing to change anything in the Flask API.

Issue #2651, opened 2018-03-05
get_json's silent option cache None and causes weird side-effect


RobertDeRose:
#### Sample Code

```python
app = Flask(__name__)

@app.route("/", methods=["POST"])
def broke_behvior():
    print(request.is_json)
    rv = request.get_json(silent=True)
    print(rv)
    rv = request.get_json()
    print(rv)
    return "", 200

client = app.test_client()
client.post("/", data="'invalid json'")
```

#### Output

```bash
False
None
<Response streamed [400 BAD REQUEST]>
```

### Actual Behavior

```bash
False
None
None
<Response streamed [200 OK]>
```

The fact that calling `get_json` with `silent` and then calling it without `silent` seems to break expected behavior. When I don't call `silent` I am depending on the exception being thrown, however, because the code `caches` even on error, you get this weird behavior.

While this might be expected behavior by the authors, the docs don't make this clear.

I would think, invalid json should not be cached, yet I can see the value of it, however, it adds an layer of uncertainty and therefore, I my opinion, is a bug.

### Environment

* Python version: 3.4
* Flask version: 0.12.2
* Werkzeug version: 0.14.1

davidism:
See also #2087 and #2181.

Why are you using silent then not silent in the same request? I don't think this behavior is inconsistent.

RobertDeRose:
@davidism I'm not actually, it's a contrived example to show the behavior.  We have some before and after hooks that call `get_json(slient)` because we didn't want to throw if there was an error, but there was no way to know this would cause invalid json to be cached as None, that's not very predictable behavior and it's not documented.

davidism:
Then stop using `silent` if it's not appropriate for your use case.

davidism:
I'm fine with a PR clarifying the docs, but I don't think the behavior should change.

RobertDeRose:
I'll make a PR, but I disagree that the behavior isn't odd to the point of being a bug.  Using `silent` anywhere else in the Request's life time changes the behavior of calling the function later in the Request's life.  That's not a good thing.  Weird side-effects like this introduce hard to find bugs, like this did for our usage of Flask.

davidism:
Thought about this more. I can see the issue where an extension or `before_request` might call `get_json(slient=True)`, thus hiding the error from a user-written view when `get_json` is called.

davidism:
I'm marking this as 1.1, but if some can submit a PR in the next day or so I'll put it in 1.0.

PR #2665, opened 2018-03-19
Fix the bug of cache function in `get_json` method.

fprocess:
The BUG is:
    The cache object cache the **value** for the first call. Then the
    succedent call get the **same value** in the first time, no matter
    what are the arguments.

This PR fixed this BUG

Link to https://github.com/pallets/flask/issues/2651

BTW: This is my first PR on github. If have any problem in this PR, please tell me in detail. Thx  :-）

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -145,6 +145,8 @@ unreleased
   (`#2635`_)
 - A single trailing slash is stripped from the blueprint ``url_prefix``
   when it is registered with the app. (`#2629`_)
+- :meth:`Request.get_json() <flask.Request.get_json>` doesn't cache the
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for contributing!

This seems suboptimal in the case where it's still valid JSON even though `silent` was passed, which is hopefully the most common case. Either parsing will succeed and both cached values will be the same, or parsing will fail and the cached value for `silent` will always be `None`.

fphonor:
@davidism  Thx for your advice, which makes my thought more clear.
I refined the code now. :-)

fphonor:
Who can help me to review the code. 
I am waiting for a response.
@davidism

PR #2668, opened 2018-03-22
Add tip for dotenv

lepture:
Add a tip for installing python-dotenv when there are env files (.env, .flaskenv)

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -585,6 +585,11 @@ def load_dotenv(path=None):
     """
 
     if dotenv is None:
+        if path or os.path.exists('.env') or os.path.exists('.flaskenv'):
+            click.secho(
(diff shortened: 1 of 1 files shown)


flask/cli.py
```diff
@@ -585,6 +585,8 @@ def load_dotenv(path=None):
     """
 
     if dotenv is None:
+        if path or os.path.exists('.env') or os.path.exists('.flaskenv'):
+            click.secho(' * TIP: $ pip install python-dotenv', fg='yellow')
```
davidism: I've been trying to write more verbose messages, something like: `Tip: There are .env files present. Use "pip install python-dotenv to use them.`

lepture:
@davidism message updated.

PR #2672, opened 2018-03-26
Add nested blueprints

jdemilledt:
Rough implementation of nested blueprints. Open to suggestions.
Fixes #593

### DIFF
--- a/flask/blueprints.py
+++ b/flask/blueprints.py
@@ -114,6 +114,10 @@ class Blueprint(_PackageBoundObject):
     #: resources contained in the package.
     root_path = None
 
+    #: List of subblueprints, and order
+    blueprints = {}
+    _blueprint_order = []
+
     def __init__(self, name, import_name, static_folder=None,
                  static_url_path=None, template_folder=None,
                  url_prefix=None, subdomain=None, url_defaults=None,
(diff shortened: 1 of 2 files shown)


davidism:
I'd need to see the entire test suite for blueprints passing on nested blueprints (and their parent's behavior) before merging something like this.

I'd really like to address #1091 first, which is only going to get more confusing with this. It's not clear how relative `url_for` behaves in the nested blueprint case. This doesn't appear to track the active blueprint for children, so `request.blueprint` will still only point one level deep.

davidism:
How does this interact with `subdomain` and `url_defaults`?

This doesn't appear to register child handlers such as `before_request` with the app. I haven't run it, but it looks like it would just fail if a child had a `before_request` handler because `Blueprint` doesn't have the `before_request_funcs` attr. Same with other handlers.

davidism:
I occasionally think about this in regards to #1548. From that perspective I think it makes more sense to extract all registration functions and attrs into a common class that both `Flask` and `Blueprint` inherit.

jdemilledt:
Those are some very good points I did not think to test.

davidism:
Thanks for looking into this! I'm going to close it for now because I think we need to find a more general solution. I definitely think Flask should have this eventually, so if you want to keep developing and contributing that would definitely be appreciated.

jdemilledt:
If you ever want assistance, feel free to email me.

harveyslash:
How does this handle `blueprint.before_request` ? 
Is that function called only for the parent blueprint or for all the blueprints that call `register_blueprint` with the parent blueprint ?

therealadityashankar:
@davidism you might be able to modify an entire app into a blueprint (like the flask-docs or some open source flask based site) and use it as a test

PR #2697, opened 2018-04-11
use builtin total_seconds of timedelta object

0416354917:
It is more efficient to use the builtin `timedelta.total_seconds()`, and cleaner.

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -996,7 +996,7 @@ def total_seconds(td):
     :returns: number of seconds
     :rtype: int
     """
-    return td.days * 60 * 60 * 24 + td.seconds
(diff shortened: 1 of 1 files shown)


davidism:
This is unlikely to be faster. Internally, `timedelta.total_seconds` does the same sort of math, except it does extra, including float division, to handle microseconds. Doing this extra work just to throw it away by converting to int doesn't seem useful. In terms of "clean", `total_seconds(td)` is actually one character less than `td.total_seconds()`, not that it matters.

davidism:
Thanks for looking into this though. There may be other places where Flask implemented things in order to support Python 2.6 which are no longer needed. If you find anything like that, let us know.

PR #2699, opened 2018-04-12
add javascript ajax example

davidism:
I removed the jQuery example when rewriting the tutorial. This adds a more general JavaScript example back in, demonstrating the same simple `add` endpoint but with `XMLHttpRequest`, `fetch`, and `jQuery`.

### DIFF
--- a/docs/patterns/jquery.rst
+++ b/docs/patterns/jquery.rst
@@ -162,5 +162,6 @@ explanation of the little bit of code above:
    argument.  Note that we can use the `$SCRIPT_ROOT` variable here that
    we set earlier.
 
-If you don't get the whole picture, download the :gh:`sourcecode
-for this example <examples/jqueryexample>`.
+Check out the :gh:`example source <examples/javascript>` for a full
+application demonstrating the jQuery on this page, as well as the same
+thing using ``XMLHttpRequest`` and ``fetch``.
--- a/examples/javascript/LICENSE
+++ b/examples/javascript/LICENSE
@@ -1,0 +1,31 @@
+Copyright © 2010 by the Pallets team.
+
+Some rights reserved.
+
+Redistribution and use in source and binary forms of the software as
+well as documentation, with or without modification, are permitted
+provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice,
+  this list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright
+  notice, this list of conditions and the following disclaimer in the
+  documentation and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE AND DOCUMENTATION IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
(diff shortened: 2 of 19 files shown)


examples/javascript/js_example/templates/plain.html
```diff
@@ -0,0 +1,27 @@
...
+  by all browsers.
+{% endblock %}
+
+{% block script %}
+  <script>
+    function addSubmit(ev) {
+      ev.preventDefault();
+      var request = new XMLHttpRequest();
+      request.addEventListener('load', addShow);
+      request.open('POST', '{{ url_for('add') }}');
```
ThiefMaster: I'd use `{{ url_for('add')|tojson }}` instead of a quoted string. While not really relevant here it gives people a better example on how to properly pass data from Python to JS.

examples/tutorial/flaskr/auth.py
```diff
@@ -56,7 +56,7 @@ def register():
         elif db.execute(
             'SELECT id FROM user WHERE username = ?', (username,)
         ).fetchone() is not None:
-            error = 'User {} is already registered.'.format(username)
+            error = 'User {0} is already registered.'.format(username)
```
ThiefMaster: since when do we care about python 2.6 again? ;)

docs/patterns/jquery.rst
```diff
@@ -162,5 +162,6 @@ explanation of the little bit of code above:
    argument.  Note that we can use the `$SCRIPT_ROOT` variable here that
    we set earlier.
 
-If you don't get the whole picture, download the :gh:`sourcecode
-for this example <examples/jqueryexample>`.
+Check out the :gh:`example source <examples/javascript>` for a full
+application demonstrating the jQuery on this page, as well as the same
```
ThiefMaster: "the jQuery" reminds me of "hello sir pls help me with my jquery it shows syntax error" SO questions from people who don't know that jQuery is just JS in the end.

What about making the `fetch` one the default example? All modern browsers (IE11 isn't really modern) support it and there are good polyfills/ponyfills available for those who do need to support IE.

And you even use a polyfill below (`whatwg-fetch`) so it's going to work everywhere.

examples/javascript/LICENSE
```diff
@@ -0,0 +1,31 @@
+Copyright © 2010 by the Pallets team.
```
ThiefMaster: 2010?

examples/javascript/js_example/templates/fetch.html
```diff
@@ -0,0 +1,35 @@
...
+    function parseJSON(response) {
+      return response.json();
+    }
+
+    function addShow(data) {
+      var span = document.getElementById('result');
+      span.innerText = data.result;
+    }
+
+    document.forms[0].addEventListener('submit', addSubmit);
```
ThiefMaster: I'd really avoid using `document.forms[..]` - it's a huge mess to rely on the global index of a field... and this is code newbies look at!

docs/patterns/jquery.rst
davidism: I do want to go through the docs and rewrite them at some point, but I just didn't have time now. Running the app shows the jQuery demo last of the three, for what it's worth.

examples/tutorial/flaskr/auth.py
davidism: For at least a few more hours.

davidism:
Addressed the comments in 1e84c67beb3825564a581d76de0fcc69d57570cf.

Issue #2706, opened 2018-04-18
Different app discovery behavior on commands run and shell 


rochacbruno:
The commands `run` and `shell` are behaving differently to locate the `app`, I can send a PR but first I want to confirm if this is a problem or expected behavior.

### Expected Behavior

Assuming the tree

```bash
/topdir
|____/appdir
|_________/app.py 
```
then

```bash
export FLASK_APP=app.py
cd topdir
```

If you run

```bash
flask run
```

The command can find the `app` in the subfolder and run successfully

Expected the same to happen for `shell` 

```bash
cd topdir
flask shell
>>> 
```

### Actual Behavior

```bash
cd topdir
flask shell

# in 0.12.2
Error: The file/path provided (app.py) does not appear to exist.  Please verify the path is correct.  If app is not on PYTHONPATH, ensure the extension is .py

# n 0.13.dev
Error: Could not import "app"."

```

Looks like the `run` command is traversing the subfolders to locate an app while the `shell` command is not doing the same.

and the same applies for the new `routes` on master

```bash
$ cd topdir
$ flask routes                                                          
Usage: flask routes [OPTIONS]

Error: Could not import "app"."
```

### Environment

* Python version: 3.6.0
* Flask version: 0.12.2 & 0.13.dev (master branch)
* Werkzeug version: 0.14.1

rochacbruno:
I just figured out that the behavior of app loading is the same for `run` just the exception behavior is not the same.

The `run` command seems to "run" an `app.py` but fails later on the first request it throws:

```
flask.cli.NoAppException
flask.cli.NoAppException: Could not import "app"."
```

So the expected behavior is to do the same exception catch in `run` command if `app` cannot be located and fail early in the command and not after first request.

rochacbruno:
```
flask.cli.NoAppException
flask.cli.NoAppException: Could not import "app"."

Traceback (most recent call last)
File "/home/videos/flask_drops_repo/venv/lib/python3.6/site-packages/flask/cli.py", line 325, in __call__
self._flush_bg_loading_exception()
File "/home/videos/flask_drops_repo/venv/lib/python3.6/site-packages/flask/cli.py", line 313, in _flush_bg_loading_exception
reraise(*exc_info)
File "/home/videos/flask_drops_repo/venv/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
raise value
[...]

PR #2707, opened 2018-04-18
Add a message `Serving Flask app (module "app.py" will be lazily loaded)` to `flask run` related to #2706

rochacbruno:
When running `flask run` having `FLASK_DEBUG=1` the `lazy_loader` options defaults to True and it causes the `app` to be loaded only when first request comes causing errors to be raised only to werkzeug debugger (on browser)

# problem

When starting the command `flask run` you see `Serving Flask app..` message even if the `app` module cannot be located because the error will only raise later. It can cause confusion and the false impression that the `app` module was successfully found.

To be noticed and have it explicitly in logs I added:

` * Serving Flask app "app.py" (lazy loading)` to the messages

```bash
$ flask run                                                                    
 * Serving Flask app "app.py" (lazy loading)
```
Relates to #2706

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -933,7 +933,7 @@ class Flask(_PackageBoundObject):
         options.setdefault('use_debugger', self.debug)
         options.setdefault('threaded', True)
 
-        cli.show_server_banner(self.env, self.debug, self.name)
+        cli.show_server_banner(self.env, self.debug, self.name, False)
 
         from werkzeug.serving import run_simple
 
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -614,7 +614,7 @@ def load_dotenv(path=None):
     return new_dir is not None  # at least one file was located and loaded
 
 
-def show_server_banner(env, debug, app_import_path, eager_loading=True):
+def show_server_banner(env, debug, app_import_path, eager_loading):
     """Show extra startup messages the first time the server is run,
     ignoring the reloader.
     """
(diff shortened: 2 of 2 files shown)


rochacbruno:
@davidism

davidism:
I really don't want to add more messages to the dev server. The standard way for users to see errors when running the app is to see the error in the debugger. Thanks for looking into this, but I don't think this should change.

rochacbruno:
@davidism ok, we can avoid adding extra messages, but what do you think of changing the current one to `* Serving Flask app (The module "app.py" will be lazily loaded)` as it is explicit.

I know the usual way is seeing errors in the debugger, but this is not an application error this is a load error, the same way `shell` and `routes` command shows explicit error in the command line and also errors are shown in command line when `lazy loader` is False.

I can imagine a scenario where developer is counting on the stdout messages to assert some functional test.

davidism:
That's better, but the message is really out of proportion compared to the rest of the output (and for how important the message will usually end up being).

` * Serving Flask app "app.py" (lazy loading)` seems better.

davidism:
I don't really predict this preventing any questions one way or the other though. Whether the messages are long or short, most users who are confused by not seeing the traceback immediately will probably not know what "lazy loading" is. I still think it's fine letting users see the error in the debugger only.

rochacbruno:
@davidism updated to `* Serving Flask app "app.py" (lazy loading)`

rochacbruno:
@davidism I also could not find any entry to that `lazy loader` behavior in http://flask.pocoo.org/docs/0.12/cli/#cli

If you think it is a good idea to have it I can add it to the docs later.

davidism:
Docs describing when the app is loaded, where to see errors, and why lazy loading is used with the reloader would be good. They should be in `server.rst`, not `cli.rst`.

PR #2711, opened 2018-04-19
Minor session serializer fixes.

pope1ni:
A couple of minor fixes for the extensible session serializer added by #2352.

There were two issues observed:

1. The example usage of ``register()`` passes and ``index`` to the ``force`` argument.
2. It is not possible to insert a tag as the penultimate item in the order list.

Regarding the latter, `a.append(x)` is not the same as `a.insert(-1, x)` - it is `a.insert(len(a), x)`.

It would be good to get these out before the release of 1.0.0 so that the index default is correct.

- [x] add tests that fail without the patch
- [x] ensure all tests pass with ``pytest``
- [x] add documentation to the relevant docstrings or pages
- [ ] ~~add ``versionadded`` or ``versionchanged`` directives to relevant docstrings~~
- [ ] ~~add a changelog entry if this patch changes code~~ *Not required as the feature is unreleased?*

### DIFF
--- a/flask/json/tag.py
+++ b/flask/json/tag.py
@@ -243,7 +243,7 @@ class TaggedJSONSerializer(object):
         for cls in self.default_tags:
             self.register(cls)
 
-    def register(self, tag_class, force=False, index=-1):
+    def register(self, tag_class, force=False, index=None):
         """Register a new tag with this serializer.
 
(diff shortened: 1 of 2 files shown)


davidism:
Nice catch, thanks!

PR #2717, opened 2018-04-22
Add LICENSE in README

filipefilardi:
**Describe what this patch does to fix the issue.**

This patch doesn't fix any issue.

I added the license in readme to make easier to find it and make clear which license flask uses. 

**Link to any relevant issues or pull requests.**
none

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,6 +48,7 @@ Links
 
 * Website: https://www.palletsprojects.com/p/flask/
 * Documentation: http://flask.pocoo.org/docs/
+* License: https://github.com/pallets/flask/blob/master/LICENSE
 * Releases: https://pypi.org/project/Flask/
 * Code: https://github.com/pallets/flask
 * Issue tracker: https://github.com/pallets/flask/issues


README.rst
```diff
@@ -62,3 +62,9 @@ Links
 .. _Werkzeug: https://www.palletsprojects.com/p/werkzeug/
 .. _Jinja: https://www.palletsprojects.com/p/jinja/
 .. _pip: https://pip.pypa.io/en/stable/quickstart/
+.. _LICENSE: https://github.com/pallets/flask/blob/master/LICENSE
+
+License
```
ThiefMaster: I think this is too "noisy". Adding a link is fine (even though most people probably don't care if it's there or not), but having the section seems to be overkill.

filipefilardi: Thank you for your reply. I've removed the section. I understand your point

davidism:
I don't understand this PR. If you're looking at the README, whether on GitHub, or locally, or in a package, the license is right next to it, called LICENSE, which is a fairly standard name. The only place you don't see it is PyPI, where you only see BSD.

filipefilardi:
@davidism thank you for your reply about my PR.

Yes, the name is intuitive. I understand the desire to have a minimalist readme, I just suggested a different approach.

I have the impression that people do not give much attention to LICENSE unless it is highlighted alongside the contributors or if is somewhere in readme. Neither is the case of this repo.

I'm not saying that this is a problem or necessarily bad. I believe it is important to read the license before using the software or modifying it.

That's why i suggested the section like in my first commit.

PR #2720, opened 2018-04-24
Add project_urls so that PyPI will show GitHub stats.

lepture:
Current Flask in PyPI: https://pypi.org/project/Flask/

There is a "GitHub statistics" section, that is because the old homepage in setup.py is set to GitHub repo. But now, homepage url is replaced by pallets project page. When releasing a new version to PyPI, that section will go.

Warehouse can get the package's GitHub info from homepage url and project_urls. Let's add it back.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,8 @@
 # -*- coding: utf-8 -*-
 import io
 import re
+from collections import OrderedDict
+
 from setuptools import setup
 
 with io.open('README.rst', 'rt', encoding='utf8') as f:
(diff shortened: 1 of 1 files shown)


davidism:
Interesting, haven't seen that arg before. I see it was added to the setuptools docs, but there's no description of what keys are recognized. Is that documented somewhere?

lepture:
No, I don’t think so. I found that my Authlib doesn’t have that GitHub section, and I wonder why. So I searched the code in warehouse, the logic is that it will detect GitHub repo by homepage and project_urls. Here is my log: https://github.com/lepture/authlib/issues/45

I haven’t tested it. Maybe we can hold on this PR until I tested it with Authlib.

davidism:
I'll give it a try on a project when I get a chance too, and merge this before 1.0 if it works.

davidism:
Looks like it works fine, although I can't find the code in Warehouse that detects what URL to use for the GitHub stats.

I changed to an `OrderedDict` just to be sure they always build in the intended order.

davidism:
And don't worry, I caught that I committed the wrong URLs.

PR #2721, opened 2018-04-26
Add `python_requires` metadata for PyPI

wimglenn:
Prevents pip from seeing the dist on unsupported Python versions.


davidism:
Need to exclude 3.3 as well.

Issue #2722, opened 2018-04-26
Add a parameter to disable load dotenv using Flask CLI


fabiocerqueira:
Running Flask CLI with pipenv using the new version(1.0) 

```
pipenv run flask run
```

I always get the tip message about the load_dotenv feature. Since pipenv [already load](https://docs.pipenv.org/advanced/#automatic-loading-of-env) the `.env` file, It might be interesting have some parameter to disable Flask to load `.env` and does not show the tip message.

### Expected Behavior

Disable load dotenv using the Flask CLI. Example:

```
pipenv run flask run --ignore-dotenv
```

Or just remove the tip message and let it only in the documentation.

### Actual Behavior
Showing the dotenv tip message even if the developer doesn't want to use it
`* Tip: There are .env files present. Do "pip install python-dotenv" to use them`

### Environment

* Python version: 3.6.4
* Flask version: 1.0
* Werkzeug version: 0.14.1

lepture:
The tip is harmless for you since you know that you have loaded `.env` with pipenv. It is useful to other people, they have a `.env` file, but `.env` seems not work.

ThiefMaster:
I see his point though, this kind of "tip" gets very annoying quickly when you see it on every startup...

lepture:
pipenv is using: `PIPENV_DONT_LOAD_ENV`

```
$ PIPENV_DONT_LOAD_ENV=1 pipenv shell
```

How about we use `FLASK_DONT_LOAD_ENV`.

fabiocerqueira:
@lepture I like the idea to use the same approach as pipenv with enviroment variable instead to add a new parameter

davidism:
Thoughts on `FLASK_LOAD_ENV=0` instead of `FLASK_DONT_LOAD_ENV=1`?

lepture:
@davidism that seems weird to me. If a env var is not set, I think its value is None/False.

davidism:
I'm going with `FLASK_SKIP_DOTENV`, as suggested in the PR. I think it more clearly describes what's happening.

PR #2724, opened 2018-04-26
Use env var to disable loading dotenv

lepture:
Fix issue: https://github.com/pallets/flask/issues/2722

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -544,7 +544,10 @@ class FlaskGroup(AppGroup):
         # script that is loaded here also attempts to start a server.
         os.environ['FLASK_RUN_FROM_CLI'] = 'true'
 
-        if self.load_dotenv:
+        val = os.environ.get('FLASK_DONT_LOAD_ENV')
+        load_dotenv =  not val or val in ('0', 'false', 'no')
+
+        if self.load_dotenv and load_dotenv:
             load_dotenv()
 
         obj = kwargs.get('obj')
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -28,7 +28,7 @@ from werkzeug.utils import import_string
 from . import __version__
 from ._compat import getargspec, iteritems, reraise, text_type
 from .globals import current_app
-from .helpers import get_debug_flag, get_env
+from .helpers import get_debug_flag, get_env, get_load_dotenv
 
 try:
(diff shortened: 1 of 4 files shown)


lepture:
I need to add a test case.

flask/cli.py
```diff
@@ -544,7 +544,10 @@ def main(self, *args, **kwargs):
         # script that is loaded here also attempts to start a server.
         os.environ['FLASK_RUN_FROM_CLI'] = 'true'
 
-        if self.load_dotenv:
+        val = os.environ.get('FLASK_DONT_LOAD_ENV')
+        load_dotenv =  not val or val in ('0', 'false', 'no')
```
ThiefMaster: too much space. also, is this the same logic as pipenv uses?

lepture: @ThiefMaster No, this is the same logic `FLASK_DEBUG` is using.

ThiefMaster: ok fine for me as well

lepture:
Oops. Tests should not fail since I didn't add any changes to break the test. Weird.

lepture:
@davidism any idea why tests failed? I've tested master branch on my local machine, tests failed too.

davidism:
I can take a look tomorrow. The most common issue I had with CLI tests was that a test would forget to reset the env vars. Not sure if that's related here.

lepture:
The weird thing is master branch's tests fail on my local machine.

flask/cli.py
davidism: `load_dotenv` is the name of the function that will be called later, it's been overwritten as a boolean.

davidism:
Needs a test.

rhymes:
Shouldn't it be called "FLASK_IGNORE_DOTENV" or "FLASK_SKIP_DOTENV" ?

"Flask don't load env" it's not exactly accurate because the flag purpose is to skip reading variables from .env, not to skip the shell environment.

davidism:
Discussion at #2722, it was chosen to match pipenv's `PIPENV_DONT_LOAD_ENV`. I agree that either of those sound better though; I like `SKIP`.

davidism:
Added the check to `app.run` as well in 6e1e3e03ca188a4d9b3aef493e572157c8ae2313.

Issue #2601, opened 2018-01-18
Internal error when requesting route with url-part == defaults


d3QUone:
http://flask.pocoo.org/docs/0.12/api/#url-route-registrations

Getting a 5xx-error when requesting a route with defaults parameter 

### Code Example

```python
view = Blueprint("a", __name__, url_prefix="/a")

@view.route("/b", defaults={"consumer_id": "test"}, methods=["POST"])
@view.route("/b/<consumer_id>", methods=["POST"])
def check_consumers(consumer_id):
    return "consumer_id: {}".format(consumer_id)
```

---

### Expected Behavior

Send any other parameter - OK
```
curl -X POST '0.0.0.0:5500/a/b/anything'
```
```
consumer_id: anything
```
Send parameter == default one
```
curl -X POST '0.0.0.0:5500/a/b/test'
```
```
consumer_id: test
```

---

### Actual Behavior

Send parameter == `defaults` and get an internal error
```
curl -X POST '0.0.0.0:5500/a/b/test' -vvv
```
```
* Hostname was NOT found in DNS cache
*   Trying 0.0.0.0...
* Connected to 0.0.0.0 (127.0.0.1) port 5500 (#0)
> POST /a/b/test HTTP/1.1
> User-Agent: curl/7.35.0
> Host: 0.0.0.0:5500
> Accept: */*
> 
* Empty reply from server
* Connection #0 to host 0.0.0.0 left intact
curl: (52) Empty reply from server
```

Traceback:
```pytb
Note: this exception is only raised in debug mode
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1590, in dispatch_request
    self.raise_routing_exception(req)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1576, in raise_routing_exception
    raise FormDataRoutingRedirect(request)
FormDataRoutingRedirect: A request was sent to this URL (http://0.0.0.0:5500/a/b/test) but a redirect was issued automatically by the routing system to "http://0.0.0.0:5500/a/b".  Make sure to directly send your POST-request to this URL since we can't make browsers or HTTP clients redirect with form data reliably or without user interaction.
```

---

### Environment

* Python version: 2.7.6
* Flask version: 0.12.2
* Werkzeug version: 0.14.1

[...]

PR #2729, opened 2018-04-27
Added short note about Werkzeug redirect behaviour

Dunstrom:
Addresses issue [#2601](https://github.com/pallets/flask/issues/2601) with a short note about Werkzeug redirect behaviour in the documentation.


lepture:
It seems not clear enough with a note like this.

PR #2733, opened 2018-04-27
Fixed Debug default value in the docs

kivS:
Fix Debug Default: By default Debug is only True when ENV is 'development' instead of 'production'


lepture:
Thanks.

PR #2737, opened 2018-04-27
fix #2736 - remove explicit check for list as a return value in make_response

### LINKED ISSUE #2736: make_response changes breaks custom response class
### Expected Behavior

I should be able to return a single value from my view(that is, not a tuple) and have my value passed to my custom response class. In this case I have a response class that detects whether JSON or pickle data was requested by the client and dumps the return value appropriately. This allows me to simply return a list or dict from my view without having to detect and dump the appropriate type.

```python
@mod.route('/')
def return_json_or_pickle():
    return ['foo', 'bar', 'bat']
```

As of Flask 0.12.2 this worked - make_response would pass the list to my custom response class and my client would get what it wanted.

### Actual Behavior

As of 1.0, commit #2256 was merged which added an explicit check to see if the return value was a list; treating a list as a tuple return in that case. I now get a typeerror:

```pytb
TypeError: The view function did not return a valid response tuple. The tuple must have the form (body, status, headers), (body, status), or (body, headers).
```

If I change my view function to return a tuple:

```python
@mod.route('/')
def return_json_or_pickle():
    return ['foo', 'bar', 'bat'], 200
```
the previous functionality is restored! This confirms that tuple handling is of course perfectly fine. I have a [pull request](https://github.com/pallets/flask/pull/2737) prepared which should do a very nice job of resolving the issue - it simply omits `list` from the type check in make_response. Since a list is not a valid response anyway, there's no reason I can see for it to be there.

### Environment

* Python version: 3.6
* Flask version: 1.0
* Werkzeug version: n/a

jesseops:
This PR addresses an issue introduced in #2256 which breaks a custom response class that supports returning lists from view functions.

A list is not an expected response value and is not intended to be handled by `make_response`. I'm unsure why it made it's way into this instance check, but there are no consequences to removing it, based on what I can see in the tests & related code.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1923,7 +1923,7 @@ class Flask(_PackageBoundObject):
         status = headers = None
 
         # unpack tuple returns
-        if isinstance(rv, (tuple, list)):
+        if isinstance(rv, tuple):
             len_rv = len(rv)
 
             # a 3-tuple is unpacked directly


davidism:
OK, your fork is probably going to be messed up after this, since you made the PR from your master and I just force pushed it to look like 1.0-maintenance. You can probably fix it by force pushing origin/master and origin/1.0-maintenance to your fork once this is merged, or just delete and fork again.

PR #2740, opened 2018-04-28
Adding Windows PowerShell specfic docs

powellnathanj:
Adding Windows PowerShell specfic docs

### DIFF
--- a/examples/tutorial/README.rst
+++ b/examples/tutorial/README.rst
@@ -31,6 +31,11 @@ Or on Windows cmd::
     py -3 -m venv venv
     venv\Scripts\activate.bat
 
+Or on Windows PowerShell::
+
+    python -m venv venv
+    venv\Scripts\activate
+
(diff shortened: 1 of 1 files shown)


examples/tutorial/README.rst
```diff
@@ -31,6 +31,11 @@ Or on Windows cmd::
     py -3 -m venv venv
     venv\Scripts\activate.bat
 
+Or on Windows PowerShell::
+
+    python -m venv venv
```
davidism: Still need to use `py -3` instead of `python` here. Does `venv\Scripts\activate` work in CMD as well? If so we can consolidate the two for this step.

davidism:
Can you resubmit this patch as a branch off 1.0-maintenance, rather than from your master branch. If I rebase this to 1.0-maintenance for you, it will mess up your fork since you're not in a branch.

PR #2745, opened 2018-04-29
Update quickstart.rst

dmarasquin:
- Fixes the syntax error on the quickstart doc by escaping the single quote.
- Adds request import and removes duplicated word

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -293,7 +293,7 @@ Python shell. See :ref:`context-locals`. ::
 
     @app.route('/user/<username>')
     def profile(username):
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -315,6 +315,8 @@ a route only answers to ``GET`` requests. You can use the ``methods`` argument
 of the :meth:`~flask.Flask.route` decorator to handle different HTTP methods.
 ::
 
+    from flask import Flask, request
```
davidism: The `Flask` class isn't referenced in this code block, don't need to import it.

PR #2748, opened 2018-05-01
Fix blueprint route for ""

lepture:
1. url_prefix = '/posts'
2. rule = ''

Expected result `/posts`, instead got `/posts/`.

### DIFF
--- a/flask/blueprints.py
+++ b/flask/blueprints.py
@@ -10,6 +10,7 @@
     :license: BSD, see LICENSE for more details.
 """
 from functools import update_wrapper
+from werkzeug.urls import url_join
 
 from .helpers import _PackageBoundObject, _endpoint_from_view_func
 
@@ -49,8 +50,6 @@ class BlueprintSetupState(object):
         url_prefix = self.options.get('url_prefix')
         if url_prefix is None:
             url_prefix = self.blueprint.url_prefix
-        if url_prefix:
-            url_prefix = url_prefix.rstrip('/')
         #: The prefix that should be used for all URLs defined on the
         #: blueprint.
         self.url_prefix = url_prefix
@@ -67,7 +66,8 @@ class BlueprintSetupState(object):
         """
         if self.url_prefix is not None:
             if rule:
-                rule = '/'.join((self.url_prefix, rule.lstrip('/')))
+                rule = '/'.join((
+                    self.url_prefix.rstrip('/'), rule.lstrip('/')))
             else:
                 rule = self.url_prefix
         options.setdefault('subdomain', self.subdomain)
--- a/flask/blueprints.py
+++ b/flask/blueprints.py
@@ -66,7 +66,10 @@ class BlueprintSetupState(object):
         blueprint's name.
         """
         if self.url_prefix is not None:
-            rule = '/'.join((self.url_prefix, rule.lstrip('/')))
+            if rule:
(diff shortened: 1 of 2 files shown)


davidism:
I think the new behavior is more technically correct. Just like the root of the application is `'/'`, not `''`, the root of a blueprint is the prefix plus a slash. However, we can go back to the previous behavior. Maybe become more strict in 1.1.

flask/blueprints.py
```diff
@@ -66,7 +66,10 @@ def add_url_rule(self, rule, endpoint=None, view_func=None, **options):
         blueprint's name.
         """
         if self.url_prefix is not None:
-            rule = '/'.join((self.url_prefix, rule.lstrip('/')))
+            if rule:
+                rule = '/'.join((self.url_prefix, rule.lstrip('/')))
```
davidism: I think this makes `@bp.route('/')` with prefix `/test` become `/test` instead of `/test/` now, which is also incorrect.

lepture: see my test.

```
'/'.join(['/test', '']) => /tests/
```

lepture:
@davidism I think `1.0.1` should be a fix version, but it actually break the routing behavior. Flask 1.0 won't force a redirect of `/test` to `/test/`, but Flask 1.0.1 will.

flask/blueprints.py
davidism: `prefix='/foo/', rule=''` results in `/foo` instead of what people would probably expect, `/foo/`. I think we need to take a different approach to merging, we're trying to cover too many corner cases.

davidism:
I think I might just revert the whole thing. I've been messing with it and can't figure out a clean way to differentiate between intended and unintended slashes. At some point we just have to hope that users understand that `prefix + rule` can result in double slashes if they aren't consistent.

davidism:
I've adjusted it a little bit to allow more cases.

The following are still not allowed: `('', '')` and `('', 'bar')` because they result in "rule must start with slash". This means the dev needs to be consistent with whether the bp's rules start with slashes.

```python
prefix = ''
@bp.route('')  # ValueError
@bp.route('/')  # '/'
@bp.route('bar')  # ValueError
@bp.route('/bar')  # '/bar'

prefix = '/'
@bp.route('')  # '/'
@bp.route('/')  # '/'
@bp.route('bar')  # '/bar'
@bp.route('/bar')  # '/bar'

prefix = '/foo'
@bp.route('')  # '/foo'
@bp.route('/')  # '/foo/'
@bp.route('bar')  # '/foo/bar'
@bp.route('/bar')  # '/foo/bar'

prefix = '/foo/'
@bp.route('')  # '/foo/'
@bp.route('/')  # '/foo/'
@bp.route('bar')  # '/foo/bar'
@bp.route('/bar')  # '/foo/bar'
```

Given that neither of these `ValueError` cases would have been possible in 0.12 either, I'm comfortable leaving it out, I think it's more consistent with how app routes work. I can make them both work but I think the consistency it enforces is better.

lepture:
GLTM now.

nacin:
@davidism I think a6a6810 broke something else we were doing in 1.0, which was:

```
prefix = '/team'
@bp.route('/')  # '/team/'
@bp.route('.csv')  # '/team.csv', but now fails
```

davidism:
That was never intended to work.

PR #2749, opened 2018-05-02
Inline variable

Stranger6667:
Trivial change


davidism:
Thanks for the interest, but this PR doesn't change anything meaningfully. Also, this fixture doesn't do any teardown, so there's no point yielding instead of returning.

PR #2753, opened 2018-05-02
1.0 release codename

decaz:
The release is incomplete without a codename so I propose to name the 1.0 release codename "Vodka" 🥂.


davidism:
Thanks for the interest, but I decided to stop using code names unless we were actually talking about them secretly before they were released.

davidism:
Eh, sure.

PyDever:
ooh codenames sound fun

Issue #2514, opened 2017-11-09
Unable to generate the PDF version of the documentation


matrixise:
commit rev: 470112dd6e708a18eb4ec1daa10c4ea276620dbf
python: 3.6.2

in the `docs` directory, just execute `make latexpdf`, there are issues with `\maketitle` and an unicode char.
```
(/usr/share/texlive/texmf-dist/tex/latex/hyperref/nameref.sty
(/usr/share/texlive/texmf-dist/tex/generic/oberdiek/gettitlestring.sty))
! Undefined control sequence.
\maketitle ...tnoterule \relax \ifsphinxpdfoutput 
                                                  \begingroup \def \\{, } \p...
l.66 \maketitle
               
? 
```

```
! Package inputenc Error: Unicode char 🎉 (U+1F389)
(inputenc)                not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.20767 Celebrate 🎉
                      
? 
```

```
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 1
      Refer to 'Flask.log' for details
Latexmk: Use the -f option to force complete processing,
 unless error was exceeding maximum runs of latex/pdflatex.
make[1]: *** [Makefile:33: Flask.pdf] Error 12
make[1] : on quitte le répertoire « /home/stephane/src/github.com/pallets/flask/docs/_build/latex »
make: *** [Makefile:101: latexpdf] Error 2
```

davidism:
I run into these issues constantly when building new docs. We can't run the automatic doc builder because it gets stuck, so I have to manually run it on new releases and skip through all the error prompts. We are switching to Read The Docs (pallets/meta#15), but if anyone understands Latex and wants to take a look before then it would be appreciated.

lepture:
We shouldn't care about things like this.

@mitsuhiko is there a certain reason that we don't use RTD?

davidism:
There is some internal discussion going on about that, but I'd still like to fix it.

danielopitz:
This should be working now :)

matrixise:
Ok I will try ;-)

PR #2754, opened 2018-05-03
Fixed LaTeX errors on documentation PDF generation

danielopitz:
This PR fixes problems with the generation of the PDF documentation from LaTeX files. 

First issue was a problematic emoji character in the documentation, which the LaTeX distribution used in the build process didn't understand. The least invasive fix was to replace it with a space. Alternative fixes would be removal of the character causing errors or adapting the build process.

Second issue was the missing `\ifsphinxpdfoutput` macro. In this case it was only used to check if the output is PDF, which is now handled with an appropriate if statement.

Relevant issues:
https://github.com/pallets/flask/issues/2514
https://github.com/pallets/flask-docs/issues/3

### DIFF
--- a/docs/flaskstyle.sty
+++ b/docs/flaskstyle.sty
@@ -1,11 +1,17 @@
 \definecolor{TitleColor}{rgb}{0,0,0}
 \definecolor{InnerLinkColor}{rgb}{0,0,0}
 
+% Replace Unicode character 'PARTY POPPER' (U+1F389) with a non-breaking space.
+% pdfLaTeX doesn't support Unicode.
+\DeclareUnicodeCharacter{1F389}{\nobreakspace}
+
 \renewcommand{\maketitle}{%
   \begin{titlepage}%
     \let\footnotesize\small
     \let\footnoterule\relax
-    \ifsphinxpdfoutput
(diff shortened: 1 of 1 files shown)


davidism:
Is there a way to tell LaTeX about emoji, or is that too complicated?

danielopitz:
There are two ways to do it I know of:

Method A:
1. Change the font to one with emoji (DejaVu Sans has monochrome emoji)
2. Change the LaTeX processor from pdfLaTeX to XeLaTeX or LuaLaTeX, both understand Unicode in text and at least XeLaTeX allows to use any OpenType fonts.

Method B:
1. Have a LaTeX font containing emoji (!= OpenType / TrueType)
2. Have a mapping file containing all mappings (something like `\DeclareUnicodeCharacter{1F389}{\emojiPartyPopper}`). Such a mapping probably exists somewhere in the wild.

davidism:
Thanks for fixing this. I'm totally lost with LaTeX. :sweat_smile: I cherry picked it to 0.12-maintenance and 1.0-maintenance.

For the future, remember to make PRs from branches, then we can directly rebase without messing up your master branch.

Issue #2755, opened 2018-05-03
templates_auto_reload not taken into account


romuald:
### Synopsis

If the configuration templates_auto_reload is set up after has set up it's jinja environment (example: after setting up some filters), it is ignored by Flask

### Expected Behavior

Templates should be reloaded when app.config['TEMPLATES_AUTO_RELOAD'] is set to True

```python
# application.py
from flask import Flask, render_template

app = Flask(__name__)

@app.route('/')
def hello(name=None):
    return render_template('hello.html', name=name)

# Templates are reloaded if configuration is set up before the template filter
# app.config.update({'TEMPLATES_AUTO_RELOAD': True})

@app.template_filter('dummy')
def dummy(value):
    return value

# Setting this up here does not do anything
app.config.update({'TEMPLATES_AUTO_RELOAD': True})

# Neither does this:
app.templates_auto_reload = True

app.run()

#% python application.py
```

### Actual Behavior

The option is ignored because the jinja environment was already created

### Environment

* Python version: 3.6.5
* Flask version: 1.0.2
* Werkzeug version: 0.14.1
* debug is set to False in this case

ozzywalsh:
Is this really a problem?
Shouldn't config be set before any routes, template filters etc are defined?

lepture:
Close it for now. It seems not a problem, I can't figure out a use case.

romuald:
I've had this exact use case while developing my software:

- load local libraries that have template filters
- use a configure method that updates the configuration dict
- myapp.run() to start the service

lepture:
@romuald why not set `TEMPLATES_AUTO_RELOAD=true` before `.run()`?

davidism:
Configuration should happen as early as possible. Nothing is guaranteed to read configuration after it initializes.

romuald:
@ozzywalsh that's what I've done, moved the templates filter in a specific module specifically located after the configuration

PR #2756, opened 2018-05-03
templates_auto_reload workaround

romuald:
When jinja was already set up (by having some template filters for
example), modify the jinja_env to it will take the modification into
account.
Otherwise this decorator doesn't do anything in that case

Workaround for #2755, this does not fix the config dict not being taken into account

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -708,6 +708,11 @@ class Flask(_PackageBoundObject):
     def _set_templates_auto_reload(self, value):
         self.config['TEMPLATES_AUTO_RELOAD'] = value
 
+        # workaround to not create jinja_env if it has not been created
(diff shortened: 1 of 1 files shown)


romuald:
I have no idea how to fix this correctly, my best guess would be to modify `add_template_filter` to reference functions in a specific dict until `create_jinja_environment` is called

lepture:
@davidism I can't see the necessary for this use case.

davidism:
Thanks for working on this. While this fix is technically fine, I'm going to close it as I don't want to keep creating the impression that modifying config after some setup has already run is a good idea. The solution here is to adjust your code so that all configuration loads before setup.

PR #2764, opened 2018-05-08
docs: installation: Fix "now you can continue" link

mgalgs:
At the end of the installation docs we currently
point the reader back into the middle of the
installation doc itself, which is probably not very
helpful to the reader since they just finished
reading that.

Instead, point to the next section in the docs
(Quickstart).


davidism:
The ref is correct. Users get to that section by clicking a link earlier in the page, then the link at the bottom of the section takes them back to continue the instructions.

mgalgs:
> Add a link to the end of the "Install Flask" section instead.

Done

PR #2781, opened 2018-05-13
route app.run through the cli whenever possible

miguelgrinberg:
This change expands the venerable `app.run()` function with cli support. At this point this is a proof of concept to investigate if this can be done without breaking existing applications. Feedback appreciated!

Given an application `myapp.py`, which ends with:

```python
if __name__ == '__main__':
    app.run()
```

The following features are supported:

- Running `python myapp.py` with no additional arguments is redirected to `flask run` in the cli module. The value of `FLASK_APP` is determined through introspection if not provided. If a value for `FLASK_APP` cannot be determined, then the original `app.run()` function is used.
- Running `python myapp.py` with any arguments redirects to the `flask` command with the same arguments. For example: `python myapp.py --help` is the same as `flask --help`, `python myapp.py shell` starts the shell, and so on. Custom cli commands are also supported.
- If the script passes `host` or `port` arguments in `app.run()`, those are also passed on to the `flask run` command in the cli module.
- If the script calls `app.run(debug=True)` then `FLASK_ENV=development` is added to the environment (only if `FLASK_ENV` is not set).
- If the `app.run()` call specifies any custom settings besides `host`, `port`, and `debug`, then the original logic is used for now. I can go through all the available options to `run_simple()` and see if there are other ones that can be supported through the cli.

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -899,12 +899,28 @@ class Flask(_PackageBoundObject):
             The default port is now picked from the ``SERVER_NAME``
             variable.
         """
+        # determine if this method is being invoked in the standard way (from
+        # the global scope of the __main__ module) by checking the length of
+        # the call stack
+        from inspect import stack
+        called_from_main = len(stack()) == 2
+
         # Change this into a no-op if the server is invoked from the
         # command line. Have a look at cli.py for more information.
-        if os.environ.get('FLASK_RUN_FROM_CLI') == 'true':
-            from .debughelpers import explain_ignored_app_run
-            explain_ignored_app_run()
-            return
+        if called_from_main:
+            if os.environ.get('FLASK_RUN_FROM_CLI') == 'true':
+                if os.environ.get('FLASK_RUN_FROM_APP_RUN') != 'true':
+                    from .debughelpers import explain_ignored_app_run
+                    explain_ignored_app_run()
+                return
+
+            if load_dotenv and not options:
+                # Since this is a call with no custom settings besides host,
+                # port and debug, we can attempt to route it through the cli
+                # runner, and only fall back to the implementation in this
+                # function if that does not work
+                if cli.run_from_app_run(self, host, port, debug):
+                    return
 
         if get_load_dotenv(load_dotenv):
             cli.load_dotenv()
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -874,6 +874,50 @@ debug mode.
 ))
 
 
+def run_from_app_run(app, host=None, port=None, debug=None):
+    if 'FLASK_APP' not in os.environ:
+        # attempt to determine the value of FLASK_APP through introspection
+        import inspect
+
+        app_name = None
+        for name, value in inspect.getmembers(sys.modules['__main__']):
+            if value == app:
+                app_name = name
+                break
+        if app_name is None:
+            return False
+
+        os.environ['FLASK_APP'] = '{}:{}'.format(sys.argv[0], app_name)
+
+    if 'FLASK_ENV' not in os.environ:
+        os.environ['FLASK_ENV'] = 'development' if debug else 'production'
+
+    # look for the flask executable in the same location where python is
+    executable = os.path.join(os.path.dirname(sys.executable), 'flask')
+    if not os.path.exists(executable):
+        # flask was not found in the expected location, let's assume that it
+        # is in the path
+        executable = 'flask'
+
+    # generate the cli arguments
+    if len(sys.argv) == 1:
+        # if no arguments were given in the command, default to "run",
+        # and add custom host and port if needed
+        args = ['run']
+        if host:
+            args += ['-h', host]
+        if port:
+            args += ['-p', port]
+    else:
+        # pass all the original arguments directly into the cli
+        args = sys.argv[1:]
+
+    sys.argv = [executable] + args
+    os.environ['FLASK_RUN_FROM_APP_RUN'] = 'true'
+    main()
+    return True
+
+
 def main(as_module=False):
     args = sys.argv[1:]
 


flask/app.py
```diff
@@ -902,9 +902,23 @@ def run(self, host=None, port=None, debug=None,
         # Change this into a no-op if the server is invoked from the
         # command line. Have a look at cli.py for more information.
         if os.environ.get('FLASK_RUN_FROM_CLI') == 'true':
-            from .debughelpers import explain_ignored_app_run
-            explain_ignored_app_run()
-            return
+            # if the stack has two frames, it means that app.run() was called
+            # from the main script of the application, which is likely a
+            # mistake unless this is comming from an app.run() call
```
ThiefMaster: s/comming/coming

miguelgrinberg:
Guys, this change has been gathering dust since May. Anything I can do to make it more palatable and hopefully get it in?

The most important reason I want this (or something similar) in is that it beings parity between the two methods of starting a Flask app. A lot of people prefer and still use the `app.run()` method to start the app, but they are at a disadvantage because they are left out of custom CLI commands, and are also affected by the reloader bugs that originated the CLI in the first place. This change I believe addresses both concerns.

flask/cli.py
```diff
@@ -874,6 +874,50 @@ def routes_command(sort, all_methods):
...
+        if app_name is None:
+            return False
+
+        os.environ['FLASK_APP'] = '{}:{}'.format(sys.argv[0], app_name)
+
+    if 'FLASK_ENV' not in os.environ:
+        os.environ['FLASK_ENV'] = 'development' if debug else 'production'
+
+    # look for the flask executable in the same location where python is
+    executable = os.path.join(os.path.dirname(sys.executable), 'flask')
```
davidism: I think this will fail on Windows, where it's "flask.exe".

flask/app.py
```diff
@@ -899,12 +899,28 @@ def run(self, host=None, port=None, debug=None,
...
             variable.
         """
+        # determine if this method is being invoked in the standard way (from
+        # the global scope of the __main__ module) by checking the length of
+        # the call stack
+        from inspect import stack
+        called_from_main = len(stack()) == 2
+
         # Change this into a no-op if the server is invoked from the
         # command line. Have a look at cli.py for more information.
```
davidism: Need to update this comment.

flask/app.py
```diff
@@ -899,12 +899,28 @@ def run(self, host=None, port=None, debug=None,
...
-            explain_ignored_app_run()
-            return
+        if called_from_main:
+            if os.environ.get('FLASK_RUN_FROM_CLI') == 'true':
+                if os.environ.get('FLASK_RUN_FROM_APP_RUN') != 'true':
+                    from .debughelpers import explain_ignored_app_run
+                    explain_ignored_app_run()
+                return
+
+            if load_dotenv and not options:
```
davidism: I'm uncomfortable with this part long term. I think people are going to be surprised when suddenly their `python myapp.py shell` command fails because they passed another option to `app.run`.

flask/cli.py
```diff
@@ -874,6 +874,50 @@ def routes_command(sort, all_methods):
...
+        if host:
+            args += ['-h', host]
+        if port:
+            args += ['-p', port]
+    else:
+        # pass all the original arguments directly into the cli
+        args = sys.argv[1:]
+
+    sys.argv = [executable] + args
+    os.environ['FLASK_RUN_FROM_APP_RUN'] = 'true'
```
davidism: Should this env var be set by `app.run` as well since that's what its name implies?

flask/cli.py
```diff
@@ -874,6 +874,50 @@ def routes_command(sort, all_methods):
 ))
 
 
+def run_from_app_run(app, host=None, port=None, debug=None):
```
davidism: This is now a third place, in addition to `app.run` and `run_command`, where a different subset of args is supported.

davidism:
I suppose I approve this, but I don't like how circuitous and duplicated it's becoming (that's not your fault though). I'd really like to investigate how we can actually unify `Flask.run` and `run_command`. Needs docs, tests, and changelog before merge.

I'm currently focused on other things (in Pallets and in general), so I won't be able to make a release for a while longer unless someone wants to handle more of the process for me (check what's been merged since the last tag, ensure changelog is up to date and links to issues/PRs, decide to merge any other PRs for a feature release, etc.).

flask/cli.py
```diff
@@ -874,6 +874,50 @@ def routes_command(sort, all_methods):
...
+        # if no arguments were given in the command, default to "run",
+        # and add custom host and port if needed
+        args = ['run']
+        if host:
+            args += ['-h', host]
+        if port:
+            args += ['-p', port]
+    else:
+        # pass all the original arguments directly into the cli
+        args = sys.argv[1:]
```
davidism: If users were doing their own CLI parsing in `__main__` then calling `app.run`, this will fail, because `sys.argv` will still contain any arguments passed to `python myapp.py`, that probably don't match up with Flask's CLI arguments.

For example, `python myapp.py --config=dev`, with `app.run()` in a `__main__` block, `called_from_main` will still be true, but the args will no longer line up.

flask/cli.py
miguelgrinberg: I'll find a Windows machine to test this, but I don't think it will fail. I don't think you have to provide the extension, the OS tries a bunch of known executable extensions such as .exe, .com and .bat until one matches.

flask/cli.py
miguelgrinberg: Yes. It's unfortunate, but I see it as a way to converge the current two ways to start an app into a unified solution, because eventually the arguments in `app.run` could be deprecated in favor of CLI arguments.

flask/cli.py
miguelgrinberg: Maybe it would be more clear if this var was called `FLASK_RUN_CLI_FROM_APP_RUN`? Because the only time you want this to be set is when the `app.run` call is routed through the CLI.

flask/cli.py
miguelgrinberg: @davidism no problem at all running this patch on Windows.

ElPicador:
I know this PR is old, but I think it's useful. Do you plan to merge this one day?

We need a custom cli to run flask in dev. For now we do: 

```
os.environ["FLASK_RUN_FROM_CLI"] = "false"
app.run(...)
```

Is there a better way to do this?

davidism:
At this point, as outlined in my comments above, I don't think I want to add this much complexity to these methods. I'm still interested in refactoring how `flask.cli.run_command` and `flask.app.Flask.run` work so they're more in line, but I think it will require a different approach.

If you'd like to run the CLI instead of `app.run()`, import `flask.cli.main` and call that:

```python

if __name__ == "__main__":
    from flask.cli import main

    main()
```

sunrisa:
> At this point, as outlined in my comments above, I don't think I want to add this much complexity to these methods. I'm still interested in refactoring how `flask.cli.run_command` and `flask.app.Flask.run` work so they're more in line, but I think it will require a different approach.
> 
> If you'd like to run the CLI instead of `app.run()`, import `flask.cli.main` and call that:
> 
> ```python
> if __name__ == "__main__":
>     from flask.cli import main
> 
>     main()
> ```

This solution doesn't work for me if you're using `flask-socketio`. Get the message `WARNING in __init__: Flask-SocketIO is Running under Werkzeug, WebSocket is not available.`. Any suggestions? @davidism

davidism:
If you're using something that's not Flask to run your app, you need to run that command.

Issue #2778, opened 2018-05-13
Registered error handler (using register_error_handler) processes inappropriate exceptions


woxcab:
If register an error handler for exception class `A` then this error handler will be called for any exception `B` that isn't `A` or subclass of `A`.

## Expected Behavior
Given error handler must be called only for exceptions that's instance of `A` or instance of subclass of `A`.

## Actual Behavior
Error handler processes any exception.

## How to reproduce
Minimal code:

```python3
import flask
from werkzeug.exceptions import HTTPException

app = flask.Flask(__name__)
app.register_error_handler(HTTPException, lambda e: (str(e), e.code))

@app.route('/')
def index():
    raise ValueError()

app.run()
```

1. Run script
2. Go to  http://127.0.0.1:5000/ 

Server logs exception `AttributeError: 'ValueError' object has no attribute 'code'`, i.e. registered error handler is called for `ValueError` exception, but the error handler is registered only for `HTTPException` and its subclasses.

## Environment

* Python version: 3.5.2
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

sgunturu5:
Discussed with @davidism 
HTTPException Handler is the default exception handle for unhandled exceptions.  Hence it need to be generic. Planning to update the documentation.

ThiefMaster:
Is this really the behavior we want? I think an error handle for `Exception` should be catchall - but in any `HTTPException` error handler getting an exception that is not a subclass of `HTTPException` seems wrong/confusing.

davidism:
Flask wants to raise a 500 error when there is an unhandled exception. But it tries to use the `InternalServerError` (500) error handler to handle that 500 error (which falls back to the `HTTPException`). That seems correct, but means that they need to not make assumptions about the incoming error.

ThiefMaster:
What I mean is that the error handler should receive the 500 error, but with an InternalServerError exception -- not the original exception that caused it. IMO people should register an error handler for Exception, ValueError, etc. if they want these exceptions.

taion:
@ThiefMaster 

I spent a few minutes poking around at your recommended solution, because it was the one that seemed to make sense, and it's straightforward to implement.

However, the test suite gives an example for when the current behavior might be desired: https://github.com/pallets/flask/blob/7e714bd28b6e96d82b2848b48cf8ff48b517b09b/tests/test_user_error_handler.py#L29-L31

[...]

PR #2783, opened 2018-05-14
2778 processes inappropriate exceptions

tullyrankin:
This MR fixes issue with custom error handler not being called when exception is raised.

Issue #2778


davidism:
I'm not really clear what this change is doing. How does this address the linked issue?

PR #2786, opened 2018-05-16
[Docs] Make super() compatible for Python2

greyli:
Using `super()` with arguments for backward compatible.


lepture:
Since we are still supporting Python 2, I'm going to merge it.

PR #2787, opened 2018-05-16
Added parameter to click.echo call

deveshbatra:
Added sys.stdout as a second parameter to click.echo. This helps with the 'UnsupportedOperation: not writable' error in Jupyter.

I think that Jupyter hijacks and locks the STDOUT/STDERR (at least the one click is trying to use) and if you don't provide a stream to click.echo(), it will attempt writing to the STDOUT/STDERR, hence the error.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -626,7 +626,7 @@ def show_server_banner(env, debug, app_import_path, eager_loading):
         if not eager_loading:
             message += ' (lazy loading)'
 
-        click.echo(message)
+        click.echo(message, sys.stdout)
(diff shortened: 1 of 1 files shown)


davidism:
This is the first I'm hearing of any error, and there is not enough of a description about what that error is. Typically, a ticket describing the issue in detail should be created first.

However, this does not seem like an issue that should be fixed in Flask. It sounds like an issue with Jupyter or Click. This change bypasses everything Click does to ensure the output stream behaves consistently on different platforms.

PR #2789, opened 2018-05-17
AttributeError: 'module' object has no attribute 'SSLContext'

briehanlombaard:
We're still using Python 2.7.6 for some projects which causes the exception listed below when trying to execute `flask run`. The problem appears to be when the run command executes `flask.cli._validate_key` which then checks for Python version 2.7 or above before attempting to access `ssl.SSLContext`. The problem with that is that SSLContext was only added in 2.7.9 so the check needs to be more exact by checking for versions 2.7.9 and above.

Relevant tests already fail under Python 2.7.6 and have been updated so this patch doesn't require any additional tests as far as I can tell. Let me know if that is not the case.

```
  File "/usr/local/bin/flask", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/flask/cli.py", line 894, in main
  File "/usr/local/lib/python2.7/dist-packages/flask/cli.py", line 557, in main
    return super(FlaskGroup, self).main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1064, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)  
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 621, in make_context
    self.parse_args(ctx, args)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 880, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1404, in handle_parse_result
    self.callback, ctx, self, value)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 78, in invoke_param_callback
    return callback(ctx, param, value)
  File "/usr/local/lib/python2.7/dist-packages/flask/cli.py", line 693, in _validate_key
    is_context = isinstance(cert, ssl.SSLContext)
AttributeError: 'module' object has no attribute 'SSLContext'
```

Version information:

```
$ flask --version
Flask 1.0.2
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2]
```


tests/test_cli.py
```diff
@@ -539,3 +539,6 @@ def test_run_cert_import(monkeypatch):
     with pytest.raises(click.BadParameter):
         run_command.make_context(
             'run', ['--cert', 'ssl_context', '--key', __file__])
+
+
+def test_python_2_7_6(monkeypatch):
```
ThiefMaster: something is missing here :see_no_evil:

davidism:
They added a whole feature in a point release?!

ThiefMaster:
https://www.python.org/dev/peps/pep-0466/

PR #2796, opened 2018-05-22
Session cookie secure by default

iScrE4m:
I strongly believe that more secure options should be the default. I know this is a breaking change and might spark some controversy, as it would break loads of small hobby projects that lack https running on localhost etc. 

I'm not too sure about under which version I should edit changes.rst and which version versionchanged should mention, so I will update those later if you decide to go through with this change. Sorry for the imperfect PR for now!

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -206,7 +206,7 @@ The following configuration values are used internally by Flask:
     marked "secure". The application must be served over HTTPS for this to make
     sense.
 
-    Default: ``False``
+    Default: ``True``
 
 .. py:data:: SESSION_COOKIE_SAMESITE
 
--- a/flask/app.py
+++ b/flask/app.py
@@ -293,7 +293,7 @@ class Flask(_PackageBoundObject):
         'SESSION_COOKIE_DOMAIN':                None,
         'SESSION_COOKIE_PATH':                  None,
         'SESSION_COOKIE_HTTPONLY':              True,
-        'SESSION_COOKIE_SECURE':                False,
+        'SESSION_COOKIE_SECURE':                True,
         'SESSION_COOKIE_SAMESITE':              None,
         'SESSION_REFRESH_EACH_REQUEST':         True,
         'MAX_CONTENT_LENGTH':                   None,


miguelgrinberg:
> as it would break loads of small hobby projects

it would break the majority of projects when they are executed locally via `flask run` or `app.run()` during development. At the very least, this should be done only when the app runs in production mode.

> We can turn on the secure option when request.is_secure

Note that in many deployment situations the SSL connection does not reach the Flask app because it is terminated at a proxy server. In those cases, `request.is_secure` will apply to the connection between the proxy server and the Flask app which is not secure. You can apply the `ProxyFix` middleware along with some configuration in the proxy server to correct this, but in my experience very few people do this.

My concern with an automatic secure flag is that it may make people think they're secure, while in fact, with a default proxy server configuration they won't be.

davidism:
That breaks sites that want to work over both HTTP and HTTPS. Now, if they only wanted anonymous users with no stored state over HTTP, that would make sense, but it's not on Flask to decide that. The cookie itself is still signed and secure from tampering when sent over HTTP. If the site is served over HTTPS only (which it probably should be in 2018), then this flag doesn't really mean anything anyway and the default is still fine.

ThiefMaster:
> Note that in many deployment situations the SSL connection does not reach the Flask app because it is terminated at a proxy server. In those cases, request.is_secure will apply to the connection between the proxy server and the Flask app which is not secure. 

The frontend server is supposed to set `X-Forwarded-Proto` properly.

miguelgrinberg:
@ThiefMaster Yes, that is what I referred to when I said you need to apply "some configuration on the proxy server". Still, you have to use the ProxyFix middleware for those `X-Forwarded-` headers to be applied to the request object in the app.

@davidism the secure flag is still useful in 2018. Apps that run only on HTTPS have a rule that redirects any request on HTTP to the same URL on HTTPS, so they continue to accept HTTP requests. The browser will happily send your session cookie on a HTTP request if you don't make it secure. I don't disagree with your assessment, but the `secure` flag in the session cookie still makes sense to be set on HTTPS only apps.

davidism:
I am mostly concerned with this because it would affect the dev server. We can't rely on people enabling debug mode either, I gurantee we'll get questions about "the session is empty" if we try to enable this based on debug mode.

iScrE4m:
The way I see it, default behavior depends on what is the primary intended use for Flask. If the answer is user facing websites/applications served over web (-> in most cases going through public wifis etc.), then I hope Flask would default to the more 'safe' option (with the ability to override it). 

If I look back at myself starting, I'd expect the defaults to provide as much security as possible and if something wasn't working because I wasn't following the most secure pracitces, I would learn something new (and important), consider whether my use case really can break this. With the default being false, I'm thinking "smarter people than me wrote it" and it never comes into question. It would be really interesting to collect the data on how many publicly accessible flask based websites don't realize it and don't have the session cookie secure. It's not something you notice easily.

Personally I don't like to rely on dynamic behavior based on seemingly unrelated options like the all powerful debug mode, or in some cases unreliable `is_secure`. 

###### Somewhat unrelated fun fact: Until today I didn't realize that the session is only signed by the app key, I always thought everything in it was hashed with the key as a salt, meaning I could store secret-ish stuff in there.

miguelgrinberg:
Just for context, I checked a few other frameworks, and they all use the same `secure=false` default as Flask. I'm inclined to not make this change, it will cause unexpected failures that will be difficult to debug for most users.

References:

- Django: https://docs.djangoproject.com/en/2.0/ref/settings/#session-cookie-secure.
- Pyramid: https://docs.pylonsproject.org/projects/pyramid/en/latest/api/session.html#pyramid.session.SignedCookieSessionFactory
- aiohttp: https://aiohttp-session.readthedocs.io/en/stable/reference.html#aiohttp_session.AbstractStorage
- bottle: https://bottlepy.org/docs/dev/api.html#bottle.BaseResponse.set_cookie
- express.js: https://github.com/expressjs/session#cookiesecure
- Ruby on Rails: https://www.rubydoc.info/gems/rack/Rack/Session/Abstract/Persisted

I could not find any framework that defaults the secure flag on session cookies to true. Do you know any?

davidism:
Thansk @miguelgrinberg, was just about to ask about that.

davidism:
Based on Miguel's investigation, going to close this for now.

I'm not sure how users would even discover what's happening if this was changed. From their perspective all that happens is they never receive the cookie. The vast majority of users are new, I guarantee that most users, past me included, would have a very hard time determining that they weren't receiving cookies because of this setting. See the related issue of Chrome not storing cookies for localhost in some circumstances for evidence of that.

Secure-by-default is good, but I think the actual surface exposed by this is rather low compared to the issues it will cause during development. I'd rather people understand their system and security needs when they're deploying, and understand the configuration available to them.

---

The default cookie is secure from tampering, not from inspection. If you want to store private data in the session, you need to use a different session implementation, either an encrypted cookie or server-side storage. Maybe that can be made more clear in the docs?

PR #2797, opened 2018-05-23
Update cli.py

dayiguizhen:
SSLContext was introduced in 2.7.9.


davidism:
Nice catch, but this was already fixed in #2789, just haven't merged in 1.0-maintenance to master yet.

PR #2807, opened 2018-05-29
Test added for _AppCtxGlobals __repr__ method

eso31:
I followed the instructions of how to contribute and added a test that didn't have test coverage.

### DIFF
--- a/tests/test_appctx.py
+++ b/tests/test_appctx.py
@@ -159,6 +159,8 @@ def test_app_ctx_globals_methods(app, app_ctx):
     assert flask.g.pop('bar', 'more cake') == 'more cake'
     # __iter__
     assert list(flask.g) == ['foo']
+    #__repr__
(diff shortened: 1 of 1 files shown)


tests/test_appctx.py
```diff
@@ -159,6 +159,8 @@ def test_app_ctx_globals_methods(app, app_ctx):
     assert flask.g.pop('bar', 'more cake') == 'more cake'
     # __iter__
     assert list(flask.g) == ['foo']
+    #__repr__
+    assert flask.g.__repr__() == '<flask.g of \'flask_test\'>'
```
lepture: It should be `repr(flask.g)`.

davidism: Avoid escaping quotes if not necessary: `"<flask.g of 'flask_test'>"`.

davidism: If you add a test for when the app context isn't pushed as well, then we'll be covering both branches.

PR #2809, opened 2018-05-30
test_url_generation_with_parameter added

eso31:
I added a test for the url generation when a parameter is used.

### DIFF
--- a/tests/test_appctx.py
+++ b/tests/test_appctx.py
@@ -26,6 +26,17 @@ def test_basic_url_generation(app):
         rv = flask.url_for('index')
         assert rv == 'https://localhost/'
 
+def test_url_generation_with_parameter(app):
+    app.config['SERVER_NAME'] = 'localhost'
+    app.config['PREFERRED_URL_SCHEME'] = 'https'
+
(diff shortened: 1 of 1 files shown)


lepture:
Your commit is actually decreasing the test coverage since tests code are included in coverage. https://codecov.io/gh/pallets/flask/commit/ffe4acb667dfd78f28558d5eccbb38f7eabeec93

And a `route` should return something instead of `pass`.

ThiefMaster:
I don't think the return value is relevant when just using it to generate URLs and not actually call it.

Anyway, you could register the route without a function and `build_only=True`.

eso31:
Thank you for your comments, I tried to make a similar test with a different scenario and when I did the test coverage I didn't see that it actually decremented it. I also tried to register the route with build_only=True but I could test the url generation without the function.

davidism:
Thanks for working on this. This test doesn't seem necessary. Other tests around this are testing the `url_for` function specifically, this one tests Werkzeug's builder, which has already been tested. Even fixing the unused function, this wouldn't increase coverage.

Issue #2824, opened 2018-06-11
flask --version should output werkzeug version


jaapz:
### Expected Behavior

Running `flask --version` should output Flask, Python and Werkzeug versions, as those are the versions asked in the issue template on github.

### Actual Behavior

`flask --version` only outputs Flask and Python versions (and GCC, for some reason?).

### Environment

* Python version: 3.6.5
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

emil45:
I do not think it's a good idea to show users the `werkzeug` version. 
Most of the regular users of Flask don't know nothing about any `werkzeug` package.
Why do not you also display the version of `Jinja2` ?

jaapz:
Like I said in the issue, people need to provide the werkzeug version when they want to file an issue on github. I think adding that version in the `flask --version` command people can more easily obtain all versions they need to provide.

RyanSquared:
Perhaps it would be best to specify something like `--version --verbose` to print out the versions of _all_ "secondary" dependencies (in that they're required directly by Flask)?

adamfaliq42:
Werkzeug
Jinja
MarkupSafe
ItsDangerous
Click

davidism:
I'm not sure where the discussion is, but I deliberately did not include all the dependencies when the option was added. Some of them are inconsequential, any project will have multiple other dependencies that contribute to issues which must be pared down first to create an MCVE, and there's already a handy command to get all versions: `pip list`.

Adding Werkzeug is fine since I ask for that in the issue template, and there's already a PR open for that.

PR #2825, opened 2018-06-11
Fixes issue #2824 flask --version

eso31:
Fixes issue [2824](https://github.com/pallets/flask/issues/2824)

Now when you run ```flask --version``` command it shows you the Werkzeug version as well. I changed the order as well to be Python, Flask and Werkzeug as it is shown in the issue template.

Let me know if it has to be done in some other way.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -14,6 +14,7 @@ from __future__ import print_function
 import ast
 import inspect
 import os
+import platform
 import re
 import ssl
 import sys
@@ -21,7 +22,6 @@ import traceback
 from functools import update_wrapper
 from operator import attrgetter
 from threading import Lock, Thread
-import werkzeug
 
 import click
 from werkzeug.utils import import_string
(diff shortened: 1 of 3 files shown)


emil45:
It is best to remain consistent, the version of `Flask` imported in the global scope (line 28).
So `werkzeug` version should be also be imported in the global scope, rather than within the function.

As for the idea itself, I responded to you in issue.

flask/cli.py
```diff
@@ -21,6 +21,7 @@
 from functools import update_wrapper
 from operator import attrgetter
 from threading import Lock, Thread
+from werkzeug import __version__ as werkzeug_version
```
cdvv7788: What about importing werkzeug and referencing the version with `werkzeug.__version__`?
It is clear enough, and it is not as if it is being used anywhere else.

flask/cli.py
```diff
@@ -259,10 +260,11 @@ def locate_app(script_info, module_name, app_name, raise_if_not_found=True):
 def get_version(ctx, param, value):
     if not value or ctx.resilient_parsing:
         return
-    message = 'Flask %(version)s\nPython %(python_version)s'
+    message = 'Python %(python_version)s\nFlask %(version)s\nWerkzeug %(werkzeug_version)s'
```
cdvv7788: Should .format be preferred over %?

xavierhardy: Given the fact that it was originally there, I don't think so; it's not the point of this pull request. By the way, % is faster and more retro-compatible than .format.
https://hackernoon.com/a-closer-look-at-how-python-f-strings-work-f197736b3bdb

cdvv7788: Python2 is gonna die soon :P. F-strings are too recent.

xavierhardy: Sure, then make a separate pull request, since most of this code uses `%` over `.format`.

```sh
~/src/flask$ find ./ -name "*.py" -exec grep -H "%" {} \; | wc -l
90
~/src/flask$ find ./ -name "*.py" -exec grep -H "\.format" {} \; | wc -l
37
```
But again, this is not the point of this pull request.

PR #2826, opened 2018-06-13
add init-db call to flask tutorial

senarclens:
otherwise the user is confronted with an unpleasant error
message when accessing the dev server the first time

> Describe what this patch does to fix the issue.
no code change
added init-db call to flask tutorial's README.rst


doobeh:
Thanks for the submission @senarclens, cookie for you! 😄   :cookie:

PR #2832, opened 2018-06-16
Add `JSON_STRICT` config option

abought:
## Summary
Provide a new config option to control how JSON is serialized. This would prevent situations where Flask returns a response that is not actually JSON specification compliant.

This proposal is an opt-in mechanism. For backwards compatibility, it is disabled by default.

## Purpose
The default Python JSON serializer is not strictly [RFC-compliant](https://docs.python.org/3/library/json.html#standard-compliance-and-interoperability). In particular, it [outputs](https://docs.python.org/3/library/json.html#infinite-and-nan-number-values)
 `Infinity` and `NaN` values.

This means that by default, JSON APIs written in Flask may produce output that cannot be parsed by the browser, especially when dealing with numerical data.

Enabling the newly provided `JSON_STRICT` option would treat this scenario as an application error server-side and raise an exception. It would be up to the application developer to identify their preferred handling of such values.

## Changes and scope
This adds a new `JSON_STRICT` config option, adding to the [existing config flags](http://flask.pocoo.org/docs/1.0/config/#JSON_AS_ASCII) that also control serialization. 

The default value is chosen to leave existing applications unaffected. If `JSON_STRICT` is `True`, the application will raise an error in situations where it previously would have produced invalid JSON. 

## Previous discussion
This pitfall has been noted previously elsewhere. Others have expressed an interest in a simple, explicit configurable error option, but I am not aware of any previous pull requests that implemented the idea.

- https://coderwall.com/p/amb0va/python-s-json-module-does-not-always-produce-valid-json
- https://github.com/miguelgrinberg/Flask-SocketIO/issues/208

Particularly due to backwards compatibility concerns, the Python standard library has rejected previous suggestions to change the default `json` module behavior:
https://bugs.python.org/issue26105

## Evaluation of alternatives
Libraries such as `simplejson` [provide](https://simplejson.readthedocs.io/en/latest/#simplejson.dumps) an [ECMA-262 compliant](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-serializejsonproperty) `ignore_nan` option, in which out-of-range values are serialized as `null`.

However, this option is not supported by the Python standard library, and in practice the best choice of replacement value may be somewhat application specific.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -315,6 +315,14 @@ The following configuration values are used internally by Flask:
 
     Default: ``True``
 
+.. py:data:: JSON_STRICT
+
+    Raise an error when encountering values that cannot be serialized in a
+    strictly JSON compliant manner (such as ``NaN`` and ``Infinity``),
+    rather than encoding them as such.
+
+    Default: ``False``
+
 .. py:data:: JSONIFY_PRETTYPRINT_REGULAR
 
     ``jsonify`` responses will be output with newlines, spaces, and indentation
--- a/flask/app.py
+++ b/flask/app.py
@@ -304,6 +304,7 @@ class Flask(_PackageBoundObject):
         'PREFERRED_URL_SCHEME':                 'http',
         'JSON_AS_ASCII':                        True,
         'JSON_SORT_KEYS':                       True,
+        'JSON_STRICT':                          False,
         'JSONIFY_PRETTYPRINT_REGULAR':          False,
         'JSONIFY_MIMETYPE':                     'application/json',
         'TEMPLATES_AUTO_RELOAD':                None,
--- a/flask/json/__init__.py
+++ b/flask/json/__init__.py
@@ -102,6 +102,9 @@ def _dump_arg_defaults(kwargs):
         if not current_app.config['JSON_AS_ASCII']:
             kwargs.setdefault('ensure_ascii', False)
 
+        if current_app.config['JSON_STRICT']:
+            kwargs.setdefault('allow_nan', False)
+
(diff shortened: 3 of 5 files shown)


davidism:
We already support simplejson if it's installed. This seems like a good case for a custom serializer rather than a configuration option.

abought:
Thanks for your comment. If you prefer a refined or alternate approach would, I would be happy to iterate on this and follow up...?

The key idea is that by default, a web application framework is producing JSON output that cannot be parsed by web browsers. I think there is a niche within flask core for a simple, obvious, and configurable way to warn about this.

The current mechanisms in flask are powerful, but only after the fact. Before a developer can write a custom serializer, first they must know the issue is occurring. The current default behavior is hard to track down, because it involves special values like `infinity` and the application claims to be acting normally.

Per your other note: although  flask uses simplejson **if installed**, it does not expose or enable the relevant underlying options. (Mainly because functions such as [jsonify](http://flask.pocoo.org/docs/1.0/api/#flask.json.jsonify) treat kwargs as data, rather than options)

davidism:
I'd prefer not to add another config option. You can accomplish the same thing with:

```python
from flask.json import JSONEncoder

class StrictEncoder(JSONEncoder):
    def __init__(self, *args, ignore_nan=True, **kwargs):
        kwargs["ignore_nan"] = ignore_nan
        super().__init__(*args, **kwargs)

app.json_encoder = StrictEncoder
```

davidism:
Basically, if your app happens to produce these, and you don't want it to, then you should be sanitizing your data before serialization anyway. Either you do this and raise a mysterious 500 error, or you ensure your data is correct before sending it.

PR #2834, opened 2018-06-17
Remove unnecessary period in documentation

greyli:
Except for this one, I also notice that there are [two sentences](http://flask.pocoo.org/docs/dev/testing/#testing-cli-commands) end with period instead of colon before a code snippet, should we change it?


davidism:
I generally don't like the colon end style, so those other two were on purpose.

PR #2835, opened 2018-06-18
Fix pep8 no blank line issue

pradeeps:
Fixed pep8 no blank line issue in flask/_compat.py

### DIFF
--- a/flask/_compat.py
+++ b/flask/_compat.py
@@ -83,6 +83,7 @@ if hasattr(sys, 'pypy_version_info'):
     class _Mgr(object):
         def __enter__(self):
             return self
+        
         def __exit__(self, *args):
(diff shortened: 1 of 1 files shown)


ThiefMaster:
adding this linebreak doesn't really improve anything. it's just `_compat.py` anyway...

pradeeps:
Why can't we add this even if this not an improvement? just adhering to pep8...

davidism:
Thank you for your interest. We generally do not accept style changes. If you would like to contribute, consider triaging issues, adding test coverage, or improving documentation.

ThiefMaster:
In modules that solely serve as compatibility layers with different Python versions you usually go for concise solutions, not for nice spacing everywhere.

PR #2840, opened 2018-06-26
Add Dreamhost deploy doc

ogdenstudios:
This PR adds a short .rst doc that covers the process for deploying a Flask app to Dreamhost, specifically. This configuration works for any project that follows the Flaskr tutorial template (in that it makes an app generator method in the module). There are some links to Dreamhost's documentation on installing Python 3/Virtualenv on their servers, I kept them as external links rather than pulling them in to the documentation to make sure they stay accurate if DH changes anything.

No tests since it's just documentation.

### DIFF
--- a/docs/deploying/dreamhost.rst
+++ b/docs/deploying/dreamhost.rst
@@ -1,0 +1,66 @@
+Dreamhost Setup
+===============
+
+These instructions are specifically for deploying to Dreamhost.
+
+1. Add hosting to a new domain in the ``Manage Domains`` tab of your
+   control panel.
+
+2. During setup, click the ``Passenger (Ruby/NodeJS/Python apps only):``
+   checkbox to enable Phusion Passenger on your domain.
+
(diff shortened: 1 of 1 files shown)


ogdenstudios:
Oh word that makes sense! Thanks for considering it, anyway.

davidism:
I don't mind adding something for Passenger in general, but I don't want to add things for specific hosting companies.

ogdenstudios:
Yea, that makes complete sense. I personally had a hard time stumbling through Dreamhost's documentation + the deploy docs with Flask. Had to kind of piece them together, and then I had to deploy again and had forgotten the procedure. Writing it down helped a lot for me and I thought others might find it useful. Someone on the mailing list was into it, so I made the PR. But it makes a ton of sense not to tie ourselves to Dreamhost's procedures if they change. Would y'all be willing to link to this as an external resource somewhere that I host separately? I sent to DH but I don't expect a prompt response or anything.

PR #2845, opened 2018-06-28
Add `status` kwarg into jsonify()

fourjr:
Currently, there is no way to modify the default 200 status in `jsonify`. This PR adds a kwarg for it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,10 @@ Version 1.1
 
 Unreleased
 
+-   :func:`jsonify` allows for a ``status`` kwarg (`#2845`_)
+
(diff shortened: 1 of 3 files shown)


davidism:
```python
response = jsonify(data)
response.status_code = 404
return response
```

greyli:
Besides, you can also return a tuple like this:
```py
return jsonify(data), 400
```

tautology0:
@greyli you can't do a tuple as it returns a Response object, you get the error:
`TypeError: <Response 35 bytes [200 OK]> is not JSON serializable`

@davidism this produces manky code as you require three lines rather than one. This means that if you do full boundary checking, half your code base ends up mangling around this strange jsonify restriction. You can't even template it as it requires a request context.

PR #2849, opened 2018-06-29
:art: Refactor __init__.py imports

gcrsaldanha:
This pull request does the following changes for simplicity and elegance:
- Remove unnecessary comment about signals section
- Import and expose `jsonify` in a single line: following the other imports pattern (see line 21 to 32):
```python
from .module import value
```
- Also, simplify the comment about exposing `json.jsonify`
- Make the imports more [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)-like using parenthesis over back-slash (`\`)

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

I'm not sure if changes like these - prettifying - should be preceded by an issue. If so, feel free to close this pull requests and I will file an issue asap. Thanks!

### DIFF
--- a/flask/__init__.py
+++ b/flask/__init__.py
@@ -20,19 +20,19 @@ from jinja2 import Markup, escape
 
 from .app import Flask, Request, Response
 from .config import Config
-from .helpers import url_for, flash, send_file, send_from_directory, \
-     get_flashed_messages, get_template_attribute, make_response, safe_join, \
-     stream_with_context
-from .globals import current_app, g, request, session, _request_ctx_stack, \
-     _app_ctx_stack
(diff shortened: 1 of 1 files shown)


davidism:
Please stop using GitHub emoji placeholders in titles, they don't render. (Or just emoji in general for commits.)

davidism:
Also, we don't accept style PRs. I will be handling formatting at some point in the future.

gcrsaldanha:
@davidism sorry about the emoji in the title! Also, I didn't know about style PRs, thanks!

PR #2850, opened 2018-06-29
Fix grammatical error: incorrect verb conjugation

billfienberg:
Before: 
> However, as a project **get** bigger

After: 
>However, as a project **gets** bigger


ITchen09:
Not very good at GitHub,Can you teach me

PR #2851, opened 2018-07-01
Update Copyright Date to 2018

MohamedAlFahim:
I am not sure if this is required, but it seems to be a safe idea.


davidism:
This is not how copyright works. This just causes noise.

PR #2855, opened 2018-07-08
use ipython when `flask shell`

Ehco1996:
try use `ipython` as defualt shell when run cmd `flask shell`


davidism:
`pip install flask-shell-ipython`

davidism:
We decided not to add shell integrations internally. A developer can choose their preferred shell with a plugin.

PR #2858, opened 2018-07-10
Remove check against deprecated property is_xhr

jcomo:
This eliminates the deprecation warning fired by werkzeug's `Request.is_xhr` property for older versions of Flask by removing the check.

https://github.com/pallets/werkzeug/blob/da57cd54c97909856e05f490418000f9ab851358/werkzeug/wrappers.py#L696-L702

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,12 @@ Flask Changelog
 ===============
 
 
+Version 0.12.5
+--------------
+
+-   Remove check against deprecated property ``Request.is_xhr``.
+
+
 Version 0.12.4
 --------------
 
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Use Flask 1.0, it is already changed there.

jcomo:
Easier said than done. It would involve migrating a large codebase to a new major version. Not sure that's the right solution here.

ThiefMaster:
you could subclass Request and use a version of the property that doesn't show the warning.

but fwiw, chances are good that switching to 1.0 would be pretty painless!

jcomo:
We _could_ do that. What I'm trying to understand is why you're against putting into the maintenance branch.

davidism:
Because 0.12 is only getting security fixes at this point.

Silence the warning if you don't want to see the warning. Python's `warning` module describes how to do this.

davidism:
Or better yet, pay attention to the warning and stop using `request.is_xhr` since it's unreliable.

PR #2859, opened 2018-07-11
update README.rst / "docs/tutorial/index.rst"

y2kbcm1:
-clarified some sentences.
-fixed some confusing nominalizations.
-made some phrase concise.
-in "... provided by the community," it is confusing because there is no community mentioned beforehand.


davidism:
This just seems different, not better.

PR #2871, opened 2018-07-24
update description for database.rst become more understandable

cktan98:
update the contents become more understandable


davidism:
Thanks, but I don't think this is clearer, just different. Many of the changes are less grammatical.

PR #2874, opened 2018-07-26
fix little grammar detail in flaskr/db.py

feute:
Just add a missing 'is' in the docstring in db.py, from the flaskr example.

> If this request connected to the database

This should be: If this request **is** connected to the database.


davidism:
It's correct as-is.

feute:
Now I see it; it was kind of confusing, like something was missing there.

davidism:
Thanks for looking at it though. :-)

PR #2889, opened 2018-08-20
Improve the recommended way to use Celery

ArtemGordinsky:
Currently [docs](http://flask.pocoo.org/docs/1.0/patterns/celery/) are recommending to call `run()` method after creating a Flask context. It seems that calling `super().__call__()` instead would be a better approach, because calling run() directly skips Celery's internal threading stack management:

https://github.com/celery/celery/blob/master/celery/app/task.py#L382
```python
    def __call__(self, *args, **kwargs):
        _task_stack.push(self)
        self.push_request(args=args, kwargs=kwargs)
        try:
            return self.run(*args, **kwargs)
        finally:
            self.pop_request()
            _task_stack.pop()
```

I've applied this change in our project and can confirm that it still works without any issues.

### DIFF
--- a/docs/patterns/celery.rst
+++ b/docs/patterns/celery.rst
@@ -49,7 +49,7 @@ This is all that is necessary to properly integrate Celery with Flask::
         class ContextTask(celery.Task):
             def __call__(self, *args, **kwargs):
                 with app.app_context():
-                    return self.run(*args, **kwargs)
+                    return super().__call__(*args, **kwargs)
 
         celery.Task = ContextTask
         return celery


davidism:
We already went through this in reverse in #2055. After extensive testing and digging through the code, I came to the conclusion that what is currently shown is correct. You'll need an authoritative reason that using `__call__` instead of `run` is actually correct in order for me to change this again. I'm not willing to investigate this myself again.

davidism:
While the Celery code you've linked *appears* to indicate that your change is correct, from what I remember there's actually code in a seemingly unrelated area that monkeypatches it in other situations, and that messes up things with your change.

davidism:
Closing this for now. I think my previous conclusion is still valid. If you get a definitive answer, comment here and we can revisit it.

PR #2904, opened 2018-09-11
add badges to README

AlmogCohen:
Add the following badges for easier access: travis, appveyor, coverage,
supported pypi versions, pypi version. All badges include links to the
proper urls. Badges also add better visibility into our code coverage
as well as give better visibility into the build status.

If the contributors, we can also simplify the docs by removing some
of the links at the bottom that are now included in the badges.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,38 @@
 Flask
 =====
 
+
+.. start-badges
+
+.. list-table::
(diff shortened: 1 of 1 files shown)


AlmogCohen:
So be it...
I still believe there is an added value in the graphicality of it and better accessibility for build status etc...

PR #2914, opened 2018-09-19
Reduce unnecessary function calls.

garenchan:
When propagate exceptions, the function call of "_find_error_handler" makes no sense.
https://github.com/pallets/flask/blob/0b7a74dda3284b03fac49851eaf130e876f183bd/flask/app.py#L1733-L1748

### DIFF
--- a/flask/app.py
+++ b/flask/app.py
@@ -1728,9 +1728,7 @@ class Flask(_PackageBoundObject):
         .. versionadded:: 0.3
         """
         exc_type, exc_value, tb = sys.exc_info()
-
         got_request_exception.send(self, exception=e)
-        handler = self._find_error_handler(InternalServerError())
 
         if self.propagate_exceptions:
             # if we want to repropagate the exception, we can attempt to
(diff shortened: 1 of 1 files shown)


docs/errorhandling.rst
```diff
@@ -78,7 +78,7 @@ This can be done by registering error handlers.
 
 An error handler is a normal view function that return a response, but instead
```
henrycjc: While you're at it - can you correct the grammar? It should be "that return`s` a response "

garenchan: OK. I'll do that. Thanks!

flask/app.py
```diff
@@ -1611,12 +1611,17 @@ def _find_error_handler(self, e):
         """
         exc_class, code = self._get_exc_class_and_code(type(e))
 
+        _order = []
         for name, c in (
             (request.blueprint, code), (None, code),
             (request.blueprint, None), (None, None)
         ):
-            handler_map = self.error_handler_spec.setdefault(name, {}).get(c)
+            if (name, c) in _order:
```
davidism: The other change further down is fine, but I don't find this one necessary.

garenchan: If I can use orderedset, it wouldn't be so ugly!

garenchan:
@davidism, I have deleted the unnecessary changes you pointed out.

PR #2915, opened 2018-09-19
fix @click.command example by adding parens

loomchild:
Otherwise the example fails with the following error: 
"name = name or cmd.name AttributeError: 'function' object has no attribute 'name'". 

More details: https://stackoverflow.com/a/51923415/4619705


davidism:
Cherry picked to 1.0-maintenance as well: 363205bdc3abcfeed08c7e85e397fb7a7dc9ef07

PR #2916, opened 2018-09-19
Make script output into comments to not break syntax highlighting

tombusby:
http://flask.pocoo.org/docs/1.0/quickstart/#url-building

In the URL Building section of the quickstart, the script output is shown right after the code.

This breaks syntax highlighting for the whole code-block. To fix this, I have made the output into comments.


tombusby:
An alternative could be to break up the block with "This produces the output:" or something.

davidism:
Split in 0b7a74dda3284b03fac49851eaf130e876f183bd

Issue #2900, opened 2018-09-02
avoid creating a new app context when using json argument in test client


mbarakaja:
At this moment, when a Flask test client method uses the **json** argument, a new application context is created in the line below.

https://github.com/pallets/flask/blob/b9b88b0cdf70190dee5c7d31a83efb1cae888a63/flask/testing.py#L82

The problem is that when the above created context goes out of scope, anything registered to `teardown_appcontext` is called. This is the case of Flask-SQLAlchemy, which remove the current session when that happens.

This is problematic is some tests scenarios.

I am using PostgreSQL SAVEPOINT to rollback any changes made by a test case. My approach is similar to this example provided by [SQLAlchemy here](http://docs.sqlalchemy.org/en/latest/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites), so I'm not going to put all my code here, just a shallow example.

I have a class that creates a nested transaction and rollback everything when needed.

```python
client = app.test_client()
trn = TransactionManager(database)

with app.app_context():
    trn.start() # start a new nested transaction

    client.post('/products', data=json.dumps({'name': 'Cake'}))
    client.get('/products')
    
    trn.end() # Rollback before the session is removed.
```

The above works well because everything happens inside the same transaction and each test client interaction can see changes made to the database by previous client calls.

But using the **`json`** argument, the POST call will first create an application context, which immediately will go out of scope making Flask-SQLAlchemy's `teardown_appcontext` registered function **remove the current session before the next client call**, making each database interaction in a different transaction.

Why do we need an application context to dump the JSON string?
https://github.com/pallets/flask/blob/b9b88b0cdf70190dee5c7d31a83efb1cae888a63/flask/testing.py#L83

I don't understand, but if the the above logic is necessary I think the code should check first if there is an application context pushed already using **`has_app_context`**.

Am I doing something wrong or I am missing something?

kumy:
Hi, I passed hours trying to find why my blended objects were detached from the session just after using `get()` function. I finally reached to the same problematic line. Removing the context creation or replacing it with `test_request_context` fixed my problem.

What do Flask guru think about this part of code?
https://github.com/pallets/flask/blob/c92001d2fbaf4d5aa43bd44e56416ed27313377c/flask/testing.py#L81-L86

garenchan:
I think this may be better than using ``test_request_context``:
```
from flask import current_app

try:
    push_ctx = not current_app or current_app != app
    if push_ctx:
[...]

PR #2921, opened 2018-09-23
Fix issue #2900: not use app_context context manager while accessing configuration

garenchan:
If we just want to access configuration, it's important to not use push
and pop methods of the actual app context object since that would mean
the cleanup handlers will be called. And app context object implements 
"context management protocol" to facilitate the use of these two methods.

Fix issue #2900.


davidism:
This is not the approach I want to take. I'm not sure what the correct solution is, but messing with the internal structures to bypass behavior is not ideal.

Hooksie:
@davidism I'll go ahead and write the patch as described above, assuming it passes testing that way.

PR #2922, opened 2018-09-23
Add default returns for Request wrapper class properties

stephen-st:
The `Request` wrapper class contains properties that only return on valid logic checks. If the logic fails, a reference to the property object is returned instead of a value that is valid within the context of the property:

```python
# pytest source (abridged)
        unpatched_size = app.request_class.max_content_length
...
        assert unpatched_size == DEFAULT_UPLOAD_SIZE

# pytest output
...
>       assert unpatched_size == DEFAULT_UPLOAD_SIZE
E       assert <property object at 0x7fbb9ee964a8> == 67108864
```

This aims to tidy this up by always returning None or something relevant to the property, like an empty string.

This also avoids misleading assertion failures like "the `max_content_length` property is not X" rather than "None is not X".

### DIFF
--- a/flask/wrappers.py
+++ b/flask/wrappers.py
@@ -145,7 +145,7 @@ class Request(RequestBase, JSONMixin):
     @property
     def max_content_length(self):
         """Read-only view of the ``MAX_CONTENT_LENGTH`` config key."""
-        return current_app.config.get('MAX_CONTENT_LENGTH') if current_app else None:
+        return current_app.config.get('MAX_CONTENT_LENGTH') if current_app else None
 
     @property
     def endpoint(self):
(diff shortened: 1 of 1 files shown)


stephen-st:
Hi @garenchan, thanks for your feedback. You're right, I misunderstood how the `Request` class and its properties are and should be used. I played around with it some more and now have a much better understanding of where I went wrong.

Perhaps this PR only involves stylistic changes for readability then. Based on other PRs, it looks like that may not be a valid reason for introducing changes. However, I'd be happy to squash my commits to reduce history bloat if the moderators like them.

davidism:
This doesn't change anything significant, and the `None` change breaks existing behavior. We don't normally accept style changes like this, but thank you for the interest.

PR #2928, opened 2018-09-28
Fix #2926 `Default port overrides requesting a system assigned port`

### LINKED ISSUE #2926: Default port overrides requesting a system assigned port
### Expected Behavior

When running a flask application with `app.run(port=0)`, flask should pass the '0' port down to werkzeug to use a port assigned by the operating system.

### Actual Behavior

The logic in flask/app.py:932 sets the port to the default 5000, which can cause port conflicts, which is the exact situation that setting `port=0` should avoid. This is a regression from the behaviour exhibited in prior versions.

```pytb
>>> app.run(port=0)
  File "<stdin>", line 1, in <module>
  File "/.../lib/python3.6/site-packages/flask/app.py", line 943, in run
    run_simple(host, port, self, **options)
  File "/.../lib/python3.6/site-packages/werkzeug/serving.py", line 814, in run_simple
    inner()
  File "/.../lib/python3.6/site-packages/werkzeug/serving.py", line 774, in inner
    fd=fd)
  File "/.../lib/python3.6/site-packages/werkzeug/serving.py", line 660, in make_server
    passthrough_errors, ssl_context, fd=fd)
  File ".../lib/python3.6/site-packages/werkzeug/serving.py", line 577, in __init__
    self.address_family), handler)
  File "/.../lib/python3.6/socketserver.py", line 453, in __init__
    self.server_bind()
  File "/.../lib/python3.6/http/server.py", line 136, in server_bind
    socketserver.TCPServer.server_bind(self)
  File "/.../lib/python3.6/socketserver.py", line 467, in server_bind
    self.socket.bind(self.server_address)
OSError: [Errno 98] Address already in use
```

### Environment

* Python version: 3.6
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

vorelq:
Fix #2926 

In `app.py`, compare possible port values with `None` explicitly, instead of relying on bool casting. Thanks to this, `0` is treated as legitimate port value, and not overwritten by the default value.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,8 @@ Unreleased
     :pr:`3195`
 -   Allow customizing the :attr:`Flask.url_map_class` used for routing.
     :pr:`3069`
+-   The development server port can be set to 0, which tells the OS to
+    pick an available port. :issue:`2926`
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
 .. _#2957: https://github.com/pallets/flask/issues/2957
--- a/flask/app.py
+++ b/flask/app.py
@@ -972,7 +972,8 @@ class Flask(_PackageBoundObject):
             sn_host, _, sn_port = server_name.partition(":")
 
         host = host or sn_host or _host
-        port = int(port or sn_port or _port)
+        # pick the first value that's not None (0 is allowed)
+        port = int(next((p for p in (port, sn_port) if p is not None), _port))
 
         options.setdefault("use_reloader", self.debug)
         options.setdefault("use_debugger", self.debug)
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1916,21 +1916,24 @@ def test_run_server_port(monkeypatch, app):
 
 
 @pytest.mark.parametrize(
-    "host,port,expect_host,expect_port",
+    "host,port,server_name,expect_host,expect_port",
     (
-        (None, None, "pocoo.org", 8080),
-        ("localhost", None, "localhost", 8080),
-        (None, 80, "pocoo.org", 80),
-        ("localhost", 80, "localhost", 80),
+        (None, None, "pocoo.org:8080", "pocoo.org", 8080),
+        ("localhost", None, "pocoo.org:8080", "localhost", 8080),
+        (None, 80, "pocoo.org:8080", "pocoo.org", 80),
+        ("localhost", 80, "pocoo.org:8080", "localhost", 80),
+        ("localhost", 0, "localhost:8080", "localhost", 0),
+        (None, None, "localhost:8080", "localhost", 8080),
+        (None, None, "localhost:0", "localhost", 0),
     ),
 )
-def test_run_from_config(monkeypatch, host, port, expect_host, expect_port, app):
+def test_run_from_config(monkeypatch, host, port, server_name, expect_host, expect_port, app):
     def run_simple_mock(hostname, port, *args, **kwargs):
         assert hostname == expect_host
         assert port == expect_port
 
     monkeypatch.setattr(werkzeug.serving, "run_simple", run_simple_mock)
-    app.config["SERVER_NAME"] = "pocoo.org:8080"
+    app.config["SERVER_NAME"] = server_name
     app.run(host, port)
 
 


flask/app.py
```diff
@@ -929,7 +929,11 @@ def run(self, host=None, port=None, debug=None,
             sn_host, _, sn_port = server_name.partition(':')
 
         host = host or sn_host or _host
-        port = int(port or sn_port or _port)
+
+        for possible_port_value in [port, sn_port, _port]:
```
ThiefMaster: This could be simplified with something like this:

    port = next((int(x) for x in (port, sn_port, _port) if x is not None), None)

Not sure about the default value if all three are None.

vorelq: thanks for the suggestion @ThiefMaster ! I was considering doing a one-liner but decided not to, mainly because I was worried about readability... But your version doesn't looks pretty nice, definitely worth doing this for getting rid of 3 LOC, I'll amend it in a sec. I keep forgetting that `next` has this default value parameter, that's super useful.

As for default value, let's probably stick to how the old code handled it - `int(_port)` was the default if both `port` and `sn_port` were false-y.

vorelq:
how about now @ThiefMaster ? I left the `_port` value out of that `is None` condition and always use it as default value (expecting it's a valid value), mostly because that's what the old code did.

davidism:
It looks like you haven't configured your name and email in Git. If you'd like that to show up, be sure to configure it and amend your commit.

vorelq:
will do, and also squash these 2 commits into one. Having a separate commit just for that test seems unnecessary

vorelq:
thanks both @davidism and @ThiefMaster ! anything else I need to do?

CHANGES.rst
```diff
@@ -20,9 +20,11 @@ Unreleased
     stricter about header encodings than PEP 3333. (`#2766`_)
 -   Allow custom CLIs using ``FlaskGroup`` to set the debug flag without
     it always being overwritten based on environment variables. (`#2765`_)
+-   Fix issue with port set to 0 being overriden by the default value
 
 .. _#2766: https://github.com/pallets/flask/issues/2766
 .. _#2765: https://github.com/pallets/flask/pull/2765
+.. _#2926: https://github.com/pallets/flask/issues/2926
```
davidism: This is the wrong link, this PR would be `.pull/2928`. And needs to referenced on the entry line above. That sentence should end with a period.

vorelq: Adjusted. Just out of curiosity, how do I know if PR or issue should be linked?

davidism:
Just noticed you committed this against master, but it needs to be on 1.0-maintenance.

vorelq:
thanks, fixed. Branch and PR are now against `1.0-maintenance`

flask/app.py
```diff
@@ -929,7 +929,7 @@ def run(self, host=None, port=None, debug=None,
             sn_host, _, sn_port = server_name.partition(':')
 
         host = host or sn_host or _host
-        port = int(port or sn_port or _port)
+        port = int(next((p for p in (port, sn_port) if p is not None), _port))
```
ocamlmycaml: this is kinda cryptic and hard to understand. I'd suggest
`int(next(p for p in (port, sn_port, _port) if p is not None))`
I guess that also allows `sn_port` to be 0, so if you use that, you should add a test for that. Alternatively you could specifically allow `port` alone to be 0 using:
`int(port) if port is not None else int(sn_port or _port)`

garenchan:
CI build failed because of an issue that has been solved on the master branch. So if cherry-pick is still unfinished, this problem will persist.
image

flask/app.py
vorelq: Hi @ocamlmycaml, sorry for the long delay.

I added a test for `sn_port` being zero in https://github.com/pallets/flask/pull/2928/commits/583da58419e82adc672b5565587d143fbf5b66a6

regarding readability, I kinda agree with you:

```
int(next(p for p in (port, sn_port, _port) if p is not None))
```

is shorter and more intuitive than

```
int(next((p for p in (port, sn_port) if p is not None), _port))
```

I'm having a little problem with your suggestion because it changes the behavior of original code, checking also if `_port` is not None.

How about I rename `_host` and `_port` to `_default_*`? That would make it 

```
int(next((p for p in (port, sn_port) if p is not None), _default_port))
```

which IMO shows the purpose a little better.

PR #2930, opened 2018-10-01
Tutorial: Stop swallowing all OSErrors when creating the instance folder

Nechepso:
Improve error reporting of the application factory built during the tutorial by not swallowing all `OSError` exceptions that could happen when trying to create the instance folder.

This PR is inspired by the following SO entry: https://stackoverflow.com/a/273227

AFAIK this is portable code (see https://stackoverflow.com/a/273872 for a discussion on errno's portability), and I tested it under Linux and Windows.

### DIFF
--- a/docs/tutorial/factory.rst
+++ b/docs/tutorial/factory.rst
@@ -35,6 +35,7 @@ directory should be treated as a package.
 .. code-block:: python
     :caption: ``flaskr/__init__.py``
 
+    import errno
     import os
 
     from flask import Flask
@@ -58,8 +59,9 @@ directory should be treated as a package.
         # ensure the instance folder exists
         try:
             os.makedirs(app.instance_path)
-        except OSError:
-            pass
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
 
         # a simple page that says hello
         @app.route('/hello')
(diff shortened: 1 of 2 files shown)


docs/tutorial/factory.rst
```diff
@@ -174,4 +176,4 @@ Visit http://127.0.0.1:5000/hello in a browser and you should see the
 "Hello, World!" message. Congratulations, you're now running your Flask
 web application!
 
-Continue to :doc:`database`.
+Continue to :doc:`database`.
```
ThiefMaster: please remove this change. files should end with a linebreak

Nechepso:
First, I'm impressed by your reactivity. :+1:

I understand your concerns: this change makes that area of the tutorial more complex (and probably needlessly).

Besides, I just convinced myself it was insufficient: `EEXIST` is used to signal that _something_ already exists, but not necessarily a folder. I didn't know of `exists_ok`, but I found an interesting bug report about it: https://bugs.python.org/issue13498. Things now seem much less simple to me... :-)

In the end, as you asked, what really matters was if I had a precise use case in mind. And... no. ^^ So I think you're right, the tutorial is better as it is now than with this change (feel free to close this PR, or should I?)

Thanks for the review!

davidism:
Thanks for your interest! I'll close this, but I'm happy to review any PRs in the future.

PR #2936, opened 2018-10-09
Fix #2935: Copy current session object in copy_current_request_context

### LINKED ISSUE #2935: copy_current_request_context copies old session object
### Expected Behavior

When using `copy_current_request_context` in the middle of a request, I'm expecting it to copy the current state of `flask.request` and `flask.session` at the time it's called. 

### Actual Behavior

The current state of `flask.request` is copied, but `flask.session` is from before the current request and doesn't contain changes made during the current request, before `copy_current_request_context` was called. To me this seems like unexpected behaviour. If it's not I'd like to understand a bit more about why, and also to find out whether there's any workaround or alternative approach I can take. 

### Example

Here's a fully functional example you can save and run. It adds a test value to `flask.request` and `flask.session` during the request, then uses `copy_current_request_context` to decorate a method that will retrieve those same values in another thread. Two requests are made; on the first request the copied context can find the new value added to `flask.request` but cannot find the value added to `flask.session`. On the second request, the copied context can find the value added to `flask.request`, and now finds the value that was previously added to `flask.session` in the first request.

```python
import concurrent.futures
import random

from flask import Flask, copy_current_request_context, request, session


app = Flask(__name__)
app.config['SECRET_KEY'] = 'test'
executor = concurrent.futures.ThreadPoolExecutor()


@app.route('/')
def session_context():
    test_value = random.randint(1, 1001)
    request.test_value = test_value
    session['TEST_VALUE'] = test_value

    original_context = (
        ('request', request.test_value),
        ('session', session.get('TEST_VALUE'))
    )

    @copy_current_request_context
    def debug_session():
        return (
            ('request', request.test_value),
            ('session', session.get('TEST_VALUE'))
        )

    future = executor.submit(debug_session)
    print('original_context:', original_context)
    print('copied_context:', future.result())
    return 'ok'


if __name__ == '__main__':
    client = app.test_client()
    print('### First request ###')
    client.get('/')
    print('### Second request ###')
    client.get('/')
```

Example output:

```bash
$ python session_context.py
### First request ###
original_context: (('request', 27), ('session', 27))
copied_context: (('request', 27), ('session', None))
### Second request ###
original_context: (('request', 63), ('session', 63))
copied_context: (('request', 63), ('session', 27))
```

### Environment

* Python version: 3.7.0
* Flask version: Flask==1.0.2
* Werkzeug version: Werkzeug==0.14.1

Also, for whatever it's worth: I'm investigating this as part of improving this project: https://github.com/dchevell/flask-executor

dchevell:
Updates `flask.copy_current_request_context` to add a copy of the current session object to the request context copy. This prevents `flask.session` pointing to an out-of-date object.

- [x] add tests that fail without the patch
- [x] ensure all tests pass with ``pytest``
- [x] add documentation to the relevant docstrings or pages
- [x] add ``versionadded`` or ``versionchanged`` directives to relevant docstrings
- [x] add a changelog entry if this patch changes code

Please refer to #2935 for background on this

### DIFF
--- a/flask/ctx.py
+++ b/flask/ctx.py
@@ -122,8 +122,8 @@ def copy_current_request_context(f):
     """A helper function that decorates a function to retain the current
     request context.  This is useful when working with greenlets.  The moment
     the function is decorated a copy of the request context is created and
-    then pushed when the function is called.  A copy of the current session is
-    also included in the copied request context.
+    then pushed when the function is called.  The current session is also
+    included in the copied request context.
 
     Example::
 
@@ -324,8 +324,8 @@ class RequestContext(object):
         .. versionadded:: 0.10
 
         .. versionchanged:: 1.1
-           A copy of the current session object is added to the request context
-           copy. This prevents `flask.session` pointing to an out-of-date object.
+           The current session object is used instead of reloading the original
+           data. This prevents `flask.session` pointing to an out-of-date object.
         """
         return self.__class__(self.app,
             environ=self.request.environ,
--- a/flask/ctx.py
+++ b/flask/ctx.py
@@ -14,7 +14,7 @@ from functools import update_wrapper
 
 from werkzeug.exceptions import HTTPException
 
-from .globals import session, _request_ctx_stack, _app_ctx_stack
+from .globals import _request_ctx_stack, _app_ctx_stack
 from .signals import appcontext_pushed, appcontext_popped
 from ._compat import BROKEN_PYPY_CTXMGR_EXIT, reraise
 
@@ -141,10 +141,6 @@ def copy_current_request_context(f):
             return 'Regular response'
 
     .. versionadded:: 0.10
-
-    .. versionchanged:: 1.0.3
-       A copy of the current session object is added to the request context
-       copy. This prevents `flask.session` pointing to an out-of-date object.
     """
     top = _request_ctx_stack.top
     if top is None:
@@ -152,7 +148,6 @@ def copy_current_request_context(f):
             'when a request context is on the stack.  For instance within '
             'view functions.')
     reqctx = top.copy()
-    reqctx.session = session.copy()
     def wrapper(*args, **kwargs):
         with reqctx:
             return f(*args, **kwargs)
@@ -282,14 +277,14 @@ class RequestContext(object):
     that situation, otherwise your unittests will leak memory.
     """
 
-    def __init__(self, app, environ, request=None):
+    def __init__(self, app, environ, request=None, session=None):
         self.app = app
         if request is None:
             request = app.request_class(environ)
         self.request = request
         self.url_adapter = app.create_url_adapter(self.request)
         self.flashes = None
-        self.session = None
+        self.session = session
 
         # Request contexts can be pushed multiple times and interleaved with
         # other request contexts.  Now only if the last level is popped we
@@ -327,10 +322,15 @@ class RequestContext(object):
         request object is locked.
 
         .. versionadded:: 0.10
+
+        .. versionchanged:: 1.1
+           A copy of the current session object is added to the request context
+           copy. This prevents `flask.session` pointing to an out-of-date object.
         """
         return self.__class__(self.app,
             environ=self.request.environ,
-            request=self.request
+            request=self.request,
+            session=self.session
         )
 
     def match_request(self):
--- a/flask/ctx.py
+++ b/flask/ctx.py
@@ -122,7 +122,8 @@ def copy_current_request_context(f):
     """A helper function that decorates a function to retain the current
     request context.  This is useful when working with greenlets.  The moment
     the function is decorated a copy of the request context is created and
-    then pushed when the function is called.
+    then pushed when the function is called.  A copy of the current session is
+    also included in the copied request context.
 
     Example::
 
@@ -133,13 +134,17 @@ def copy_current_request_context(f):
         def index():
             @copy_current_request_context
             def do_some_work():
-                # do some work here, it can access flask.request like you
-                # would otherwise in the view function.
+                # do some work here, it can access flask.request or
+                # flask.session like you would otherwise in the view function.
                 ...
             gevent.spawn(do_some_work)
             return 'Regular response'
 
     .. versionadded:: 0.10
+
+    .. versionchanged:: 1.0.3
+       A copy of the current session object is added to the request context
+       copy. This prevents `flask.session` pointing to an out-of-date object.
     """
     top = _request_ctx_stack.top
     if top is None:
--- a/flask/ctx.py
+++ b/flask/ctx.py
@@ -14,7 +14,7 @@ from functools import update_wrapper
 
 from werkzeug.exceptions import HTTPException
 
-from .globals import _request_ctx_stack, _app_ctx_stack
+from .globals import session, _request_ctx_stack, _app_ctx_stack
 from .signals import appcontext_pushed, appcontext_popped
 from ._compat import BROKEN_PYPY_CTXMGR_EXIT, reraise
 
@@ -147,6 +147,7 @@ def copy_current_request_context(f):
             'when a request context is on the stack.  For instance within '
             'view functions.')
     reqctx = top.copy()
+    reqctx.session = session.copy()
     def wrapper(*args, **kwargs):
         with reqctx:
             return f(*args, **kwargs)
--- a/tests/test_reqctx.py
+++ b/tests/test_reqctx.py
@@ -156,7 +156,9 @@ class TestGreenletContextCopying(object):
 
         @app.route('/')
         def index():
+            flask.session['fizz'] = 'buzz'
             reqctx = flask._request_ctx_stack.top.copy()
+            reqctx.session = flask.session.copy()
 
             def g():
                 assert not flask.request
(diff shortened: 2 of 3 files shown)


ThiefMaster:
Looks good, but please shorten your commit messages a bit. General recommendation is max 51 chars in the first line, and yours are even beyond the limit where GitHub truncates them.

Also, I wonder if we should do a deep copy on the session... otherwise people might get a bad surprise in case of mutable objects stored there.

dchevell:
Fixed the commit messages @ThiefMaster. 

Re: deep copying, you're right - worth noting that the same issue is already true for `flak.request` context which is also susceptible into the same issue of people modifying mutable objects from the copy. All the same I'm happy to make that change

dchevell:
@ThiefMaster is there a chance users would store stuff in the session object that would fail a deepcopy? I don't want to break people's stuff. Perhaps a more conservative approach would be:

```python
try:
    reqctx.session = deepcopy(session)
except TypeError:
    reqctx.session = session.copy()
```

ThiefMaster:
https://github.com/pallets/flask/blob/master/flask/json/tag.py

By default only the object types handled in there can be stored in a session

dchevell:
Ah, no worries then. Alright, switched to deepcopy.

dchevell:
Anything further needed to review this @ThiefMaster ?

tests/test_reqctx.py
```diff
@@ -156,7 +158,9 @@ def test_greenlet_context_copying(self, app, client):
 
         @app.route('/')
         def index():
+            flask.session['fizz'] = 'buzz'
             reqctx = flask._request_ctx_stack.top.copy()
+            reqctx.session = deepcopy(flask.session)
```
ThiefMaster: i don't think this should be in the test.

dchevell: Isn't the purpose of `test_greenlet_context_copying` to test the raw context copying functionality, separately from `test_greenlet_context_copying_api`? Without that line, the test fails. I can remove the session test from that altogether, but in that case I'm not sure what `test_greenlet_context_copying` is testing for.

ThiefMaster: ah I didn't realize this is some greenlet stuff. nevermind then i guess.

but in that case we're missing a testcase for your feature in a normal wsgi context

dchevell: Hmm, in that case shouldn't there already be a test for `copy_current_request_context` outside of the greenlet testing? This is just a small modification to that existing feature.

ThiefMaster: yes... probably. i hoped we'd have such a test. :/

dchevell: I can write one but it seems a bit outside the scope of this change in that case - I'm not actually certain what that test should look like, because the feature doesn't really do anything useful when used directly inline without a greenlet/thread. I can write one anyway but that might turn this PR into something else :)

ThiefMaster: agreed, let's keep this pr as it is

dchevell:
@davidism do you mean along the lines of, `@copy_current_request_context(deepcopy_session=True)` ? 

Do you think there are cases to want both behaviours? deepcopy seems safer, since the intent of the method isn't to provide references back to the original objects. I can imagine a case where a user would prefer deepcopy, but not one where they'd prefer a shallow copy.

ThiefMaster:
Maybe an attribute on the Flask class or the session interface (pointing to the copy function to use)? I'd rather subclass once instead of having to specify this again and again.

dchevell:
I can do that. First, though, what's the benefit here? What does offering shallow vs deep copy achieve?  Adding API surface increases cost (testability, complexity, documentation, etc.) and given the purpose of this helper function is to create a copy for use in threads or greenlets I'm not sure what shallow copies buy us. Happy to be overruled but I'd like to make sure before I touch any other code.

dchevell:
Just checking in with @davidism to see if there's value / benefit in adding new API's to allow multiple copy behaviours - I'm still erring on the side of not, since this is already pretty edge-casey and it seems like an edge case of an edge case. Still, open to other opinions.

okeuday:
A separate problem is that `flask.copy_current_request_context` is calling `RequestContext` `copy` which does not copy the contents of `environ`.  By not having a copy of `environ` when using `copy_current_request_context` with threads, it is possible for an old request to get data from a new request (in flask 1.0.2) (e.g., `request.method`).  Such behavior seems to be justified by the code comment of the `RequestContext` `copy` function when it says "Because the actual request object is the same this cannot be used to move a request context to a different thread unless access to the request object is locked.".  This indicates that `flask.copy_current_request_context` was previously meant to do a shallow copy and any deep copy functionality should require a boolean flag argument or a separate function.

I would like to use `flask.copy_current_request_context` if it was copying the `request` `environ`, so I would suggest adding a `deep` boolean keyword argument (defaulting to `False`) to the `RequestContext` `copy` function which is utilized when a boolean flag is passed to `flask.copy_current_request_context`.  However, I know that is separate from @dchevell 's pursuit of a session copy.

dchevell:
@okeuday To confirm, that's talking about deepcopying the *request*, not the *session*, right? (Of course, it would make sense to go all or nothing)

*Can* you deepcopy a request? I mean, you can't do it directly - you'll get `TypeError: can't pickle _thread.lock objects`. I'm sure this is solvable but unlike the session object it's not so trivial. 

What are your thoughts on reverting this change back to the original shallow copy implementation, then opening a separate issue for a true deepcopy for both `flask.request` and `flask.session` as you and @davidism suggested? It seems like a worthy pursuit, but we're moving beyond the scope of this PR and the original issue.

okeuday:
@dchevell I would agree that a deep copy is best that handles the `flask.request` and `flask.session`.  I am not sure if it needs a separate PR because they both seem related and it seems like people would likely want both to get a deep copy.  `flask.copy_current_request_context` seems like a name that covers both a shallow copy and a deep copy, so it seems like it needs a function argument to switch between the different behavior.  For the request copy, when not modifying `RequestContext` `copy`, I found it necessary to modify `flask.copy_current_request_context` so it looks like:

    def _copy_current_request_context(function):
        top = _request_ctx_stack.top
        if top is None:
            raise RuntimeError('This decorator can only be used at local scopes '
                               'when a request context is on the stack.  '
                               'For instance within view functions.')
        reqctx = top.copy()
        # environ is used by headers too, so only a copy of environ is necessary
        reqctx.request.environ = reqctx.request.environ.copy()
        def wrapper(*args, **kwargs):
            with reqctx:
                return function(*args, **kwargs)
        return update_wrapper(wrapper, function)

It should be better to modify `RequestContext` `copy` to do the `environ` copy if a `deep` keyword argument is passed.

dchevell:
@okeuday not sure I agree - the scope of this PR was simply, copies of the current request contain the previous session object. Shallow copy / deep copy covers the same area of code, but it's an entirely separate set of concerns. Other than touching the same code, I don't see a relationship between them. 

I agree that a deepcopy has value, just not that it fits within the scope of #2935

okeuday:
@dchevell Ok, np, I moved the deep copy of the current request information to https://github.com/pallets/flask/issues/2960

dchevell:
If #2960 covers the need for offering an API for both shallow and deepcopying, I think we should revert the deepcopy change here back to the original shallow copy implementation, then a fix for #2960 should implement a deepcopy option which implements that behaviour for the request *and* session objects

dchevell:
Any more thoughts on this? Given #2960 I would like to revert this to a shallow copy implementation, similar to the existing functionality for how the request object is copied, and then I'd be happy to follow up #2960 to investigate deep copy implementations for both.

/cc @davidism @ThiefMaster

flask/ctx.py
```diff
@@ -133,20 +135,25 @@ def copy_current_request_context(f):
...
+       A copy of the current session object is added to the request context
+       copy. This prevents `flask.session` pointing to an out-of-date object.
     """
     top = _request_ctx_stack.top
     if top is None:
         raise RuntimeError('This decorator can only be used at local scopes '
             'when a request context is on the stack.  For instance within '
             'view functions.')
     reqctx = top.copy()
+    reqctx.session = deepcopy(session)
```
davidism: This should be added to `RequestContext.copy()` instead, along with the docs changes.

flask/ctx.py
```diff
@@ -133,20 +135,25 @@ def copy_current_request_context(f):
...
-                # would otherwise in the view function.
+                # do some work here, it can access flask.request or
+                # flask.session like you would otherwise in the view function.
                 ...
             gevent.spawn(do_some_work)
             return 'Regular response'
 
     .. versionadded:: 0.10
+
+    .. versionchanged:: 1.0.3
```
davidism: This feels like a 1.1 change instead of a bugfix change. And since you branched off master, that's what would happen anyway.

davidism:
Looking at the docs for `RequestContext.copy()`:

> Creates a copy of this request context with the same request object. This can be used to move a request context to a different greenlet. Because the actual request object is the same this cannot be used to move a request context to a different thread unless access to the request object is locked.

It's clear copying was never meant to use deepcopy, it requires the user to lock around multiple threads modifying. So let's go back to `copy`.

dchevell:
Thanks for the pointers @davidism  - would you mind checking the latest diff to confirm if it matches what you were thinking?

flask/ctx.py
```diff
@@ -321,10 +322,15 @@ def copy(self):
         request object is locked.
 
         .. versionadded:: 0.10
+
+        .. versionchanged:: 1.1
+           A copy of the current session object is added to the request context
```
davidism: It's not a copy, it's the same object, so this is a bit misleading. Should be "The current session object is used instead of reloading the original cookie from the copied request." Same with the other places you used "copy".

dchevell: Fair point @davidism - fixed the docstrings and the changelog

PR #2942, opened 2018-10-15
Allow positional arguments to @login_required decorator in tutorial

hoheinzollern:
Updating posts was not working due to a positional argument not playing well with the @login_required function (using Python 3.6).

I changed the function to accept positional arugments, which fixes the problem.


ThiefMaster:
Can I see a traceback of it failing withour your change? Flask view functions only use kwargs but no posargs.

hoheinzollern:
Ok, that was my bad, sorry. I added @login_required to the get_post function to do some testing, which had a positional argument. If I keep that decoration then I can show a trace.
Not sure whether that makes sense to keep or not.

ThiefMaster:
I think the decorator only makes sense on view functions.

PR #2944, opened 2018-10-15
Docs: add warning about custom commands with Click v7.x turning underscores to dashes

zoltan-fedor:
Starting Click v7.0 subcommands that are named by the function automatically have underscores replaced with dash (see the 4th item from the bottom for the changelog of Click v7.0 at http://click.palletsprojects.com/en/7.x/changelog/).

This caused the Custom Commands example in the documentation not working with Click v7.x, so I have added a warning for the Click v7.x users.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -270,6 +270,16 @@ This example adds the command ``create_user`` that takes the argument
 
     $ flask create_user admin
 
+.. warning:: `Click`_ v7.0 has introduced a change (see `v7.0 changelogs`_):
+    "Subcommands that are named by the function now automatically have the
+    underscore replaced with a dash. If you register a function named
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, but I'd rather get a release of Click that addresses that well before these Flask docs changes would ever be published.

zoltan-fedor:
I thought it is better to put in a warning instead of just changing the example to `flask create-user admin` , as I assume many are still using pre-v7.0 version of Click, where the underscores don't get replaced with dashes, so if the docs are updated simply stating that the command is now with dashes, then that wouldn't work for anybody still using Click 6.x (which I assume is still substantial number of users).

But if you prefer, I can make a PR just to having `flask create-user admin` instead of `flask create_user admin` without the warning for the Click version.

Thanks

PR #2953, opened 2018-10-19
Fix #2952: Pin itsdangerous to below v1

### LINKED ISSUE #2952: New itsdangerous version invalidates cookie based sessions
### Expected Behaviour
User sessions should remain valid across versions.

### Actual Behaviour

itsdangerous v1.0 was release about 16 hours ago.
Users are signed out when an build of flask that has depended on the new itsdangerous v1.0 is released. This is because the new version has a breaking change that changes the session hashing algorithm.

https://github.com/pallets/itsdangerous/blob/master/CHANGES.rst
https://github.com/pallets/flask/blob/master/flask/sessions.py#L6 
https://github.com/pallets/flask/blob/master/setup.py#L39
https://pypi.org/project/itsdangerous/#history

### Environment

* Python version: 3.6
* Flask version: 1+

benvand:
itsdangerous v1 has a breaking change that changes the session hashing algorithm.

Users would be signed out when an build of flask that has depended on the new itsdangerous v1.0 is released.

https://github.com/pallets/flask/issues/2952

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
     install_requires=[
         'Werkzeug>=0.14',
         'Jinja2>=2.10',
-        'itsdangerous>=0.24',
+        'itsdangerous>=0.24,<1',
         'click>=5.1',
     ],
     extras_require={


ThiefMaster:
Updating all apps to the latest itsdangerous should help.

Anyway, while I still don't think we need to do a new Flask release, I do think we should pin *all* our dependencies to the current major versions to require an explicit Flask update to get a major version change in a dependency.

davidism:
Closing, see linked issue for response.

davidism:
@ThiefMaster we can discuss version pinning policies in a new issue if you'd like.

benvand:
@ThiefMaster Totally agree about pinning deps
We have a complex environment of multiple apps and allowing flasks dependencies to have breaking changes damages the interoperability of those apps.

PR #2963, opened 2018-10-24
Fix #2961：ignore colon followed by slash when split app_import_path

### LINKED ISSUE #2961: Flask doens't run when APP is specified with full path
### Expected Behavior
Previously in our development environment we ran flask app from vscode but after we updated to the latest version of flask we started experiencing problems. After some investigation I found out that in our run configuration we specify the whole path to the app.py-file. Changing to just referencing it directly (i.e. app.py instead of C:\Path\To\My\app.py) I got it to work.

I recreated the fault in another simpler environment:

The following command runs 
``` env "FLASK_APP=`pwd`/app.py" "FLASK_ENV=Development" "FLASK_DEBUG=1" env/Scripts/python.exe -m flask run```

but when you try to access localhost:5000/ you get the following error

```Traceback (most recent call last):
  File "C:\Users\AlexanderS\code\flask-bugg\env\lib\site-packages\flask\cli.py", line 330, in __call__
    rv = self._load_unlocked()
  File "C:\Users\AlexanderS\code\flask-bugg\env\lib\site-packages\flask\cli.py", line 317, in _load_unlocked
    self._app = rv = self.loader()
  File "C:\Users\AlexanderS\code\flask-bugg\env\lib\site-packages\flask\cli.py", line 372, in load_app
    app = locate_app(self, import_name, name)
  File "C:\Users\AlexanderS\code\flask-bugg\env\lib\site-packages\flask\cli.py", line 246, in locate_app
    'Could not import "{name}".'.format(name=module_name)
flask.cli.NoAppException: Could not import "C".
```

### Actual Behavior
The application should run without errors.

### Environment

* Python version: 3.6.5
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

Recreated the bug with the follow app.py

```python
from flask import Flask

app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello world2'
```

garenchan:
Flask currently supports importing app through a combination of module path and app variable name, such as '/usr/app.py:my_app'. When the module path contains a colon, it will conflict with this import way and a `flask.cli.NoAppException` will be raised.

A file path on a Windows system may contain a colon followed by a slash. So we solved this problem on Windows by ignoring the colon followed by a slash when we split `app_import_path`. 

Fix issue #2961.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -369,7 +369,7 @@ class ScriptInfo(object):
             app = call_factory(self, self.create_app)
         else:
             if self.app_import_path:
-                path, name = (self.app_import_path.split(':', 1) + [None])[:2]
+                path, name = (re.split(r':(?![\\/])', self.app_import_path, 1) + [None])[:2]
                 import_name = prepare_import(path)
                 app = locate_app(self, import_name, name)
             else:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -261,8 +261,21 @@ def test_get_version(test_apps, capsys):
 def test_scriptinfo(test_apps, monkeypatch):
     """Test of ScriptInfo."""
     obj = ScriptInfo(app_import_path="cliapp.app:testapp")
-    assert obj.load_app().name == "testapp"
-    assert obj.load_app().name == "testapp"
+    app = obj.load_app()
+    assert app.name == "testapp"
+    assert obj.load_app() is app
+
+    # import app with module's absolute path
+    cli_app_path = os.path.abspath(os.path.join(
+            os.path.dirname(__file__), 'test_apps', 'cliapp', 'app.py'))
+    obj = ScriptInfo(app_import_path=cli_app_path)
+    app = obj.load_app()
+    assert app.name == 'testapp'
+    assert obj.load_app() is app
+    obj = ScriptInfo(app_import_path=cli_app_path + ':testapp')
+    app = obj.load_app()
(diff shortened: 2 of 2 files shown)


flask/cli.py
```diff
@@ -369,7 +369,7 @@ def load_app(self):
             app = call_factory(self, self.create_app)
         else:
             if self.app_import_path:
-                path, name = (self.app_import_path.split(':', 1) + [None])[:2]
+                path, name = (re.split(r':(?![\\\/])', self.app_import_path) + [None])[:2]
```
ThiefMaster: I'm pretty sure you do not need to escape forward slashes

garenchan:
CI build failed because of an issue that has been solved on the master branch.
image

flask/cli.py
garenchan: Thank you!That's exactly what you said.

flask/cli.py
```diff
@@ -369,7 +369,7 @@ def load_app(self):
             app = call_factory(self, self.create_app)
         else:
             if self.app_import_path:
-                path, name = (self.app_import_path.split(':', 1) + [None])[:2]
+                path, name = (re.split(r':(?![\\/])', self.app_import_path) + [None])[:2]
```
davidism: Should still only split once, `re.split(pattern, value, 1)`.

garenchan: Thank you! I forgot about it. I have revised it according to what you said.

Issue #2967, opened 2018-10-26
Documentation: Explicitly stating where to run `flask run` command for Tutorial


stepharr:
Would it be beneficial to others following the tutorial to explicitly say that you should run the `flask run` command in the parent directory of "flaskr" and not in  "flaskr"? I stumbled on this getting an error when trying to run the flask developer server from the flaskr directory while following along in the [tutorial](http://flask.pocoo.org/docs/1.0/tutorial/factory/#the-application-factory)
https://github.com/pallets/flask/blob/master/docs/tutorial/factory.rst

PR #2971, opened 2018-10-27
Tutorial: Project Layout - clarify flask run

stepharr:
Clarify documentation as discussed here: https://github.com/pallets/flask/issues/2967


davidism:
Thanks, going with an alternate wording instead.

Issue #2981, opened 2018-10-31
Patterns Method Overrides


Bernardoow:
I tried use [Method override pattner](http://flask.pocoo.org/docs/1.0/patterns/methodoverrides/) and it was crashed when method.encode('ascii', 'replace') because the next method needs string value and not byte value.

Bernardoow:
This errors just happend in MethodView. In function works fine.

Trace:
`  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/views.py", line 88, in view
    return self.dispatch_request(*args, **kwargs)
  File "/home/bernardo/.virtualenvs/flaskmethodoverrides-NLvu9JR3/lib/python3.6/site-packages/flask/views.py", line 150, in dispatch_request
    meth = getattr(self, request.method.lower(), None)
TypeError: getattr(): attribute name must be string
`

davidism:
This sounds like an issue with `MethodView`, not with that example.

PR #2982, opened 2018-10-31
Added one tip to use this pattern in Python3.

Bernardoow:
I just added one warning to people that is using python3.

#2981

### DIFF
--- a/docs/patterns/methodoverrides.rst
+++ b/docs/patterns/methodoverrides.rst
@@ -29,7 +29,7 @@ This can easily be accomplished with an HTTP middleware::
         def __call__(self, environ, start_response):
             method = environ.get('HTTP_X_HTTP_METHOD_OVERRIDE', '').upper()
             if method in self.allowed_methods:
-                method = method.encode('ascii', 'replace')
+                method = method.encode('ascii', 'replace')# Remove this line if you are using Python3. 
                 environ['REQUEST_METHOD'] = method
             if method in self.bodyless_methods:
                 environ['CONTENT_LENGTH'] = '0'


docs/patterns/methodoverrides.rst
```diff
@@ -29,7 +29,7 @@ This can easily be accomplished with an HTTP middleware::
         def __call__(self, environ, start_response):
             method = environ.get('HTTP_X_HTTP_METHOD_OVERRIDE', '').upper()
             if method in self.allowed_methods:
-                method = method.encode('ascii', 'replace')
+                method = method.encode('ascii', 'replace')# Remove this line if you are using Python3.
```
grahamlyons: Could you add an explanation of why you recommend removing that line so that people can get a better understanding?

Bernardoow: I tried use Method override pattner and it was crashed when method.encode('ascii', 'replace') because the next method needs string value and not byte value.

Bernardoow: with this line, the app crashs.

Bernardoow: @grahamlyons Please execute test in python3 that enter in this conditional.

[Python Method encode 3](https://docs.python.org/3/howto/unicode.html#converting-to-bytes) -> convert to Bytes.

[Python Method encode 2](https://docs.python.org/2/howto/unicode.html#the-unicode-type)  -> convert to String.

garenchan: I tested this example in Python 3 and  it worked fine.
Normally, the variable `method` should be `str` type. I don't know how you made it into `bytes`. So can you give more details?

garenchan: Now the description of this problem is very clear. I think you'd better post the stack traceback of the error in the issue #2981. This will help you describe the problem more accurately.

Just as an example, I think you can remove this line of code directly, and it will work fine under python 2 as well.  As to which solution is better, you can consult @davidism!

Bernardoow: @garenchan I apologize for my mistakes.

garenchan: That's alright.

davidism:
As far as I can tell that `.encode` call isn't necessary on Python 2 or Python 3. I'll just remove it.

davidism:
#3068

Issue #2984, opened 2018-11-01
HTTPException handler catches non-error exceptions like RequestRedirect


taion:
This is related to https://github.com/pallets/flask/issues/2841, but the emphasis is different.

### Expected Behavior

Non-error exceptions should not be caught by registered error handlers.

```python
app.register_error_handler(HTTPException, my_error_handler)

@app.route('/path-with-slash/')
def my_route():
    return '', 204
```

Attempting to visit `/path-with-slash` should redirect to `/path-with-slash/`, without invoking `my_error_handler`.

### Actual Behavior

As the `RequestRedirect` raised internally in Werkzeug is a subclass of `HTTPException`, `my_error_handler` will get called. This is unexpected because no error has occurred, and the only appropriate way to handle this case is to add boilerplate to re-raise any Werkzeug `RoutingException`s.

### Environment

* Python version: 3.7.0
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

taion:
The most straightforward fix would be to add a hard-coded exception to not handle `RoutingException` instances with the error handling logic in Flask. `RequestRedirect` instances are not errors.

davidism:
I'm not clear how this is different than #2841.

taion:
The difference is that the `NotFound` coming from routing is (1) actually an error and (2) is indistinguishable from a `NotFound` raised within the app. Additionally, a generic error handler that preserves the error code will still work.

None of those are true for a `RequestRedirect` raised internally by Werkzeug's routing logic. It's (1) not an error and (2) trivially distinguishable from other errors, as it extends `RoutingException`. Additionally, error handlers for generic `HTTPException`s will do something wrong here, as it's insufficient to just preserve `error.code` – one must preserve the `Location` header from the baked-in response as well, which requires reaching more deeply into the internals of the exception class.

taion:
The issue here is that no error handler should catch instances of `RequestRedirect`, because they're not errors. The use of `RequestRedirect` is an internal implementation detail of Werkzeug. `RequestRedirect` in fact does not even show up in `exceptions.default_exceptions` in Werkzeug.

This is different from e.g. `NotFound`, in that `NotFound` is an HTTP error class, and a default exception class.

ThiefMaster:
could we add an option either when registering error handlers or in `app.config` on whether HTTPException or even Exception error handlers catch http exceptions?

taion:
These are the differences:

[...]

PR #2986, opened 2018-11-01
Do not handle RoutingExceptions with app error handlers

taion:
Fixes https://github.com/pallets/flask/issues/2984

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,11 +19,16 @@ Unreleased
     Python version. (`#2825`_)
 -   :func:`send_file` handles an ``attachment_filename`` that is a
     native Python 2 string (bytes) with UTF-8 coded bytes. (`#2933`_)
+-   A catch-all error handler registered for ``HTTPException`` will not
+    handle ``RoutingExcpetion``, which is used internally during
+    routing. This fixes the unexpected behavior that had been introduced
+    in 1.0. (`#2986`_)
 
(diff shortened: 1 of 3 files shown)


davidism:
Yeah, this makes sense now.

taion:
Any thoughts here?

davidism:
I did a little more research just to make sure I understood what changed from 0.12 to 1.0, in case we need to address this again later.

Prior to 1.0, error handlers for `HTTPException` classes were *only* looked up by code. So if there wasn't a specific handler registered, it wouldn't fall back to a more general handler, which people found confusing. But it *was* possible to add a handler specifically for the `301` code and have different redirect behavior. Why you would do that, I don't know, but it was possible.

I'm OK with saying that anyone who really wants to catch `RequestRedirect` should override `handle_http_exception` or `url_map` to do something different. Or if they want to redirect using exceptions, they can create their own `class Found(HTTPException): code = 302` and add a handler for that.

taion:
> Or if they want to redirect using exceptions, they can create their own `class Found(HTTPException): code = 302` and add a handler for that.

One small comment on the above – if someone does want to redirect using exceptions in a view, they wouldn't use `RequestRedirect` anyway. It's documented as a thing to use inside Werkzeug routing.

Because `RoutingException` is not actually part of `werkzeug.exceptions.default_exceptions`, installing an exception handler for 302 (well, now it's 308) in 0.12 would not have caught `RoutingException`s anyway.

PR #2991, opened 2018-11-07
Fixes for sentry docs

untitaker:
Followup to #2990

### DIFF
--- a/docs/errorhandling.rst
+++ b/docs/errorhandling.rst
@@ -57,19 +57,15 @@ The `YOUR_DSN_HERE` value needs to be replaced with the DSN value you get
 from your Sentry installation.
 
 After installation, failures leading to an Internal Server Error are automatically reported to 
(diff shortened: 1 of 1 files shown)


untitaker:
cc @atwalsh

docs/errorhandling.rst
```diff
@@ -57,19 +57,15 @@ The `YOUR_DSN_HERE` value needs to be replaced with the DSN value you get
...
 
 After installation, failures leading to an Internal Server Error are automatically reported to 
-Sentry and from there you can receive error notifications. Sentry also supports 
-capturing custom exceptions::
-    
-    import sentry_sdk
-    
-    try:
-        throwing_function()
-    except Exception as e:
```
untitaker: I removed this because it's not necessary for basic getting started. Also `push_scope` is not necessary, there is already a scope by default.

PR #2993, opened 2018-11-12
Update documentation for users to use flask-caching instead of Werkzeug cache

data-get:
Update the documentation to point users to use flask-caching instead of the deprecated Werkzeug cache. I have linked to [flask-caching's documentation](https://flask-caching.readthedocs.io/en/latest/) instead of copying flask-caching example code.


davidism:
@sh4nks @buckley-w-david

PR #2997, opened 2018-11-16
Update app.py

Blacker1230:
fix a bug when judge if methos are string

the condition may be wrong：
"GET" isinstance of (str,)


Blacker1230:
I made a mistake

PR #3001, opened 2018-11-19
1.0 maintenance

jefmud:
as per @davidism, I am submitting a PR against 1.0 maintenance

**Tiny correction to the documentation on GEvent as a WSGI standalone container.**

http://flask.pocoo.org/docs/1.0/deploying/wsgi-standalone/#gevent

```python
from gevent.wsgi import WSGIServer # incorrect
from gevent.pywsgi import WSGIServer # correct to pywsgi
from yourapplication import app

http_server = WSGIServer(('', 5000), app)
http_server.serve_forever()
```

Thanks for all the hard work!
Jeff

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,16 +1,16 @@
 environment:
   global:
-    TOXENV: py
+    TOXENV: py,codecov
 
   matrix:
-    - PYTHON: C:\Python36
-    - PYTHON: C:\Python27
+    - PYTHON: C:\Python36-x64
+    - PYTHON: C:\Python27-x64
 
 init:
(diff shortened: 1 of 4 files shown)


davidism:
I'm not sure what happened here, but you need to make a branch off 1.0-maintenance, commit to it, then make a PR against our 1.0-maintenance branch. It looks like you somehow made GitHub think it was merging our 1.0-maintenance into master, but there's also no commits from you.

jefmud:
it is correct in the repository, but not on the (current) public facing website.  Sorry, I wasn't clear in the previous message. *this issue is closed*

PR #3004, opened 2018-11-23
Document 'flask run --host' and '--port'

andybalaam:
Small additions to `flask --help` and the docs to make more mention of the `--help`, `--port` and `--host` arguments to `flask run`.

Also adds a mention of the environment variable `FLASK_RUN_HOST` since that works, but was not mentioned anywhere I could find.

I don't know whether this merits a changelog entry - do let me know if that is needed.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -101,6 +101,8 @@ replaces the :meth:`Flask.run` method in most cases. ::
     is provided for convenience, but is not designed to be particularly secure,
     stable, or efficient. See :ref:`deployment` for how to run in production.
 
+To run the development server on a specific interface or port, use the
+``--host`` or ``--port`` arguments.  Run ``flask run --help`` for details.
 
 Open a Shell
 ------------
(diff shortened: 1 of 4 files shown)


docs/cli.rst
```diff
@@ -101,6 +101,8 @@ replaces the :meth:`Flask.run` method in most cases. ::
     is provided for convenience, but is not designed to be particularly secure,
     stable, or efficient. See :ref:`deployment` for how to run in production.
 
+To run the development server on a specific interface or port, use the
+``--host`` or ``--port`` arguments.  Run ``flask run --help`` for details.
```
davidism: Earlier in the doc: "The `--help` option will give more information about any commands and options."

I specifically don't call out `--host` because in most cases, including your added example, people misuse this to make the dev server visible publicly, which is explicitly discouraged. Changing the port is already described later.

docs/cli.rst
```diff
@@ -190,14 +192,15 @@ Setting Command Options
 
 Click is configured to load default values for command options from
 environment variables. The variables use the pattern
-``FLASK_COMMAND_OPTION``. For example, to set the port for the run
-command, instead of ``flask run --port 8000``:
+``FLASK_COMMAND_OPTION``. For example, to set the port and host for the run
+command, instead of ``flask run --port 8000 --host 0.0.0.0``:
```
davidism: I do not want to encourage users to set the host to `0.0.0.0`.

flask/cli.py
```diff
@@ -725,7 +725,9 @@ def _validate_key(ctx, param, value):
     return value
 
 
-@click.command('run', short_help='Runs a development server.')
+@click.command('run',
+               short_help='Runs a development server (flask run --help for '
```
davidism: It doesn't make sense to add this here but not to any other commands. We're following the general pattern that `--help` is commonly understood to exist for command line tools, and the usage message also calls this out on bad args.

davidism:
Thanks for looking at this. For the reasons stated, I don't think I'm going to merge this.

andybalaam:
Thanks @davidism , fair enough.  I found it hard to find the `--host` and `--port` arguments, and I didn't think to try `flask run --help` even though I am reasonably familiar with command line tools.  (I did run `flask --help`.)  It was quite easy to find `app.run(host='blah')` (despite the warning that you should not use the development server in production), so I was trying to put these command line options on a similar level of visibility.  But I understand your comments.  Since I use `git <subcommand> --help` quite often maybe I should have thought of `flask run --help` so maybe it's fine :-)

godgood22:
سلام 
من هنوز 
بلد نیستم
چکار کنم

PR #3015, opened 2018-12-05
Lighttpd configuration: deploy script link

tallero:
I have moved the module requirements up and added a note about a terminal script that permits to create an example deployment on the fly.


davidism:
Thanks for the interest, but including a random, unexplained, external script is not appropriate for this page, and the other changes are just moving things around for no reason.

tallero:
I could try making it simpler or include something different.

Simply I have lost lots of time trying to configure lighttpd to work with flask.

I think it would have been awesome to link a program like this one, that automatically configure lighttpd to work with a newly created example app, here or in the quickstart.

PR #3021, opened 2018-12-08
add classifiers

CasterWx:
add classifiers ：Programming Language


garenchan:
Currently, there is no test for `Python 3.7` in CI system, so it is not appropriate to add `Python 3.7` here.
Unless we can guarantee that all tests will pass in `Python 3.7`.

davidism:
We'll add the tests, and I know it works fine, so I'll add it now.

PR #3030, opened 2018-12-14
[Docs] Add simple analogy to app factories doc

umluizlima:
Added an analogy that illustrates the app factory pattern as a sandwich parlor. It might help understanding the concepts.


davidism:
Thanks, but I don't think this should be added. It doesn't really work as an analogy, it excludes the key benefits of using a factory.

PR #3038, opened 2018-12-20
Docstring changed for flask app route

anandtripathi5:
Added methods usage in route decorator


davidism:
Thanks, but there are lots of other things that can be passed to the decorator, and they're all described in the linked docs.

PR #3040, opened 2018-12-21
docs: Update virtualenv creation process for Fedora

rht:
The other alternative is to tell to install `virtualenv` system package, then run `virtualenv -p python2.7 venv`, but this is putting exception in 2 places (1. installation and 2. venv creation).

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -98,6 +98,14 @@ Python, use the following command instead:
 
     virtualenv venv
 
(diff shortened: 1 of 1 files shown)


davidism:
The top of the section already says to use `python3 -m venv venv`. Python 2 does not have `-m venv`, and the parenthetical seems like too much explanation and is also really confusing.

PR #3047, opened 2018-12-26
 Force endpoint name to string in Python 2

### LINKED ISSUE #3033: TypeError: __name__ must be set to a string object
### Expected Behavior

I wanna write code capable of python3. So i add "from future import unicode_literals" on the top of file.Meanwhile, i use MethodView to build rest api in the same file. I wish it could work fine.

Here is an example:

```
  # -*- coding: utf-8 -*-

  from __future__ import unicode_literals
  from flask import Flask
  from flask import Blueprint, request, current_app, g, abort

  masterviews = Blueprint('newmasterviews', __name__)


  class TeamAPI(MethodView):

      def get(self):
          pass


  team_view = TeamAPI.as_view('team_api')

  app = Flask(__name__)

  app.register_blueprint(masterviews, url_prefix='/rest/master')
```

### Actual Behavior

However, when i run the web app, it thows the error,

```
"lib/python2.7/site-packages/flask/views.py", line 98, in as_view
view.__name__ = name
TypeError: __name__ must be set to a string object". 
```
I soon realize that because i use the "from future import unicode_literals" in py2 so the function name is unicode type, but the name of a function should be string type.So i wonder is it possible to change the
source code in the views.py and add the if-then workflow to convert name of function from unicode type into string type to handle the situation .

### Environment
python2.7
Flask==1.0.2
Werkzeug==0.14.1

greyli:
Using unicode endpoint name in Python 2 will raise `TypeError: __name__ must be set to a string object`.

Example application:
```py
from __future__ import unicode_literals

from flask import Flask

app = Flask(__name__)

class Index(MethodView):
    def get(self):
        return 'GET'

app.add_url_rule('/', view_func=Index.as_view('index'))
```
Full error traceback:
```pytb
  File "...\app.py", line 93, in <module>
    app.add_url_rule('/', view_func=Index.as_view('index'))
  File "python27\lib\site-packages\flask\views.py", line 105, in as_view
    view.__name__ = name
TypeError: __name__ must be set to a string object
```

This PR will force endpoint name to string in Python 2.

Close #3033

### DIFF
--- a/flask/views.py
+++ b/flask/views.py
@@ -10,7 +10,7 @@
 """
 
 from .globals import request
-from ._compat import with_metaclass
+from ._compat import with_metaclass, PY2, text_type
 
 
 http_method_funcs = frozenset(['get', 'post', 'head', 'options',
@@ -83,6 +83,9 @@ class View(object):
         The arguments passed to :meth:`as_view` are forwarded to the
         constructor of the class.
         """
+        if PY2 and isinstance(name, text_type):
+            name = str(name)
+
         def view(*args, **kwargs):
             self = view.view_class(*class_args, **class_kwargs)
             return self.dispatch_request(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


davidism:
See linked issue for close reason.

PR #3052, opened 2018-12-28
Getter and setter right way for App Config

julianogouveia:
Reading the App Config class I found a noising way to make a getter and setter for a property in the Config class, so I refactor to what is the most acceptable and reasonable form.

I hope you guys enjoy the contribution.

Thks.


davidism:
There's nothing wrong with the current version, it's just an old pattern. This would be part of a larger style cleanup, but that's something the maintainers will do on their own.

PR #3054, opened 2018-12-29
Fix DeprecationWarning on collections import in flask.sessions

### LINKED ISSUE #3053: DeprecationWarning in flask.sessions
`flask.session` imports `MutableMapping` from collections which is deprecated in Python 3.7 and will be removed in 3.8. For some reason since there are similar warnings in other libraries like werkzeug and jinja that are fixed seems to hide this warning. Importing session directly from flask folder causes this warning

### Expected Behavior

`DeprecationWarning` should not be shown since it causes incompatibility with 3.8

### Actual Behavior

Tell us what happens instead.

```python
Python 3.8.0a0 (heads/master:78de01198b, Dec 29 2018, 11:36:21)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import warnings
>>> warnings.filterwarnings("default", category=DeprecationWarning)
>>> import sessions
/home/karthi/flask/flask/sessions.py:14: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import MutableMapping
```

### Environment

* Python version: Python 3.8.0a0
* Flask version: 1.1dev
* Werkzeug version: 0.14

Upstream PR : https://github.com/python/cpython/pull/10596

tirkarthi:
This fixes #3053 by moving the import to _compat . This can be directly imported since it's the only file but moving it to _compat makes it easier to use it in future for these cases and also goes in line with fixes in jinja2 and werkzeug for similar issues. Let me know if this needs to be changed. 

Thanks for Flask :)

### DIFF
--- a/flask/_compat.py
+++ b/flask/_compat.py
@@ -28,6 +28,7 @@ if not PY2:
 
     from inspect import getfullargspec as getargspec
     from io import StringIO
+    import collections.abc as collections_abc
 
     def reraise(tp, value, tb=None):
         if value.__traceback__ is not tb:
@@ -47,6 +48,7 @@ else:
 
     from inspect import getargspec
     from cStringIO import StringIO
+    import collections as collections_abc
 
     exec('def reraise(tp, value, tb=None):\n raise tp, value, tb')
 
--- a/flask/sessions.py
+++ b/flask/sessions.py
@@ -11,17 +11,17 @@
 
 import hashlib
 import warnings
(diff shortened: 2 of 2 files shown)


tirkarthi:
Build error in Travis seems to be unrelated to this PR.

PR #3060, opened 2019-01-04
Fix DeprecationWarning: invlid escape sequence in make-release.py

BoboTiG:
Hello,

In addition to #3056, this is a little patch to fix the last remaining `DeprecationWarning: invalid escape sequence` in `make-release.py`.

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


davidism:
Thanks, but this script isn't used anymore, I just haven't removed it.

PR #3069, opened 2019-01-08
Allow customizing url map class

shidenggui:
For speeding up url route matching, I defined a custom rule class and a custom url map class. But  `Flask` currently only allows modifying url rule class, it should support modify url map class too.

```python
    #: The rule object to use for URL rules created.  This is used by
    #: :meth:`add_url_rule`.  Defaults to :class:`werkzeug.routing.Rule`.
    #:
    #: .. versionadded:: 0.7
    url_rule_class = Rule
```


ThiefMaster:
FYI, your commit isn't properly linked to your GitHub account (different email).

Issue #2931, opened 2018-10-03
Click==7.0 incompatibles with Flask 1.0.2


uwevil:
I can not run Flask commands with Click 7.0 which is released the last week. When I replace it with the version 6.7, it works fine.

### Environment
* OS: Windows 10
* Python version: 3.6.4
* Flask version: 1.0.2
* Werkzeug version: 0.14

```
astroid==2.0.4
beautifulsoup4==4.6.0
click==6.7
cloudpickle==0.5.6
colorama==0.3.9
coverage==4.5.1
doit==0.31.1
Flask==1.0.2
gunicorn==19.9.0
isort==4.3.4
itsdangerous==0.24
Jinja2==2.10
lazy-object-proxy==1.3.1
MarkupSafe==1.0
mccabe==0.6.1
nose==1.3.7
nosexcover==1.0.11
pycryptodome==3.4.7
pylint==2.1.0
pymongo==3.7.1
PyYAML==3.12
six==1.11.0
typed-ast==1.1.0
urllib3==1.22
Werkzeug==0.14.1
wrapt==1.10.11
 

(env) flask clean_db
[2018-10-03 16:34:13,364 INFO] Running ...

```

If I install Click==7.0, my error is:
```

(env) flask clean_db
Usage: flask [OPTIONS] COMMAND [ARGS]...
Try "flask --help" for help.
 
Error: No such command "clean_db".
```

greyli:
Please provide the related code and the error traceback.

davidism:
Python 2.6 is not supported by any of our projects at the versions you've listed.

uwevil:
My bad, my python version is 3.6.4

uwevil:
Curiously, when I test in my mac, it works fine. I will provide my error on my Win 10 as soon as possible.

uwevil:
With my chance, I found the problem: if my command contains "_", Click (7.0) didn't work. But when I change my commande by removing "_", it works fine.

PR #3075, opened 2019-01-15
Updated CLI documentation to mention the issue with underscores and click 7

peterklijn:
When you use Flask in combination with Click 7.x, command containing underscores are no longer found, but instead exposed using dashes, so the example in the docs `create_user` only works as `create-user`.

I've extended the docs to mention this, including an example and a workaround.

The issue is mentioned here as well:
https://github.com/pallets/flask/issues/2931
https://github.com/pallets/click/issues/1123

This is the commit in Click that changes the behaviour:
https://github.com/pallets/click/commit/5d1df0e042b2f8b0dee8f6dd9ce74edce331a950

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -270,6 +270,25 @@ This example adds the command ``create_user`` that takes the argument
 
     $ flask create_user admin
 
+If you are using Click version 7 or higher, commands with underscores (`_`)
+will automatically be converted to dashes (`-`). The command above can be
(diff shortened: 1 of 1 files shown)


davidism:
I'd rather just use the Click 7 format and fix the issue in Click.

davidism:
Thanks for working on this. I updated the docs to use dashes instead in #3269.

PR #3104, opened 2019-02-24
moving match_request to RequestContext.push to enable ModelConverter

### LINKED ISSUE #3088: URL Converters should have access to app context
### Expected Behavior

The following code should create a new URL converter that converts a specific field to a model (or `None`).

```python
class ModelConverter(BaseConverter):

    def to_python(self, value):
        return models.Example.query.filter(models.Example.name == value).first()

    def to_url(self, value):
        if isinstance(value, str):
            return value
        return value.name
```

### Actual Behavior

The converters don't have access to the app context so an error is thrown.

### Bad Solution

```python
class OrganizationConverter(BaseConverter):
    app = None
    def to_python(self, value):
        with self.app.app_context():
            return repositories.Organization.query.filter(repositories.Organization.name == value).first()

    def to_url(self, value):
        if isinstance(value, str):
            return value
        return value.name
```

This solution tries to get around it by creating a new app context. The problem is that the context disappears, killing the database session at the same time. This prevents SQLAlchemy features (such as lazy loading).

### Environment

* Python version: 3.6 and 3.7
* Flask version: 1.0.2
* SQLAlchemy version: 1.2.17
* Flask-SQLAlchemy version: 2.3.2
* Werkzeug version: 0.14.1

### Additional Resources

This issue has also come up on [stack overflow](https://stackoverflow.com/questions/33856007/passing-application-context-to-custom-converter-using-the-application-factory-pa)

eladm26:
`match_request` will be called in `RequestContext.push` after `current_app`, `request`, and `session` are already populated.

closes #3088

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,6 +71,10 @@ Unreleased
 -   The ``flask run`` command no longer fails if Python is not built
     with SSL support. Using the ``--cert`` option will show an
     appropriate error message. :issue:`3211`
+-   URL matching now occurs after the request context is pushed, rather
+    than when it's created. This allows custom URL converters to access
+    the app and request contexts, such as to query a database for an id.
+    :issue:`3088`
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
 .. _#2957: https://github.com/pallets/flask/issues/2957
--- a/src/flask/ctx.py
+++ b/src/flask/ctx.py
@@ -347,8 +347,8 @@ class RequestContext(object):
         of the request.
         """
         try:
-            url_rule, self.request.view_args = self.url_adapter.match(return_rule=True)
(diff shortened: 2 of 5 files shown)


davidism:
I don't want to add a large dependency like SQLAlchemy just for this test. Can you rewrite this to use stuff from `request` in a converter rather than a db query?

eladm26:
Sure np

eladm26:
ok @davidism changed to the test

eladm26:
@davidism any update?

davidism:
When I have time to review, I will, no need to ping me.

eladm26:
Ok man, sorry for interrupting

davidism:
Azure still having issues, Tox suite passes locally.

PR #3109, opened 2019-03-02
Added closing li tags

chivalry:
This is a minor edit that adds the closing tags for the four `li` tags in the file.


davidism:
Many HTML tags do not require closing, this is the style we use. See other closed issues about this.

PR #3111, opened 2019-03-03
Allow dictionaries return values as JSON

pgjones:
This is something I've been experimenting with in Quart and I don't see a downside. I can't find any old issues relating to this (although I've found it hard to search for). This is roughly an issue, but as a pull request (branch) you can take the code and try it out. Obliviously if you consider this a good idea I'll add much more testing and documentation. 

This supports an increasingly common use-case whereby JSON is the primary response (rather than a templated string). Given Flask simplifies returning HTML responses, it seems fitting that it should also do so for JSON responses. In practice it allows,
```python
@app.route("/")
def index():
    return {
        "api_stuff": "values",
    }
```
which is equivalent to
```python
@app.route("/")
def index():
    return jsonify({
        "api_stuff": "values",
    })
```

### Note 

This doesn't support returning anything other than an associate array at the top level in the JSON response. I'm ok with this as in practice APIs are only extensible if the top level is an associate array.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,10 @@ Unreleased
 -   Add an ``--extra-files`` option to the ``flask run`` CLI command to
     specify extra files that will trigger the reloader on change.
     :issue:`2897`
+-   Allow returning a dictionary from a view function. Similar to how
+    returning a string will produce a ``text/html`` response, returning
+    a dict will call ``jsonify`` to produce a ``application/json``
+    response. :pr:`3111`
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
 .. _#2957: https://github.com/pallets/flask/issues/2957
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -683,7 +683,7 @@ The return value from a view function is automatically converted into
 a response object for you. If the return value is a string it's
 converted into a response object with the string as response body, a
 ``200 OK`` status code and a :mimetype:`text/html` mimetype. If the
-return value is a dict, ``jsonify`` is called to produce a response.
+return value is a dict, :func:`jsonify` is called to produce a response.
 The logic that Flask applies to converting return values into response
 objects is as follows:
 
@@ -719,6 +719,39 @@ return it::
         resp = make_response(render_template('error.html'), 404)
         resp.headers['X-Something'] = 'A value'
         return resp
+
+
+APIs with JSON
+``````````````
+
+A common response format when writing an API is JSON. It's easy to get
+started writing such an API with Flask. If you return a ``dict`` from a
+view, it will be converted to a JSON response.
+
+.. code-block:: python
+
+    @app.route("/me")
+    def me_api():
+        user = get_current_user()
+        return {
+            "username": user.username,
+            "theme": user.theme,
+            "image": url_for("user_image", filename=user.image),
+        }
+
+Depending on your API design, you may want to create JSON responses for
+types other than ``dict``. In that case, use the
+:func:`~flask.json.jsonify` function, which will serialize any supported
+JSON data type. Or look into Flask community extensions that support
+more complex applications.
+
+.. code-block:: python
+
+    @app.route("/users")
+    def users_api():
+        users = get_all_users()
+        return jsonify([user.to_json() for user in users])
+
 
 .. _sessions:
 
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -679,23 +679,26 @@ See :ref:`error-handlers` for more details.
 About Responses
 ---------------
 
-The return value from a view function is automatically converted into a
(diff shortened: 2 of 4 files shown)


ThiefMaster:
Good idea, but I think this should be done on the application level.

Also, if I can return a dict to be jsonified, I'd expect to be able to do the same with let's say a number, boolean or a list - especially a list is quite common for JSON responses.

davidism:
This seems like a good idea. SO questions about "dict object is not callable" are common. This is already pretty common example of how to extend `Flask.make_response`.

I agree with only handling dict for now. List is problematic because returning a 2- or 3-tuple/list has meaning already, and iterables generally have a different meaning in WSGI. There's no way to return a string and get a JSON string. If we want to extend this to number, boolean, and null, we can address that later.

ThiefMaster:
Wouldn't it be more consistent to fail with a clear error message when returning a dict? And then document how to jsonify all returned values for the app / a blueprint?

davidism:
I had been thinking about making the error message nicer, since "X is not callable" is just confusing. I think we can both support `dict` and have a better error message for unrecognized types.

davidism:
ThiefMaster brings up that this might make it seem that we're endorsing "APIs should return dicts", which while a good general rule isn't always true. I'm not convinced one way or another yet. `jsonify` itself didn't support other types until recently, and it's still available to use directly.

Is there ever a case where we'd want to make a dict signify something about the response like a tuple does now? Perhaps as ASGI support is added?

The lowest impact way to change this is to make a nice error message that mentions jsonify and points to the docs.

pgjones:
I think "APIs should return dicts" is a fine thing to endorse, and this is simply some sugar to make things easier (it doesn't actually make returning something else as JSON any more difficult). 

I'm not aware of any extra meaning to a dict response and in my experience of ASGI with Quart I've not come across any desire to do this (other than for JSON).

I'm happy to add the error message with this, but the aim in my view is the simplified way to return JSON. I think the tuple shortcut is a strength of Flask, as it makes a HTML response very easy to construct - my hope is that this does so for JSON responses.

ThiefMaster:
Why would you want to return a dict for an API that returns a collection, and does not have metadata associated with the collection itself (and thus requires an object on the top level).

Something like [this](https://developer.github.com/v3/orgs/#list-your-organizations) in the GitHub API for example. It's a collection and thus a list - and pagination is handled solely out-of-band through headers.

In the past having an array on the top level was bad because of browser security problems that allowed circumventing the same-origin policy, but those have been fixed years ago.

ThiefMaster:
Don't get me wrong, I'm all for making it easy to write proper APIs - but I'd rather have to opt-in to this behavior (e.g. for an API blueprint of the whole app), and in that case jsonify *any* return value that can be jsonified (ie pretty much anything that's not a response or a `data, statuscode` tuple)

davidism:
Opting in isn't much different than what users already have to do to support this (they'd still need need to be aware that it's even possible, and import a separate `Flask` class, just wouldn't have to add a few lines to `make_response`).

This sort of seems to be getting into "what is good API design" territory. At some point we need to assume users with opinions about API design, such as pagination in headers or returning lists (which are both fine), will do that regardless of any default Flask provides. Or they'll choose an extension such as Flask-Restful and not be affected by it anyway. For beginners, this is one less thing to trip up on.

pgjones:
Given the discussion I've gone ahead and updated the documentation and tests (am I missing something? I expected I'd need to do more than this). I think this should be merged and would encourage your detailed review.

lepture:
This seems not a general use case, there are many cases for API response, for example:

1. list: `[]`
2. with status code: `{}, 201`

I think what you want can be done with a custom `Flask`:

```python
class MyFlask(Flask):
    def make_response(self, rv):
        if isinstance(rv, dict):
            return jsonify(rv)
        return super(MyFlask, self).make_response(rv)
```

pgjones:
Thanks for you interest @lepture, I had hoped the discussion above answered your points - but now I think it isn't clear (any chance you could point out where?). I've also written this [article](https://medium.com/octopus-wealth/returning-json-from-flask-cf4ce6fe9aeb) which I hope clearly makes the case for this change.

davidism:
I'm going to say yes to this. The coverage of a common use case and tripping point for new users outweighs the restrictions it imposes on the format, and doesn't prevent using the already well-documented patterns and libraries for more complex API design.

davidism:
@pgjones I force pushed to get this in line with the latest changes and added a changelog. You're going to have to reset your local master branch after this is merged.

```
git fetch origin
git checkout master
git reset --hard origin/master
```

davidism:
I added a section to the quickstart titled "APIs with JSON". We should probably break out the JSON support docs into their own page at some point. Some of it is in the API docs, but it doesn't make a ton of sense there.

pgjones:
🎉 😄 thanks @davidism

PR #3112, opened 2019-03-05
fix an error in quickstart doc

XikunZhang:


greyli:
Thanks, but the view function can return either  `(response, headers)` or `(response, status)`, both are valid return value.

PR #3121, opened 2019-03-12
Update README.rst for beginners

50storm:
Dear flask developers,

I'm a new flask user.

After installing flask according to README.md,I got stuck in "A Simple Example section".

I think it is just because I couldn't see any comment in "A Simple Example section".

In order to refine "A Simple Example section", I have just added some comments to "A Simple Example section".
 
you can see the modification below.
https://github.com/pallets/flask/compare/master...50storm:readme_branch

It would be greatful if you accept my requent.

Sincerely yours.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,8 @@ Install and update using `pip`_:
 A Simple Example
 ----------------
 
+1.Create hello.py
+
 .. code-block:: python
 
     from flask import Flask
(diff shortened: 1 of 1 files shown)


davidism:
The readme is really meant as a visual example more than anything. The docs have both a tutorial and quickstart guide that go into this in more detail. Thanks for working on this, but I'm going to leave it as-is for now.

PR #3124, opened 2019-03-18
fix 'static_url_path' defaulting for empty paths

elbeardmorez:
the documentation of the application object specifies static parameters:

static_url_path – can be used to specify a different path for the static
files on the web. **Defaults to the name of the static_folder folder.**

static_folder – the folder with static files that should be served at
static_url_path. Defaults to the 'static' folder in the root path of the
application.

so by my understanding,

`app=Flask(__name__, static_folder='')`

should allow for arbitrary (absolute) path structure for all static
files. it doesn't. but the following does:

`app=Flask(__name__, static_folder='', static_url_path='')`

given the above doc strings, that appears inconsistent

for this specific case (`static_folder=''`), stepping through the
codebase yielded:

<Rule '//<filename>' (GET, OPTIONS, HEAD) -> static> 
`re.compile('^\\|//(?P<filename>[^/].*?)$')`

which is an expression that can never be satisfied by a valid url

please consider this trivial patch. i hope the tests are appropriate
too, it was difficult to know how much of the mechanism to incorporate
for such a small change

commit message:

-prefix a path delimiter iff there's a path to delimit
-ensures a valid default static route rule is created on application
intialisation for the case 'static_folder=""' and implicit
'static_url_path'

todo
* add a changelog entry if this patch changes code

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -965,23 +965,26 @@ class _PackageBoundObject(object):
     )
     del _get_static_folder, _set_static_folder
 
-    def _get_static_url_path(self):
+    @property
+    def static_url_path(self):
+        """The URL prefix that the static route will be accessible from.
+
+        If it was not configured during init, it is derived from
+        :attr:`static_folder`.
+        """
         if self._static_url_path is not None:
             return self._static_url_path
 
         if self.static_folder is not None:
             basename = os.path.basename(self.static_folder)
-            return "/" + basename if basename else ""
+            return ("/" + basename).rstrip("/")
 
-    def _set_static_url_path(self, value):
+    @static_url_path.setter
+    def static_url_path(self, value):
+        if value is not None:
+            value = value.rstrip("/")
+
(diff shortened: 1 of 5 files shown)


davidism:
It looks like this was addressed in a different way by #3182, but this solution seems more correct.

flask/helpers.py
```diff
@@ -931,7 +931,8 @@ def _get_static_url_path(self):
             return self._static_url_path
 
         if self.static_folder is not None:
-            return '/' + os.path.basename(self.static_folder)
+            basename = os.path.basename(self.static_folder)
+            return basename and '/' + basename or ''
```
jab:
```suggestion
            return '/' + basename if basename else ''
```

elbeardmorez: hah.. i'd been doing mostly shell at the time (hence a command pipeline mental state). agree that that's more pythonic, thus better. would you prefer a forced push or an addition that you squash?

PR #3128, opened 2019-03-23
Fix grammar in docs

faheel:


davidism:
This is not more correct than what's currently there, and is less correct in some changes.

faheel:
Oops, looks like I need to go to bed :sweat_smile: Understood the statement now that you pointed it out.

PR #3129, opened 2019-03-23
Fix some HTML injection paths in examples

adamchainz:
These are unlikely to be copy-pasted by users but it's best practice to avoid it and other examples do.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -203,7 +203,7 @@ of the argument like ``<converter:variable_name>``. ::
     @app.route('/user/<username>')
     def show_user_profile(username):
         # show the user profile for that user
-        return 'User %s' % username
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -295,7 +295,7 @@ Python shell. See :ref:`context-locals`.
 
     @app.route('/user/<username>')
     def profile(username):
-        return '{}\'s profile'.format(username)
+        return '{}\'s profile'.format(escape(username))
```
kx-chen: Other examples also import `escape` from flask, I don't see why this shouldn't also.

adamchainz: Yes, my bad, missed this when reading through since the other examples I'm editing here are fragments without imports, but this is a whole file. Fixed!

PR #3147, opened 2019-04-08
Add Dockerfile for flaskr example

dduleba:
Hi,
I prepared  presentation about locust with flaskr as test app.
http://ddarus.pythonanywhere.com/www/index.html#/step-11

In my case using falskr in docker makes its much simpler.

Maybe it will be useful for someone else.

Best Regards,
Darek

### DIFF
--- a/examples/tutorial/Dockerfile
+++ b/examples/tutorial/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3-alpine
+
+ADD . /app
+WORKDIR /app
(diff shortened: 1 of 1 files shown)


davidism:
This may have a place in the docs somewhere, maybe in the deploy docs. But I'd like to keep the example and tutorial in sync as much as possible. Docker itself also covers Flask extensively. Thanks for working on this, but I'm going to close it for now.

PR #3149, opened 2019-04-11
use redirect for uwsgi

jli05:
If a uWSGI app handles query at the root path `/?query=xxx` then is mounted at the `https://host/app` by `uwsgi` and `nginx`, any query `https://host/app?query=xxx` would generate an HTTP error but `https://host/app/?query=xxx` will be successful.

This fix just redirects `https://host/app?query=xxx` to `https://host/app/?query=xxx`.

### DIFF
--- a/docs/deploying/uwsgi.rst
+++ b/docs/deploying/uwsgi.rst
@@ -50,7 +50,9 @@ Configuring nginx
 
 A basic flask nginx configuration looks like this::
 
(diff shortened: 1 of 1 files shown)


jli05:
This small fix only changes documentation. The failure of AppVeyor with Python 2.7 is spurious.

davidism:
Thanks for looking at this, but I've got a larger rewrite of the deploy docs in the works that has a more correct handling of a non-root path.

PR #3152, opened 2019-04-12
tox/pytest: support posargs, move to testpaths

blueyed:


setup.cfg
```diff
@@ -9,7 +9,7 @@ license_file = LICENSE
 
 [tool:pytest]
 minversion = 3.0
-testpaths = tests
+testpaths = tests examples
```
davidism: The reason this is set in tox is because I want people to be able to run `pytest` without having to install the examples.

blueyed: I see.. `{posargs:tests examples}` in tox.ini then?

blueyed: updated

davidism:
Ended up addressing this while moving to Azure. Thanks.

PR #3155, opened 2019-04-15
Bump Jinja2 requirements

zabbal:
The 2.10.1 version was released 2019-04-06 and contains important security fix.
This change allows flask users to avoid implicit insecure dependency.


ThiefMaster:
I disagree that this is an "important" security fix in this context, since Flask does not use Jinja's SandboxedEnvironment and thus the fix doesn't apply.

But I guess it's a good idea to bump the dependency anyway, even if it's just to avoid warnings from GitHub's dependency security checker.

zabbal:
I agree that it's unimportant for Flask, the "important security fix" is referring to Jinja2 version.

PR #3176, opened 2019-05-06
Allow "rt" mode for open_resource

### LINKED ISSUE #3163: open_resource() does not allow mode='rt'
In Python 3, `mode='rt'` is synonymous with `mode='r'` but Flask disallows it thinking it is not a valid mode for reading. While this is true in Python 2, Flask could (at least) provide a more meaningful error message, if not just allow it.

A simple solution might change these lines in `flask/helpers.py`:

```python
    def open_resource(self, resource, mode='rb'):
        """..."""  # truncated for example
        if mode not in ('r', 'rb'):
            raise ValueError('Resources can only be opened for reading')
        return open(os.path.join(self.root_path, resource), mode)
```

...to...

```python
    def open_resource(self, resource, mode='rb'):
        """..."""
        if mode == 'rt':
            mode = 'r'
        if mode not in ('r', 'rb'):
            raise ValueError('Resources can only be opened for reading')
        return open(os.path.join(self.root_path, resource), mode)
```

A heavier solution might use `io.open()` (and also allow `encoding`, but see #1740):

```python
    def open_resource(self, resource, mode='rb', encoding=None):
        """..."""
        if mode not in ('r', 'rb', 'rt'):
            raise ValueError('Resources can only be opened for reading')
        return io.open(os.path.join(self.root_path, resource), mode=mode, encoding=encoding)
```

### Expected Behavior

Calling `open_resource()` with `mode='rt'` should open a file in text mode (at least in Python 3).

```python
from flask import Flask

app = Flask(__name__)
with app.open_resource('README.md', mode='rt') as f:
    print(f.read())
```

### Actual Behavior

A misleading exception is raised.

```pytb
  File "mwe.py", line 5, in <module>
    with app.open_resource('README.md', mode='rt') as f:
  File "/home/goodmami/repos/OMW/py3env/lib/python3.6/site-packages/flask/helpers.py", line 1002, in open_resource
    raise ValueError('Resources can only be opened for reading')
ValueError: Resources can only be opened for reading
```

### Environment

* Python version: 3.6.7
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

seawolf42:
fixes #3163 

Note that @davidism has also recommended re-architecting this helper, David, do you want to own opening the new issue to track that?

### DIFF
--- a/flask/helpers.py
+++ b/flask/helpers.py
@@ -1024,7 +1024,7 @@ class _PackageBoundObject(object):
                          subfolders use forward slashes as separator.
         :param mode: resource file opening mode, default is 'rb'.
         """
-        if mode not in ('r', 'rb'):
+        if mode not in ('r', 'rb', 'rt'):
             raise ValueError('Resources can only be opened for reading')
         return open(os.path.join(self.root_path, resource), mode)
 


davidism:
Hey, I didn't realize this at the time, but someone else also submitted PR #3174 for this. Since they added tests as well, I'm going to use theirs. Thanks for prompting the discussion with me and working at the sprints, it was helpful!

seawolf42:
@davidism yeah, I saw that PR after I had submitted mine, so happy to defer the merge to tested code. Hopefully next year I'll be more fully productive 💯

PR #3177, opened 2019-05-06
docs/quickstart.rst - Adding URL Building details and moving it

lucasmocellin:
begginers like me get stucked at the URL Building section, as it is a little bit more advanced, so I moved URL Building down the text with Redirects and Error. Also Added a little description of the output expected when running "flask run..." with the new code in the URL Building section


lucasmocellin:
I'll make it a new one.

PR #3179, opened 2019-05-06
Fix MethodView inheritance Error (#3138)

### LINKED ISSUE #3138: MethodView methods not inherited
To reproduce:
```python
import flask
import flask.views

class BaseView(flask.views.MethodView):
    methods = ['GET', 'PROPFIND']

class MyView(BaseView):
    def get(self):
        return None, 200
    def propfind(self):
        return None, 200

app = flask.Flask(__name__)
app.add_url_rule('/test', view_func=MyView.as_view('test'))

for rule in app.url_map.iter_rules():
    print(rule, ','.join(rule.methods))
```

### Expected Behavior
````
/test OPTIONS,PROPFIND,GET,HEAD
/static/<path:filename> OPTIONS,GET,HEAD
````

### Actual Behavior
````
/test OPTIONS,GET,HEAD
/static/<path:filename> OPTIONS,GET,HEAD
````

### Environment

* Python version: 3.7.1
* Flask version: Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
* Werkzeug version: 0.15.2

### PR DESCRIPTION
**Pull request by Lee-W**
* Describe what this patch does to fix the issue.
    * Fix the error that `methods` variable in `MethodView` won't be inherited when it's not overwritten.
* Link to any relevant issues or pull requests.
    * closes #3138

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,12 +21,15 @@ Unreleased
 -   :func:`send_file` supports :class:`~io.BytesIO` partial content.
     (`#2957`_)
 -   :func:`open_resource` accepts the "rt" file mode. This still does
-    the same thing as "r". (:issue:`3163`)
+    the same thing as "r". :issue:`3163`
+-   The :attr:`MethodView.methods` attribute set in a base class is used
+    by subclasses. :issue:`3138`
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
 .. _#2957: https://github.com/pallets/flask/issues/2957
 .. _#2994: https://github.com/pallets/flask/pull/2994
 .. _#3059: https://github.com/pallets/flask/pull/3059
+.. _#3179: https://github.com/pallets/flask/pull/3179
 
 
 Version 1.0.3
--- a/flask/views.py
+++ b/flask/views.py
@@ -120,6 +120,10 @@ class MethodViewType(type):
         if "methods" not in d:
             methods = set()
 
+            for base in bases:
+                if getattr(base, "methods", None):
+                    methods.update(base.methods)
+
             for key in http_method_funcs:
                 if hasattr(cls, key):
                     methods.add(key.upper())
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -10,11 +10,10 @@
 """
 
 import pytest
+from werkzeug.http import parse_set_header
 
 import flask
 import flask.views
-
-from werkzeug.http import parse_set_header
 
 
 def common_test(app):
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: flask/views.py [hunk 1]
```diff
@@ -117,6 +117,9 @@ def __init__(cls, name, bases, d):
 
         if 'methods' not in d:
             methods = set()
+            for base in bases:
+                if base.__dict__.get('methods'):
```
qil026: should we use `hasattr()` instead?

flask/views.py
```diff
@@ -117,6 +117,9 @@ def __init__(cls, name, bases, d):
...
+                    methods = methods.union(base.methods)
```
qil026: can we simply use `methods.add()`, since you don't need `method` to do anything within the loop.

CHANGES.rst
```diff
@@ -20,11 +20,14 @@ Unreleased
     (`#3059`_)
 -   :func:`send_file` supports :class:`~io.BytesIO` partial content.
     (`#2957`_)
+-   Fix the inheritance error for :class:`MethodView`.
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
 .. _#2957: https://github.com/pallets/flask/issues/2957
 .. _#2994: https://github.com/pallets/flask/pull/2994
 .. _#3059: https://github.com/pallets/flask/pull/3059
+.. _#3059: https://github.com/pallets/flask/pull/3059
```
ThiefMaster: you duplicated the previous line

Lee-W: Thanks! I've removed it.

flask/views.py
Lee-W: I use it to check both the existence of the `method` variable and whether it's empty. Do you think it makes sense?

flask/views.py
Lee-W: Could you explain more? I don't quite get it. Thanks!

flask/views.py
ThiefMaster: `getattr(base, 'methods', None)` would do the same

Lee-W: Yes, that does make sense! Thanks for the clarification! I've fixed it.

flask/views.py
jab:
```python
if hasattr(base, 'methods'):
    methods.update(base.methods)
```

Lee-W: Yes, that does make more sense! Thanks, @jab!

PR #3180, opened 2019-05-06
docs/quickstart.rst - Adding URL Building details and moving it

lucasmocellin:
begginers like me get stucked at the URL Building section, as it is a little bit more advanced, so I moved URL Building down the text with Redirects and Error. Also Added a little description of the output expected when running "flask run..." with the new code in the URL Building section

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -249,67 +249,7 @@ trailing slash produces a 404 "Not Found" error. This helps keep URLs
 unique for these resources, which helps search engines avoid indexing
 the same page twice.
 
-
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I don't think this is a great idea: Even right now there are people who end up hardcoding URLs - so I think URL building is worth being mentioned early. It is something pretty much everyone using flask needs to know (and usually use).

davidism:
Thanks, but I agree that this is important, or at least there's not enough evidence that it should be moved. I'm open to other ideas for improving these docs if you have any.

PR #3182, opened 2019-05-06
Remove ending slash from static_url_path

### LINKED ISSUE #3134: static_url_path doesn't work if it ends with a slash
https://github.com/pallets/flask/blob/a74864ec229141784374f1998324d2cbac837295/flask/app.py#L560

Urljoin seems like more robust way to join urls. I just tried to set static_url_path = '/' and it failed, but static_url_path = '' succeed. IMHO, this is bad and unexpected behaviour.

qil026:

Fixes #3134 where flask fails if `static_url_path` ends with a slash, e.g. `/foo/`

Because the `/<path:filename>` param already starts with a slash, it is imperative the the `static_url_path` prefix does NOT end with a forward slash.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,8 @@ Unreleased
 -   The ``flask`` command entry point is simplified to take advantage
     of Werkzeug 0.15's better reloader support. This bumps the Werkzeug
     dependency to >= 0.15. :issue:`3022`
+-   Support ``static_url_path`` that ends with a forward slash.
+    :issue:`3134`
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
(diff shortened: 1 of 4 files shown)


qil026:
@davidism can you give some 👀to this PR?

CHANGES.rst
```diff
@@ -20,11 +20,15 @@ Unreleased
     (`#3059`_)
 -   :func:`send_file` supports :class:`~io.BytesIO` partial content.
     (`#2957`_)
+-   Support ``static_url_path`` that ends with a forward slash for Flask
```
ThiefMaster: I think the "for Flask app" isn't needed

qil026: yeah i also wasn't sure abt that.. let me remove it

PR #3184, opened 2019-05-06
Upload test results for Azure Pipelines

brcrista:
Passes posargs to tox for pytest to emit JUnit XML.

See: https://dev.azure.com/brcrista/Python/_build/results?buildId=956&view=ms.vss-test-web.build-test-results-tab


davidism:
Looks like @vtbassmatt beat you to it. :-) Let me know if you spot anything in the current config that should change.

PR #3186, opened 2019-05-06
Replace IRC with Discord in CONTRIBUTING file

singingwolfboy:


ThiefMaster:
I'd prefer to mention both; there are people who don't have Discord nor want to get it, and for those I think it's nice to mention the existence of the channel as well.

(but maybe `s/FreeNode/Freenode)`? :))

ThiefMaster:
Oh, and can we finally ditch that mailing list? Is anyone even using/reading it?

davidism:
Looks like this was addressed in #3159

PR #3195, opened 2019-05-13
Support dataclass in JSONEncoder (if available)

eruvanos:
Support dataclass to be encoded with jsonify using `dataclasses.as_dict()`.

### DIFF
--- a/flask/json/__init__.py
+++ b/flask/json/__init__.py
@@ -20,6 +20,10 @@ from jinja2 import Markup
 # depend anyways.
 from itsdangerous import json as _json
 
+try:
+    import dataclasses
+except ImportError:
+    dataclasses = None
 
 # Figure out if simplejson escapes slashes.  This behavior was changed
 # from one version to another without reason.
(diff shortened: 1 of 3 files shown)


flask/json/__init__.py
```diff
@@ -20,6 +20,11 @@
 # depend anyways.
 from itsdangerous import json as _json
 
+try:
+    import dataclasses
+    dataclass_available = True
+except ImportError:
+    dataclass_available = False
```
ThiefMaster: What about just setting `dataclasses = None` in this case and then also using that as a criterion on whether they are available or not?

tests/test_basic.py
```diff
@@ -1507,7 +1521,7 @@ def index():
 @pytest.mark.parametrize("use_reloader", [True, False])
 @pytest.mark.parametrize("propagate_exceptions", [None, True, False])
 def test_werkzeug_passthrough_errors(
-    monkeypatch, debug, use_debugger, use_reloader, propagate_exceptions, app
+        monkeypatch, debug, use_debugger, use_reloader, propagate_exceptions, app
```
ThiefMaster: your editor did something weird there

tests/test_basic.py
```diff
@@ -1289,6 +1289,20 @@ def test_jsonify_mimetype(app, req_ctx):
     assert rv.mimetype == "application/vnd.api+json"
 
 
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="requires python3.7 or higher")
+def test_make_response_with_jsonified_dataclass(app, req_ctx):
+    from dataclasses import dataclass
+
+    @dataclass
+    class Data:
+        name: str
```
ThiefMaster: that's a syntax error in older python versions, and thus breaking the whole file

eruvanos:
Hi,
I am facing a problem with the tests,
PyTest is failing because of the type hints (python < 3.5), but `dataclasses` can not be defined without those.

Do you have an idea, how to deal with that?

tests/test_basic.py
eruvanos: dataclasses can not be defined without those, any idea how to deal with that?

flask/json/__init__.py
eruvanos: Done

tests/test_basic.py
eruvanos: Fixed

tests/test_basic.py
eruvanos: Got it to work, did not know that it is possible to create dataclasses manually

eruvanos: Thank you for the fast review.

eruvanos:
@ThiefMaster Is something still missing or is it fine, like it is?

PR #3197, opened 2019-05-14
Allow selecting a different shell for the "flask shell" command

kissgyorgy:
If the selected shell is not installed, it will fall back to the standard Python interactive shell.
All shells got the same banner and context.

The standard shell is left as the default, so this will not break backward compatibilty and doesn't result in behavior change.

It could be argued that this should be in a different package, but these are the most common shells and having them directly in Flask itself is convenient IMO. I really hate the default Python shell, because there is no tab completion and the first thing I do is to install IPython and use that.

### DIFF
--- a/flask/cli.py
+++ b/flask/cli.py
@@ -807,9 +807,50 @@ def run_command(info, host, port, reload, debugger, eager_loading, with_threads,
     )
 
 
+def _start_bpython_shell(banner, ctx):
+    import bpython
+    bpython.embed(ctx, banner=banner)
+
+
+def _start_ipython_shell(banner, ctx):
+    from IPython import start_ipython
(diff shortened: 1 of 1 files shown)


ThiefMaster:
There's already https://pypi.org/project/flask-shell-ipython/ to switch to a different shell (ipython).

davidism:
Thanks for your interest. However, I don't want to add specific shells as built-in because it's not something I want to maintain as those shells change, or continue to add to the list.

I think previously when this came up we discussed adding a hook for packages to add `--ipython` like options rather than whole commands, but I think we decided not to do that.

I'd be open to a better default configuration that enables completion. I think Ned Batchelder has a blog about that which might be useful.

PR #3202, opened 2019-05-16
insert deliberate bug into flask.cli:main

twosigmajab:
testing


davidism:
Well, "all tests passed" isn't right. We could probably use pytest-xprocess to test this function, since we can't use the CLI test runner.

twosigmajab:
Sorry for the noise, submitted accidentally!

PR #3203, opened 2019-05-16
add Azure Pipelines to test coverage

vtbassmatt:
No functional changes


davidism:
Is there a way to get "pallets/flask" in the URL instead of the ID?

davidism:
Also, I'm disabling Travis and Appveyor, so we can just reduce this to a single "Test status:" line.

PR #3204, opened 2019-05-16
azure pipelines ci on 1.0.x branch

davidism:
backporting from master

includes some tweaks to try to get coverage on build page


davidism:
@vtbassmatt @brcrista could use some help here. I'm trying to enable codecov, I think one of you mentioned that they added native support for Azure, but it looks like the codecov-python library doesn't have that yet. I tried adding a secure variable, but it's ignored for PRs. Is there a recommended way to configure this?

vtbassmatt:
The native integration isn't enabled yet, sadly. And without sharing secrets in fork PRs (which I recommend), there's not really a workaround.

davidism:
Ok, that's fine, was making sure I didn't miss something. Thanks!

PR #3208, opened 2019-05-19
FakeSignal should stub connect_via method

singingwolfboy:
Blinker has a [`connect_via()`](https://pythonhosted.org/blinker/#blinker.base.Signal.connect_via) method. `_FakeSignal` should stub this method, as well as the others.

### DIFF
--- a/flask/signals.py
+++ b/flask/signals.py
@@ -32,19 +32,18 @@ except ImportError:
             self.name = name
             self.__doc__ = doc
 
+        def send(self, *args, **kwargs):
+            pass
+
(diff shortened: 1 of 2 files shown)


flask/signals.py
```diff
@@ -41,6 +41,8 @@ def _fail(self, *args, **kwargs):
 
         send = lambda *a, **kw: None
         connect = (
+            connect_via
+        ) = (
```
davidism: Weird Black formatting here. Can you break it up into multiple lines?

Issue #3218, opened 2019-05-25
Use @property as a decorator


davidism:
A bunch of properties in the codebase are defined using `property` without the decorator syntax, which, while valid, isn't very intuitive anymore.

```python
def _get_some_value(self):
    ...

def _set_some_value(self, value):
    ...

some_value = property(_get_some_value, _set_some_value, doc="...")
del _get_some_value, _set_some_value
```

```python
@property
def some_value(self):
    """..."""
    ...

@some_value.setter
def some_value(self, value):
    ...
```

scrosby:
I'm interested in this.

PR #3221, opened 2019-05-27
Updated property object to decorator

lagmoellertim:
Updated all of python's (still valid but outdated) property objects to decorators.

This solves the issue #3218


davidism:
Thanks for working on this! We ended up addressing the comments in #3237

PR #3232, opened 2019-05-31
Convert make_test_environ_builder into class (fixes #3207)

### LINKED ISSUE #3207: Convert make_test_environ_builder into EnvironBuilder subclass
Werkzeug's `EnvironBuilder` gained a `json_dumps` attribute, but Flask doesn't have an `EnvironBuilder` subclass right now to take advantage of that, and there's no way to configure the `Client` to use a different builder class anyway. Might be good to refactor `flask.testing.make_test_environ_builder` to be an `EnvironBuilder` subclass at some point, rather than a factory wrapper.

_Originally posted by @davidism in https://github.com/pallets/flask/pull/3181#issuecomment-493489553_

lordmauve:
Added a new class `flask.testing.EnvironBuilder` inheriting from `werkzeug.test.EnvironBuilder`. 

Logic from `make_test_environ_builder()` moved to the constructor of that class, and changed to simply instantiate the class, while issuing a `DeprecationWarning`.

I did explore making `json_dumps()` a regular method rather than a static method, to pick up `app`, but if anything was expecting to call `EnvironBuilder.json_dumps()` as a static method then this would break. Requires funky descriptor tricks to work both as a static method and an instance method under Python 2 and so didn't seem worth the code it would take.

### DIFF
--- a/flask/testing.py
+++ b/flask/testing.py
@@ -9,8 +9,9 @@
     :copyright: © 2010 by the Pallets team.
     :license: BSD, see LICENSE for more details.
 """
-
+import warnings
 import werkzeug
+import werkzeug.test
 from contextlib import contextmanager
 
 from click.testing import CliRunner
@@ -21,11 +22,9 @@ from flask.json import dumps as json_dumps
 from werkzeug.urls import url_parse
 
 
-def make_test_environ_builder(
-    app, path="/", base_url=None, subdomain=None, url_scheme=None, *args, **kwargs
-):
-    """Create a :class:`~werkzeug.test.EnvironBuilder`, taking some
-    defaults from the application.
+class EnvironBuilder(werkzeug.test.EnvironBuilder):
+    """An :class:`~werkzeug.test.EnvironBuilder`, that takes defaults from the
+    application.
 
     :param app: The Flask application to configure the environment from.
     :param path: URL path being requested.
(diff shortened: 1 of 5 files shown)


lordmauve:
Tests and stylecheck passed under Tox but somehow the Azure pipelines have not been requested for this PR. (I appear to be on a dodgy internet connection.)

flask/testing.py
```diff
@@ -45,41 +44,72 @@ def make_test_environ_builder(
...
-    if base_url is None:
-        http_host = app.config.get("SERVER_NAME") or "localhost"
-        app_root = app.config["APPLICATION_ROOT"]
+            if url.query:
+                sep = b"?" if isinstance(url.query, bytes) else "?"
+                path += sep + url.query
 
-        if subdomain:
-            http_host = "{0}.{1}".format(subdomain, http_host)
+        if "json" in kwargs:
```
davidism: Should be able to remove this part, since the base `EnvironBuilder` handles JSON.

davidism:
Looks good! Please add a changelog entry as well.

flask/testing.py
lordmauve: No, because the base class doesn't know about `app` and therefore does not take JSON serialisation settings from it.

This could be removed if `json_dumps()` was added as a method that respects `self.app`, but this is hard if json_dumps() needs to be callable as a static method. If that's not needed, then yes, I can remove this.

lordmauve: In fact, I'm pretty sure it is a safe assumption that it's not needed. Let me go ahead and do this.

flask/testing.py
```diff
@@ -45,41 +44,72 @@ def make_test_environ_builder(
...
+        super(EnvironBuilder, self).__init__(path, base_url, *args, **kwargs)
 
-        url = url_parse(path)
-        base_url = "{scheme}://{netloc}/{path}".format(
-            scheme=url.scheme or url_scheme,
-            netloc=url.netloc or http_host,
-            path=app_root.lstrip("/"),
-        )
-        path = url.path
+    def json_dumps(self, obj, **args):
```
davidism: Use `**kwargs` for keyword args.

flask/testing.py
davidism: Yeah, that should work.

flask/testing.py
lordmauve: Whoops!

Issue #3254, opened 2019-06-08
Refactoring explain_template_loading_attempts method in debughelpers.py


aligoren:
Before starting, I'm so sorry because this issue can be totally ridiculous.

Hi. I've been doing a review of the Flask code base for a while.

I found this method

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L124-L169

I'm also using SonarQube to review codes. SonarQube warned me about Cognitive Complexity. There is documentation here;

https://www.sonarsource.com/docs/CognitiveComplexity.pdf

For example,

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L130

this part can be changed like that.

```python
if reqctx and reqctx.request.blueprint is not None:
```

The other thing is;

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L146-L150

this part can be changed like that.

```python
detail = "no match"
if triple is not None:
   detail = "found (%r)" % (triple[1] or "<string>")
   total_found += 1
```

I think we can say; if the triple variable is **None**, don't run this `total_found += 1` statement.

The last one is;

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L154-L159

I think this part can be changed too. Because `total_found` variable initialized with **0** value.

So, there will be two situations always. for **0** or for **> 1**.

When `total_found` is 0 or 1, the `seems_fishy` will be True. So, we may don't need to reassign this variable for each control flow. Instead of that, we can change this code like below;

```python
info_msg = "Error: the template could not be found."
seems_fishy = True
if total_found > 1:
   info_msg = "Warning: multiple loaders returned a match for the template."
```

jab:
Thanks for taking the time to try to contribute! It’s absolutely not ridiculous to try to make code improvements like the ones you mention. Could you please submit pull request(s) with your proposed changes? And try to put only related changes in each pull request. I think that would make it easier for maintainers to review what’s being proposed. In the meantime, I’ll close this issue in favor of the more targeted PRs.

PR #3255, opened 2019-06-08
Refactored explain_template_loading_attempts method

aligoren:

Before I mentioned about this PR.

I found this method

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L124-L169

I'm also using SonarQube to review codes. SonarQube warned me about Cognitive Complexity. There is documentation here;

https://www.sonarsource.com/docs/CognitiveComplexity.pdf

For example,

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L130

this part can be changed like that.

```python
if reqctx and reqctx.request.blueprint is not None:
```

The other thing is;

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L146-L150

this part can be changed like that.

```python
detail = "no match"
if triple is not None:
   detail = "found (%r)" % (triple[1] or "<string>")
   total_found += 1
```

I think we can say; if the triple variable is **None**, don't run this `total_found += 1` statement.

The last one is;

https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/debughelpers.py#L154-L159

I think this part can be changed too. Because `total_found` variable initialized with **0** value.

So, there will be two situations always. for **0** or for **> 1**.

When `total_found` is 0 or 1, the `seems_fishy` will be True. So, we may don't need to reassign this variable for each control flow. Instead of that, we can change this code like below;

```python
info_msg = "Error: the template could not be found."
seems_fishy = True
if total_found > 1:
   info_msg = "Warning: multiple loaders returned a match for the template."
```
**518 tests are passed and 14 skipped**.

#3254

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,6 +71,7 @@ Unreleased
 -   The ``flask run`` command no longer fails if Python is not built
     with SSL support. Using the ``--cert`` option will show an
     appropriate error message. :issue:`3211`
+-   Refactored explain_template_loading_attempts. :issue: `3254`
 
 .. _#2935: https://github.com/pallets/flask/issues/2935
 .. _#2957: https://github.com/pallets/flask/issues/2957
--- a/src/flask/debughelpers.py
+++ b/src/flask/debughelpers.py
@@ -122,7 +122,11 @@ def _dump_loader_info(loader):
 
 
 def explain_template_loading_attempts(app, template, attempts):
-    """This should help developers understand what failed"""
+    """This should help developers understand what failed
+
+    .. versionchanged:: 1.0.x
+       Refactored codes.
+    """
     info = ['Locating template "%s":' % template]
     total_found = 0
     blueprint = None
--- a/src/flask/debughelpers.py
+++ b/src/flask/debughelpers.py
@@ -155,8 +155,6 @@ def explain_template_loading_attempts(app, template, attempts):
 
         info.append("       -> %s" % detail)
 
-    seems_fishy = False
-
     info_msg = "Error: the template could not be found."
     seems_fishy = True
     if total_found > 1:
--- a/src/flask/debughelpers.py
+++ b/src/flask/debughelpers.py
@@ -127,7 +127,8 @@ def explain_template_loading_attempts(app, template, attempts):
     total_found = 0
     blueprint = None
     reqctx = _request_ctx_stack.top
-    if reqctx is not None and reqctx.request.blueprint is not None:
+
+    if reqctx and reqctx.request.blueprint is not None:
         blueprint = reqctx.request.blueprint
 
     for idx, (loader, srcobj, triple) in enumerate(attempts):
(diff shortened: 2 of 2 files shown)


src/flask/debughelpers.py
```diff
@@ -143,20 +148,21 @@ def explain_template_loading_attempts(app, template, attempts):
...
-            detail = "no match"
-        else:
+        detail = "no match"
+        if triple is not None:
             detail = "found (%r)" % (triple[1] or "<string>")
             total_found += 1
+
         info.append("       -> %s" % detail)
 
     seems_fishy = False
```
isidentical: i think you dont need this.

aligoren: Yes, I think so. I totally forgot about it. I'll commit again.

src/flask/debughelpers.py
```diff
@@ -122,12 +122,17 @@ def _dump_loader_info(loader):
...
+    .. versionchanged:: 1.0.x
+       Refactored codes.
+    """
     info = ['Locating template "%s":' % template]
     total_found = 0
     blueprint = None
     reqctx = _request_ctx_stack.top
-    if reqctx is not None and reqctx.request.blueprint is not None:
+
+    if reqctx and reqctx.request.blueprint is not None:
```
davidism: This line should not be changed. `if thing` and `if thing is not None` have different meanings. The later is also faster because it avoids bool conversion.

src/flask/debughelpers.py
```diff
@@ -143,20 +148,19 @@ def explain_template_loading_attempts(app, template, attempts):
...
 
-    seems_fishy = False
-    if total_found == 0:
-        info.append("Error: the template could not be found.")
-        seems_fishy = True
-    elif total_found > 1:
-        info.append("Warning: multiple loaders returned a match for the template.")
-        seems_fishy = True
+    info_msg = "Error: the template could not be found."
+    seems_fishy = True
```
davidism: This appears to have changed the logic of `seems_fishy`.

CHANGES.rst
```diff
@@ -71,6 +71,7 @@ Unreleased
 -   The ``flask run`` command no longer fails if Python is not built
     with SSL support. Using the ``--cert`` option will show an
     appropriate error message. :issue:`3211`
+-   Refactored explain_template_loading_attempts. :issue: `3254`
```
davidism: This sort of change should not have a changelog.

src/flask/debughelpers.py
```diff
@@ -122,12 +122,17 @@ def _dump_loader_info(loader):
 
 
 def explain_template_loading_attempts(app, template, attempts):
-    """This should help developers understand what failed"""
+    """This should help developers understand what failed
+
+    .. versionchanged:: 1.0.x
```
davidism: This is not relevant to users, so should not have a changelog.

davidism:
Overall, these changes do not seem beneficial, only different. I appreciate you taking the time to look, but I'm not really clear why I should merge this right now.

aligoren:
> Overall, these changes do not seem beneficial, only different. I appreciate you taking the time to look, but always take automated code quality results with a grain of salt.

And, should I close this PR req?

Thanks.

davidism:
I'll close this for now. I'm open to changes that make code more readable or easier to maintain, but there should be a clear advantage to the new code for me to accept it. Thanks!

PR #3257, opened 2019-06-10
File upload code fixed

anuragphadnis:
Converted filename into string, File upload didn't worked without doing it

The file upload code didn't worked without converting file name to string in the function file.save()
Changed the code so that the code in the documentation will work without any modification.


ThiefMaster:
What exactly did you pass there that wasn't already a string?

davidism:
I'm not clear why this change is needed.

PR #3260, opened 2019-06-12
adds api error handling example to documentation

LBWright:

This adds documentation to the error handling page. I was having a hard time finding anything on the proper use of abort with the flask error handler. After settling on a good pattern, I added it here.

### DIFF
--- a/docs/patterns/errorpages.rst
+++ b/docs/patterns/errorpages.rst
@@ -97,3 +97,30 @@ An example template might be this:
       <p>What you were looking for is just not there.
       <p><a href="{{ url_for('index') }}">go somewhere nice</a>
     {% endblock %}
+
+
+Handling API Errors with Abort
+------------------------------
+
+When using Flask for web APIs, handling errors is as simple as the previous examples but with minor changes.
+
+Register the error handler::
+
+    from flask import jsonify
+
+    @app.errorhandler(404)
+    def resource_not_found(e):
+        # if passing in an Exception object directly, you may need to convert it to a string
+        return jsonify(error=str(e)), 404
+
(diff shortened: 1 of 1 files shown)


docs/patterns/errorpages.rst
```diff
@@ -97,3 +97,30 @@ An example template might be this:
...
+When using Flask for web APIs, handling errors is as simple as the previous examples but with minor changes.
+
+Register the error handler::
+
+    from flask import jsonify
+
+    @app.errorhandler(404)
+    def resource_not_found(e):
+        # if passing in an exception directly, you may need to convert it to a string
+        return jsonify(error=e)
```
davidism: As far as I know, `json.dumps` won't be able to serialize an `Exception` object, so this won't work.

LBWright: You're right. That was the comment in the example. I'll update the code to be a little more clear.

LBWright: Added more clarity.

docs/patterns/errorpages.rst
```diff
@@ -97,3 +97,31 @@ An example template might be this:
...
+
+Register the error handler::
+
+    from flask import jsonify
+
+    @app.errorhandler(404)
+    def resource_not_found(e):
+        # if passing in an Exception object directly, you may need to convert it to a string
+        stringified_exception = str(e)
+        return jsonify(error=stringified_exception)
```
davidism: This no longer returns a 404 status.

LBWright: Thanks. I'll rebase to squash these commits down and update.

PR #3261, opened 2019-06-12
fix tests failing with server name warnings

davidism:
After pallets/werkzeug#1577, mismatched configured and real server names will show a warning in addition to raising 404. This caused tests that did this deliberately to fail.

This patch removes the pytest fixture we were using to fail on warnings, instead using the standard `-Werror` option. This speeds up the tests by ~3x.


davidism:
Azure is having issues with a bad Python build, but the full suite passes locally.

PR #3267, opened 2019-06-20
Use a working example to generate secret_key

solsticedhiver:
* Fixing the example given in the doc that only gives you a non-printable ascii output.
* SystemRandom is sure to use for cryptographic purpose (see random module doc)

The one-liner given in the doc provides only a non-printable output not suitable for any use. Fixing that with another one-liner.

https://docs.python.org/3/library/random.html


davidism:
The current output, as shown in the example, is fine as-is. `os.urandom` is already a source of (as true as possible) randomness.

PR #3276, opened 2019-07-01
Add closing tags to example's nav bar

jnduli:
In the tutorial, the base.html file does not have closing `</li>` tags. This PR adds the corresponding `</li>` tags to the nav bar.


davidism:
These are not required to be valid HTML. See other closed PRs for previous discussions.

PR #3282, opened 2019-07-01
use app.name as app.logger name

### LINKED ISSUE #2866: DispatcherMiddleware with different loggers per app in flask 1.0
After upgrading to flask 1.0 logging from different apps using DispatcherMiddleware, each log emitted is written to all handlers in the different apps. I assume this caused by `app.logger` always having the name `flask.app`, maybe?

Here is a example:

```
from werkzeug.wsgi import DispatcherMiddleware
from flask import Flask
from logging.handlers import RotatingFileHandler


handler1 = RotatingFileHandler('app1.log')
app1 = Flask('app1')
app1.logger.addHandler(handler1)

handler2 = RotatingFileHandler('app2.log')
app2 = Flask('app2')
app2.logger.addHandler(handler2)


@app1.route("/")
def hello():
    app1.logger.error("from app1")
    return ''


@app2.route("/")
def hello2():
    app2.logger.error("from app2")
    return ''


app = DispatcherMiddleware(app1, {
    '/app2': app2
})
```

Run with
```
uwsgi --socket 0.0.0.0:8000 --protocol=http -w app --callable app
```

And then make a request to / and /app2/. Each error log will be written in both logfiles.

### Environment

* Python version: 3.6.5
* Flask version: 1.0.2
* Werkzeug version: 0.14.1

My actual app is using `current_app.logger` with blueprints with the same behaviour, but I assume it the same issue.

davidism:
#2436 greatly simplified how logging is set up in Flask. Part of that removed the `LOGGER_NAME` config and hard-coded `"flask.app"`. However, #2866 points out that due to this it is no longer possible to distinguish logs when dispatching multiple apps.

This partially reverts the logger name behavior change from 1.0. There is still no config, but the logger's name is now the same as `app.name` (basically the value passed to `Flask(import_name)`).

The information that's currently logged does make sense as coming from the user's project rather than Flask itself, although we may introduce the `"flask.app"` logger again in the future for logging things truly internal to Flask and not the user's app.

This may require projects to update their logging config to reference their app's name instead of `"flask.app"`.

Closes #2866

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,12 @@ Unreleased
     -   :meth:`Flask.finalize_request` is called for all unhandled
         exceptions even if there is no ``500`` error handler.
 
+-   :attr:`Flask.logger` takes the same name as
+    :attr:`Flask.name` (the value passed as
+    ``Flask(import_name)``. This reverts 1.0's behavior of always
+    logging to ``"flask.app"``, in order to support multiple apps in the
+    same process. This may require adjusting logging configuration.
+    :issue:`2866`.
 -   :meth:`flask.RequestContext.copy` includes the current session
     object in the request context copy. This prevents ``session``
     pointing to an out-of-date object. :issue:`2935`
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -382,7 +382,7 @@ The following configuration values are used internally by Flask:
 
 .. versionchanged:: 1.0
     ``LOGGER_NAME`` and ``LOGGER_HANDLER_POLICY`` were removed. See
-    :ref:`logging` for information about configuration.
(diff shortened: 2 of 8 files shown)


twosigmajab:
Just noticed http://flask.pocoo.org/docs/logging/ (redirects to http://flask.pocoo.org/docs/1.0/logging/ and) doesn't have these changes (nor the other 1.1.0 changes, I'm guessing). Anything I can do to help with that?

davidism:
I need to log in to pocoo and disable the old site. The docs are at https://flask.palletsprojects.com.

jab:
Oh awesome!  🎉
I guess we need to get redirects set up too - let me know if I can help!

PR #3285, opened 2019-07-03
Move dotenv warning to stderr

ThiefMaster:
It was printed to stdout, which broke custom flask subcommands whose output was consumed by scripts (which might expect json or something else that's not just text).


davidism:
This doesn't match the rest of the `flask run` output, which prints to stdout. Is that going to be an issue?

ThiefMaster:
I thought about this too, but `flask run` is not something that will evr produce output which is useful when processed programmatically, while `flask <something>` may very much output something that's consumed by another program.

PR #3286, opened 2019-07-04
Added table of contents to README.rst

abhinavsagar:


ThiefMaster:
Does this really add any benefit? The readme is so short that this feels more noisy than useful to me.

davidism:
Thanks, but I don't plan on changing the readme format at this time.

PR #3290, opened 2019-07-07
Comment out helpful tips in ISSUE_TEMPLATE so they're visible only when editing

aadibajpai:
The comments are now visible only to the issue creator and do not need to be manually removed before submitting the issue.


jab:
Thank you for the contribution!

aadibajpai:
Thanks for the framework in the first place!

davidism:
Then again, I did that on purpose to make sure people were paying attention.

aadibajpai:
> 
> 
> Then again, I did that on purpose to make sure people were paying attention.

Not sure it makes a difference, I found out I read it when it appears suddenly because it does not appear in the preview. That happened when opening this PR

jab:
Sorry @davidism, want me to revert?

PR #3292, opened 2019-07-08
restore and deprecate json_available

### LINKED ISSUE #3288: Removing json_available reduces backward compatibility
### Expected Behavior

json_available should be available

```python
from flask import json_available
```

### Actual Behavior

```pytb
ImportError: cannot import name 'json_available'
```

### Environment

* Python version: 3.7.3
* Flask version: 1.1.0
* Werkzeug version: 0.15.4

davidism:
`flask.json_available` was removed because it was marked deprecated in the code and doesn't do anything. `flask.json` is always available, and libaries that weren't sure about that should have used `try/except ImportError` while it was relevant. However, since a few dormant but popular extensions use it, it's added back in for now.

Anything doing `if json_available` or `if json_available == True`, such as Flask-Testing and Flask-DebugToolbar, will show a `DeprecationWarning`. `is` can't be overridden, so that won't show a warning, but I don't think things that were using this flag did that.

I've marked it for removal in 2.0.0, which will be a pretty big change anyway. Please reach out to extensions you use if you see this warning and offer to help maintain it if it's an extension you want to continue using as Flask moves forward.

closes #3288

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,15 @@
 .. currentmodule:: flask
 
+Version 1.1.1
+-------------
+
+Unreleased
+
+-   The ``flask.json_available`` flag was added back for compatibility
+    with some extensions. It will raise a deprecation warning when used,
+    and will be removed in version 2.0.0. :issue:`3288`
+
+
 Version 1.1.0
 -------------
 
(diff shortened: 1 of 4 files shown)


frankcash:
when should one be able to pull this change?

aadibajpai:
Makes sense, thanks!

PR #3296, opened 2019-07-10
fixed example code in "uploading files" documentation

xorb0ss:
The documentation for [uploading files](`docs/patterns/fileuploads.rst`) was not working due to an incorrect function name. 

I also changed the `file` variable to something _"less"_ reserved, even though it's not officially a reserved keyword, I don't agree it's good practice using it

### DIFF
--- a/docs/patterns/fileuploads.rst
+++ b/docs/patterns/fileuploads.rst
@@ -45,7 +45,7 @@ executes them, but who has PHP installed on their server, right?  :)
 Next the functions that check if an extension is valid and that uploads
 the file and redirects the user to the URL for the uploaded file::
 
-    def allowed_file(filename):
+    def allowed_ext(filename):
         return '.' in filename and \
                filename.rsplit('.', 1)[1].lower() in ALLOWED_EXTENSIONS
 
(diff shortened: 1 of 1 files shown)


davidism:
This is now incorrect. After uploading a file with the `upload_file` route, you are redirected to view the file at the `uploaded_file` route.

xorb0ss:
The code, as is, does not run, because the endpoint referenced, `uploaded_file` does not exist. My commit fixes this:
`werkzeug.routing.BuildError: Could not build url for endpoint 'uploaded_file' with values ['filename']. Did you mean 'upload_file' instead?`

davidism:
Keep reading that section :wink:

> Now one last thing is missing: the serving of the uploaded files. In the `upload_file()` we redirect the user to `url_for('uploaded_file', filename=filename)`, that is, `/uploads/filename`. So we write the `uploaded_file` function to return the file of that name.

xorb0ss:
My bad!

ThiefMaster:
`file` is the most ridiculous builtin (since you use `open()` and not `file()`), and luckily it was removed in Python 3. So I don't think shadowing it is bad :) especially with it not shadowing anything in py3.

PR #3328, opened 2019-08-09
fixed typo in logging docstring

christopherzimmerman:
Really groundbreaking stuff here, noticed this when I was glancing at the logging source.  Just a simple typo fix.

### DIFF
--- a/src/flask/logging.py
+++ b/src/flask/logging.py
@@ -72,7 +72,7 @@ def _has_config(logger):
 
 
 def create_logger(app):
-    """Get the Flask apps's logger and configure it if needed.
+    """Get the Flask app's logger and configure it if needed.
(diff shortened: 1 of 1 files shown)


src/flask/logging.py
```diff
@@ -72,7 +72,7 @@ def _has_config(logger):
 
 
 def create_logger(app):
-    """Get the the Flask apps's logger and configure it if needed.
+    """Get the Flask apps's logger and configure it if needed.
```
jab: Can you change “apps’s” to “app’s” while you’re at it?

christopherzimmerman: figured that might have been too many fixes for one PR ;) -- it's changed now.

PR #3332, opened 2019-08-13
docs: Change max content length value to megabyte

Eido95:
16 * 1024 * 1024 = 16 MiB (Mebibyte)
16 * 1000 * 1000 = 16 MB (Megabyte)

The example is in megabytes, not in mebibytes.

### DIFF
--- a/docs/patterns/fileuploads.rst
+++ b/docs/patterns/fileuploads.rst
@@ -147,7 +147,7 @@ config key::
     from flask import Flask, Request
 
     app = Flask(__name__)
-    app.config['MAX_CONTENT_LENGTH'] = 16 * 1024 * 1024
+    app.config['MAX_CONTENT_LENGTH'] = 16 * 1000 * 1000
 
 The code above will limit the maximum allowed payload to 16 megabytes.
 If a larger file is transmitted, Flask will raise a


carc1n0gen:
>The mebibyte is closely related to the megabyte (MB). This can either be a synonym for mebibyte, or refer to 10^6 bytes = 1,000,000 bytes, depending on context [1]

[[1]](https://simple.wikipedia.org/wiki/Mebibyte). Unfortunately the references are dead links.  But most people I've ever interacted always think in 1024 of something, but don't use the "mebi|gibi|etc" versions.

Eido95:
@carc1n0gen , the quote you provided ends with:
> ... The two numbers are relatively close together, but mistaking the two has nevertheless led to problems.

This "mistake" led me to problems, which wouldn't be arise if 1000<sup>2</sup> was used rather than 1024<sup>2</sup>.

Additionally, because docs are meant for educational and guidance, it is better to stick with the [standards](https://en.wikipedia.org/wiki/IEEE_1541-2002), rather than unambiguous prefixes - [unless pointed out](https://physics.nist.gov/cuu/Units/binary.html).

Either way, the docs should clarify it better.

PR #3333, opened 2019-08-13
Add tip for startup of development server

tobias-hd:
An Windows, it is not possible for me to run `flask run` in development mode. I always run into
```
python.exe: No module named C:\Users\...\flask
```
This holds for PowerShell as well as for Git Bash.

I think it could be helpful to provide the solution to this issue in this docu.


carc1n0gen:
This issue will be fixed with a new release of werkzeug.  There's probably no reason to update docs since it will just get fixed.

davidism:
Thanks, but for the reason stated I'm going to close this

PR #3334, opened 2019-08-15
Simple 'if' use a single statement as func `templates_auto_reload` does

songroger:
code style related


ThiefMaster:
I don't think this particular change improves readability.

PR #3337, opened 2019-08-16
add a note about named params for execute()

zieone:
Flask apparently breaks sqlalchemy and requires you to pass named parameters as
[{}] 
no idea why, I'm to upset with Flask at the moment to go figure it out.
feel free to fix Flask for real, or fix this to be more sane, as I haven't gone spelunking, but this is the only way I could make named params actually work.


davidism:
I can't say why that style is not working for you, but there's nothing about Flask that would change how the SQLAlchemy API works. Please consult the SQLAlchemy docs and open an issue there if you feel something does not work as documented.

I understand you may be frustrated while trying to solve an issue, but please consider the tone you use when communicating in issues and pull requests in the future.

PR #3346, opened 2019-08-30
Update logging.rst

bearnun:
Describe default log level.


davidism:
Updated to explain why we're mentioning this.

PR #3360, opened 2019-09-10
Add feature flask.cli.with_testrequestcontext,

MaicoTimmerman:
For a project of mine I had the need to have a request context available in a cli command. This PR adds the code used, together with test and relevant code.

I have not added the ``versionadded`` yet, should it contain 1.1.2?

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -396,6 +396,28 @@ If you're sure a command doesn't need the context, you can disable it::
         ...
 
 
+
+Request Context
+~~~~~~~~~~~~~~~
+Commands requiring a request context can use
+:meth:`~cli.with_testrequestcontext` to to simulate a full request. ::
+
+    import click
+    from flask import request
+    from flask.cli import with_testrequestcontext
+
+    @click.command()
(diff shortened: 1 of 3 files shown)


davidism:
I don't think I want to include this, using a request context during something that's not request-like is a bad pattern to enable. Things outside requests should only require the app context. For example, in the test you've shown, it would make more sense to extract the report logic from the view that generates it, so it can be called independently of being in a request, such as by passing CLI options.

If it's not something you can change, you can already get a request context without another decorator (or at least a simpler decorator) by using `with_appcontext` and `current_app`:

```python
@app.cli.command('report')
@with_appcontext
def report_command():
    with current_app.test_request_context():
        # report stuff
```

davidism:
Flask-SQLAlchemy doesn't require a request context, only an app context.

MaicoTimmerman:
In our case we select i18n columns based on the language, which is actually dependent on request context  ;)

davidism:
That's an issue with i18n then, not Flask-SQLAlchemy. Whatever i18n provider you use should have a way to set the language outside a request context.

PR #3368, opened 2019-09-21
get_cookie_name in SessionInterface for easier overriding in SecureCookieSessionInterface

doronhorwitz:
Most of the other arguments passed into ``response.set_cookie()`` in ``SecureCookieSessionInterface.save_session()`` use the overridable methods of ``SessionInterface`` to fill in each argument. Except for the cookie name, which is statically set to ``app.session_cookie_name``.

Sometimes a cookie name needs to be set dynamically based on the current request, for example, a cookie name that is dependent on the current subpath in the URL. At the moment, the best way of doing this is to override the ``save_session()`` method of ``SecureCookieSessionInterface`` and make it dynamically pass in the cookie name to ``response.set_cookie()``. 

To avoid having to override the whole of ``save_session()`` this PR adds in an overridable ``get_cookie_name()`` method on ``SessionInterface`` to match the other ``get_xxx()`` methods of that class.

Note: I originally posted a question about this in Stackoverflow (_["Something like get_cookie_name in Flask's SessionInterface?"](https://stackoverflow.com/questions/55281983/something-like-get-cookie-name-in-flasks-sessioninterface)_) and also asked the question (using the Stackoverflow link) in the Flask Discord chat. I was not satisfied for the response I got, so I've opened up a PR.

### DIFF
--- a/src/flask/sessions.py
+++ b/src/flask/sessions.py
@@ -173,6 +173,13 @@ class SessionInterface(object):
         """
         return isinstance(obj, self.null_session_class)
 
+    def get_cookie_name(self, app):
+        """Returns the name of the session cookie.
+
+        Uses ``app.session_cookie_name`` which is set to ``SESSION_COOKIE_NAME``
+        """
+        return app.session_cookie_name
+
     def get_cookie_domain(self, app):
         """Returns the domain that should be set for the session cookie.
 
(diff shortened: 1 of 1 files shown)


doronhorwitz:
closing to fix the email address in the commit

PR #3369, opened 2019-09-22
get_cookie_name in SessionInterface for easier overriding in SecureCookieSessionInterface

doronhorwitz:
Most of the other arguments passed into ``response.set_cookie()`` in ``SecureCookieSessionInterface.save_session()`` use the overridable methods of ``SessionInterface`` to fill in each argument. Except for the cookie name, which is statically set to ``app.session_cookie_name``.

Sometimes a cookie name needs to be set dynamically based on the current request, for example, a cookie name that is dependent on the current subpath in the URL. At the moment, the best way of doing this is to override the ``save_session()`` method of ``SecureCookieSessionInterface`` and make it dynamically pass in the cookie name to ``response.set_cookie()``. 

To avoid having to override the whole of ``save_session()``, this pull request adds in an overridable ``get_cookie_name()`` method on ``SessionInterface`` to match the other ``get_xxx()`` methods of that class.

Note: I originally posted a question about this in Stackoverflow (_["Something like get_cookie_name in Flask's SessionInterface?"](https://stackoverflow.com/questions/55281983/something-like-get-cookie-name-in-flasks-sessioninterface)_) and also asked the question (using the Stackoverflow link) in the Flask Discord chat. I was not satisfied with the response I got, so I've opened up this pull request.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
 .. currentmodule:: flask
 
+Version 2.0.0
+-------------
+
+Unreleased
+
+-   Add :meth:`sessions.SessionInterface.get_cookie_name` to allow
+    setting the session cookie name dynamically. :pr:`3369`
+
+
 Version 1.1.2
 -------------
 
--- a/src/flask/sessions.py
+++ b/src/flask/sessions.py
@@ -173,6 +173,13 @@ class SessionInterface(object):
         """
         return isinstance(obj, self.null_session_class)
 
+    def get_cookie_name(self, app):
+        """Returns the name of the session cookie.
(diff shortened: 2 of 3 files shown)


davidism:
Makes sense. Needs a changelog and tests.

doronhorwitz:
Thanks @davidism. Added a test and changelog.

doronhorwitz:
@davidism, are my tests and changelog adequate? Is there anything more required?

davidism:
Looks good, I'll get to it when I can.

doronhorwitz:
great, thanks.

Issue #3350, opened 2019-09-03
Documentation for older versions of Flask is missing


untitaker:
https://flask.palletsprojects.com/en/0.12.x/deploying/wsgi-standalone/#proxy-setups used to be a valid link and no longer is.

@davidism is this intended? I can probably fix it in rtd.

untitaker:
as far as I can tell renaming 0.12-maintenance to 0.12.x and enabling the build in rtd would fix it

davidism:
I might need to port some docs fixes back to 0.12.x, but I don't mind re-enabling it.

davidism:
0.12.x docs are now available here: https://flask.palletsprojects.com/en/0.12.x/

untitaker:
thank you very much

PR #3376, opened 2019-09-24
update docs for new theme

davidism:
Apply the update docs config and theme to the 0.12.x branch so it can be built on Read the Docs.

related to #3350


marasondo:
newm

Issue #3354, opened 2019-09-08
DOCs: escape import needed to follow quickstart tutorial.


LeonardoFurtado:
### Expected Behavior
All tutorial steps should work

### Actual Behavior
error in method not imported

```pytb
return "User  %s" % escape(username)
NameError: name 'escape' is not defined
```

I think all the steps of the tutorial should be followed without errors, this helps the beginners. At [Quickstart](https://flask.palletsprojects.com/en/1.1.x/quickstart/#quickstart), [Variable Rules](https://flask.palletsprojects.com/en/1.1.x/quickstart/#variable-rules) generate an error because the escape method was not imported earlier in the tutorial. I could open a PR for that, but repeating the previous code doesn't make much sense to me, and I can't think of a way to not make it messy. So I'll open this issue before I sleep and forget it.

It may seem like a silly issue, but I always worry about documentation and how easy it should be for beginners to follow.

LeonardoFurtado:
Is updating the tutorial to use fstring a viable solution?

Jamim:
Hello @LeonardoFurtado,

> Is updating the tutorial to use fstring a viable solution?

I don't think so because `flask` still supports Python 2.7.

LeonardoFurtado:
> Hello @LeonardoFurtado,
> 
> > Is updating the tutorial to use fstring a viable solution?
> 
> I don't think so because `flask` still supports Python 2.7.

it's true...

agentm007:
Not sure why this was closed. It seems to still be a problem with the documentation.

LeonardoFurtado:
> Not sure why this was closed. It seems to still be a problem with the documentation.

This is reflected in many points of the documentation, I do not know if correcting all these points can make the examples very large. I will open this again :).

MarcoGorelli:
Why not just move this import (`escape`) further up in the documentation?

I think it's already implicit that to run an example, you need to already have run what comes before it. For example, to run the one in https://flask.palletsprojects.com/en/1.1.x/quickstart/#rendering-templates, it's implicit that to run `app = Flask(__name__)` from the beginning of the tutorial.

But looking for missing imports in later sections of the documentation isn't obvious, and errors early on can be discouraging to beginners

LeonardoFurtado:
> 
> But looking for missing imports in later sections of the documentation isn't obvious, and errors early on can be discouraging to beginners

Very good, that's exactly what I meant.

PR #3378, opened 2019-09-24
Added escape

johnkershaw:
Example 3 (user | post | path) currently fails because 'escape' has not yet been imported. Either re-insert the import here, or add 'escape' to the original import at the top.


LeonardoFurtado:
being discussed in #3354

PR #3382, opened 2019-10-06
Add hyperlink of contributing-testsuite in CONTRIBUTING

Mellcap:
Add hyperlink of `contributing-testsuite` in CONTRIBUTING.rst and remove the duplicate dot.


ThiefMaster:
Sphinx should do the linking automatically; also we do not want to link to GitHub in the docs.

PS: Please keep the first line of your commit messages short enough so GitHub doesn't wrap it. Ideally it's max 50 chars, and GitHub already gives you a bit more before wrapping.

ThiefMaster:
Sorry, didn't notice that this isn't a PR for the docs (the `docs` prefix in your commit msg was a bit misleading and this project doesn't use these commit msg prefixes anyway - `docs(CONTRIBUTING): `  is an incredible waste of space in the commit message).

I think adding the link is a good idea - but I'd make it a separate bullet point.

And please, amend your commit so its commit message doesn't get wrapped by GitHub. The preferred max length for the first line of a commit message is 50 characters. Github already gives you a bit more..

CONTRIBUTING.rst
```diff
@@ -121,7 +121,7 @@ Start coding
 .. _pre-commit: https://pre-commit.com
 .. _create a pull request: https://help.github.com/en/articles/creating-a-pull-request
 
-.. _contributing-testsuite:
+.. _contributing-testsuite: https://github.com/pallets/flask/blob/master/CONTRIBUTING.rst#running-the-tests
```
ThiefMaster: Isn't it enough to link to `#running-the-tests`?

Mellcap: Yeah, you are right, I have changed it.

PR #3386, opened 2019-10-11
Support loading configuration from toml files

pgjones:
TOML is a very popular format now, and is taking hold in the Python
ecosystem via pyproject.toml (among others). This adds toml as an
extra dependency usuable if installed with the toml extra.

I've recently added similar code in [Quart](https://gitlab.com/pgjones/quart/commit/e49d4f6eeaa71f1883f52f4154bb95d9871ef4ed) and I've been using toml configs with Flask and Quart for some time now. In my view TOML has crossed a threshold of usage to warrant direct support in Flask.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,7 @@ setup(
             "sphinxcontrib-log-cabinet",
             "sphinx-issues",
         ],
+        "toml": ["toml"],
     },
     entry_points={"console_scripts": ["flask = flask.cli:main"]},
 )
--- a/src/flask/config.py
+++ b/src/flask/config.py
@@ -201,6 +201,36 @@ class Config(dict):
             raise
         return self.from_mapping(obj)
 
+    def from_toml(self, filename, silent=False):
+        """Updates the values in the config from a TOML file. This function
+        behaves as if the TOML object was a dictionary and passed to the
+        :meth:`from_mapping` function.
+
+        Note: You will need to install Flask with the toml extra to use
+        this.
+
+        :param filename: the filename of the TOML file.  This can either be an
+                         absolute filename or a filename relative to the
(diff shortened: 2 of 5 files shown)


davidism:
I'm not super happy that `from_json` got in in the first place. The intention of Flask's built-in config always appeared to be that it should be Python. JSON is less expressive and more verbose, and while TOML is more expressive it still doesn't cover the types of config that are common now, such as loading values from the environment.

On the other hand, `config.from_*` does offer some functionality over loading a file yourself. Maybe we need a compromise instead, such as `from_file(name, load_func)`? That way anyone can use whatever loader they want, rather than having to write another function for each format.

davidism:
In a weird coincidence, here's a tweet from @jakevdp saying the same thing: https://twitter.com/jakevdp/status/1183241890289741825

pgjones:
I see this as a question of convenience and that there is a tipping point when something is so commonly used as warrant direct inclusion. Whilst it isn't clear, I think [`from_json`](https://github.com/pallets/flask/pull/827) reached this point in the community and adding it saved users writing very similar repeated code. My view is that TOML has also reached this point. 

I think that `from_mapping` could work as the generic loader, in that you can do something like,
```python
data = toml.load(something)
app.config.from_mapping(data)
```
although it doesn't include the useful error handling I presume `from_file` would.

carc1n0gen:
Just wanted to add I really like using json files to configure flask.  I prefer config files to not be executed code.  Allowing any format the developer wants sounds like a cool idea.

pgjones:
See #3398 as the potentially better solution.

davidism:
Thanks, closing in favor of #3398.

PR #3398, opened 2019-10-15
Support loading configuration from text files

pgjones:
TOML is a very popular format now, and is taking hold in the Python
ecosystem via pyproject.toml (among others). This allows toml config
files via,

    app.config.from_file("config.toml", toml.loads)

it also allows for any other file format whereby there is a loader
that takes a string and returns a mapping.

This is an alternative to #3386.

### DIFF
--- a/src/flask/config.py
+++ b/src/flask/config.py
@@ -15,7 +15,6 @@ import warnings
 
 from werkzeug.utils import import_string
 
-from . import json
 from ._compat import iteritems
 from ._compat import string_types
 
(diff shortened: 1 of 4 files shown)


src/flask/config.py
```diff
@@ -176,31 +176,48 @@ class and has ``@property`` attributes, it needs to be
...
-        .. versionadded:: 0.11
+        .. versionadded:: 1.2
         """
         filename = os.path.join(self.root_path, filename)
-
         try:
-            with open(filename) as json_file:
-                obj = json.loads(json_file.read())
+            with open(filename) as file_:
+                obj = loads(file_.read())
```
davidism: Should this use `load(file_)`? If someone has a loader that doesn't support a file handle, they can wrap it as `lambda f: json.loads(f.read())`.

src/flask/config.py
```diff
@@ -176,31 +176,48 @@ class and has ``@property`` attributes, it needs to be
...
+
+        :param filename: the filename of the JSON file.  This can either be an
+                         absolute filename or a filename relative to the
+                         root path.
+        :param silent: set to ``True`` if you want silent failure for missing
+                       files.
+
+        .. versionadded:: 0.11
+        """
+        return self.from_file(filename, json.loads, silent=silent)
```
davidism: This should issue a `DeprecationWarning` to be removed in 2.0.

tests/test_config.py
```diff
@@ -46,6 +47,13 @@ def test_config_from_json():
     common_object_test(app)
 
 
+def test_config_from_file():
```
davidism: Make this an optional dependency with [`pytest.importorskip`](https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency) so users can run `pytest` directly for most contributions.

davidism: Rename to `test_config_from_toml_file`.

davidism:
I'm not sure if it's worth pulling in TOML just for one test. Can this be tested with `json.load` instead?

tests/test_config.py
pgjones: I've removed it and added some docs instead.

src/flask/config.py
pgjones: Done.

pgjones:
I've moved the toml example to the docs (from the tests).

Issue #3404, opened 2019-10-25
route decorator on MethodView classes


dougthor42:
### TL;DR:
+ Being able to use a Blueprint's `@pb.route` decorator on a `MethodView` class can help made code more readable
+ It should be a pretty trivial fix.
+ I'm working on tests and a merge request right now.

### Expected Behavior
Users should be able to combine blueprints with `MethodView`.

```python
from flask import Blueprint
from flask.views import MethodView

api = Blueprint("api", __name__, url_prefix="/api/v1")

@api.route("/users")
@api.route("/users/<user_id>")
class UserAPI(MethodView):

    def get(self, user_id=None):
        if user_id is None:
            return "list of all users"
        else:
            return f"just user {user_id}"

    def post(self, user_id):
        return f"POST {user_id}"

    def delete(self, user_id):
        return f"DELETE {user_id}"

    def put(self, user_id):
        return f"PUT {user_id}"

    def PATCH(self, user_id):
        return f"PATCH {user_id}"
```

### Actual Behavior
Using a blueprint's `@route` decorator on a `MethodView` results in `TypeError` is raised:

```pytb
Traceback (most recent call last):
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/app.py", line 2463, in __call__
    return self.wsgi_app(environ, start_response)
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/app.py", line 2449, in wsgi_app
    response = self.handle_exception(e)
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/app.py", line 1866, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    return self.finalize_request(rv)
  File "/c/gitlab/mocvd/sic-inventory/.venv/lib/python3.6/site-packages/flask/app.py", line 1967, in finalize_request
    response = self.make_response(rv)
[...]

PR #3405, opened 2019-10-25
Add support for decorating MethodViews with @Blueprint.route()

dougthor42:
This updates the `Blueprint.route` decorator to support decorating `MethodView` classes.

Fixes Issue #3404.

#### Comments
+ I assume this will be part of the v2.0.0 release, so that's what I used for `versionchanged` and the changelog entry. If this is incorrect please let me know.
+ I wasn't sure where to put the tests - the code changes are part of `blueprints.py` but really only effect `views`, so I put it in `view.py`.
+ Let me know what other tests should be added.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Unreleased
 -   Add :meth:`Config.from_file` to load config using arbitrary file
     loaders, such as ``toml.load`` or ``json.load``.
     :meth:`Config.from_json` is deprecated in favor of this. :pr:`3398`
+-   Add support for decorating :class:`~flask.views.MethodView` with
+    blueprints' :meth:`flask.blueprint.Blueprint.route` method. :issue:`3404`
 
 
 Version 1.1.2
--- a/docs/views.rst
+++ b/docs/views.rst
@@ -136,6 +136,28 @@ That way you also don't have to provide the
 :attr:`~flask.views.View.methods` attribute.  It's automatically set based
 on the methods defined in the class.
 
+If you're using :ref:`blueprints`, you can decorate a
+:class:`~flask.views.MethodView` with the blueprint's ``@route`` decorator
+instead of manually adding the url rule to the ``app`` object. This lets you
+leverage the benefits of blueprints such as ``url_prefix``::
+
+    from flask import Blueprint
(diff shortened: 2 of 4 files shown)


davidism:
Thanks for looking at this. However, I don't think I want to merge this at this time.

You can already use `MethodView` with `Blueprint` with `bp.add_url_rule()`, which corresponds to `app.add_url_rule()`. I don't want to add a special case for certain types, even if they are built-in. This also obscures the behavior of `as_view()`, which is important for `View` classes.

davidism:
Going to close this due to lack of response. I'm open to merging this in the future, but it needs improvement.

dougthor42:
Sorry for the delay.

> I want the docs to be clear that `route` is a shortcut that doesn't pass class arguments and sets the endpoint name to the class name.

I've added a blurb for this to the docs. Let me know if you're happy with it. I also added a second example for Blueprints+MethodViews.

> apply to `add_url_rule` and `route` on both `Flask` and `Blueprint`, for any `View` subclass, not just `MethodView`

Can you elaborate on this? My naive interpretation is just:

```diff
diff --git a/src/flask/blueprints.py b/src/flask/blueprints.py
index 0e974ab2..13392174 100644
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -14,6 +14,7 @@ from functools import update_wrapper
 from .helpers import _endpoint_from_view_func
 from .helpers import _PackageBoundObject
 from .views import MethodViewType
+from .views import View

 # a singleton sentinel value for parameter defaults
 _sentinel = object()
@@ -283,7 +284,7 @@ class Blueprint(_PackageBoundObject):
             endpoint = options.pop("endpoint", f.__name__)

             # Support decorating MethodView classes. See Issue #3404
-            if isinstance(f, MethodViewType):
+            if isinstance(f, (View, MethodViewType)):
                 f = f.as_view(endpoint)

             self.add_url_rule(rule, endpoint, f, **options)
```

Is that what you're talking about or am I completely misunderstanding?

dougthor42:
Oh, and you can see the changes in https://github.com/dougthor42/flask/tree/blueprints-and-methodviews.

PR #3406, opened 2019-10-26
Change docs to use f-strings

erfanio:
I noticed old printf style string formatting in quite a few places in the documentation.
When I cloned to fix the quick start guide, I noticed it was already changed to f-strings style formatting in quickstart so I changed the rest too.

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -70,7 +70,7 @@ implement a blueprint that does simple rendering of static templates::
     @simple_page.route('/<page>')
     def show(page):
         try:
-            return render_template('pages/%s.html' % page)
+            return render_template(f'pages/{page}.html')
         except TemplateNotFound:
             abort(404)
 
--- a/docs/patterns/fabric.rst
+++ b/docs/patterns/fabric.rst
@@ -49,16 +49,16 @@ virtual environment::
     def deploy():
         # figure out the package name and version
         dist = local('python setup.py --fullname', capture=True).strip()
-        filename = '%s.tar.gz' % dist
+        filename = f'{dist}.tar.gz'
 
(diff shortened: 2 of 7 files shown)


ThiefMaster:
@VilhelmPrytz I don't think having Python3 syntax in docs is a problem

VilhelmPrytz:
> @VilhelmPrytz I don't think having Python3 syntax in docs is a problem

No, I actually agree with you 😄 But I still think it's worth mentioning somewhere in the docs that the guides work for Python 3 only (to avoid confusion).

erfanio:
> > @VilhelmPrytz I don't think having Python3 syntax in docs is a problem
> 
> No, I actually agree with you 😄 But I still think it's worth mentioning somewhere in the docs that the guides work for Python 3 only (to avoid confusion).

I totally agree and initially I was going to open an issue to discuss how to best approach this. But [getting started guide has already been changed to use f-string](https://github.com/pallets/flask/commit/413778afc19ce9f0cb484878d86375729c14faf4) so I thought someone has already made this decision.

VilhelmPrytz:
> > > @VilhelmPrytz I don't think having Python3 syntax in docs is a problem
> > 
> > 
> > No, I actually agree with you smile But I still think it's worth mentioning somewhere in the docs that the guides work for Python 3 only (to avoid confusion).
> 
> I totally agree and initially I was going to open an issue to discuss how to best approach this. But [getting started guide has already been changed to use f-string](https://github.com/pallets/flask/commit/413778afc19ce9f0cb484878d86375729c14faf4) so I thought someone has already made this decision.

Alright, that's a good point. If the decision already has been made, so be it :smile:

PR #3407, opened 2019-10-27
add send_from_filepath

aohanhongzhi:
I added a function named  send_from_filepath in helpers.py which user can download file from its absolute path and the filename just show to browser. sometimes , user store the file which rename  as MD5 value. so send_from_directory 。However, user can't get the correct name when downloading.
Best wish!


aohanhongzhi:
if have question ，please contact me as soon as possible

davidism:
Thanks for your interest, but I don't think this function belongs in the framework. It's not clear how this is better or more secure than the existing functions.

PR #3408, opened 2019-10-28
Make the documentation build reproducibly

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that flask could not be built reproducibly.

This is because it includes an absolute build directory in the documentation as the `json_module` attribute points to a Python class/module which has a string representation including its path.

This commit skips this (inherited) member from the documentation.

(This was originally filed in Debian as [#943674](https://bugs.debian.org/943674).)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -31,6 +31,7 @@ Incoming Request Data
 .. autoclass:: Request
    :members:
    :inherited-members:
+   :exclude-members: json_module
 
    .. attribute:: environ
 


davidism:
You can't exclude this, it's part of the documented api.

davidism:
If you're concerned with overall reproducibility, you should probably contribute a general fix to stabilize reprs in Sphinx.

lamby:
> You can't exclude this, it's part of the documented api.

Don't quite follow what you mean by "can't" here, but assuming you might clarify this with "should" or similar it is simply excluding this member from this internal wrapper class from the documentation - the current entry has no associated text whatsover, just the "`<module /path/to/foo.pyc>'` or whatever.

davidism:
That attribute is [documented in Werkzeug](http://werkzeug.palletsprojects.com/en/0.16.x/wrappers/#werkzeug.wrappers.json.JSONMixin.json_module), so if it's not showing up in Flask that's another thing that needs to be addressed in Sphinx.

The fact that non-reproducible internals such as module paths are output by Sphinx is a more general problem with Sphinx and should be addressed there rather than in individual docs. In this case, it would make more sense for that to be a reference to the `flask.json` module rather than a repr.

davidism:
May be related to https://github.com/sphinx-doc/sphinx/issues/6415

davidism:
It seems like even adding custom documentation doesn't get through if the attribute is aliasing a module. So I guess the best answer is to exclude it for now, and hope that linked Sphinx issue can be addressed.

PR #3412, opened 2019-11-01
Add `async` support

pgjones:
This will require Python 2 support to be dropped, which as I understand isn't likely to happen till 2020. Which is a bonus as it gives time to consider this change and the implications of it. (I also need to spend some time cleaning up this implementation, yet it serves as a proof of concept).

I've written [this blog post](https://pgjones.dev/blog/flask-async-quart-sync-2019/) to add some extra context, but I see a key aspect to this change is that whilst it adds `async` support it does not do so throughout and specifically does not support ASGI. This means that Flask is constrained (when using `async`) to have worse performance than the ASGI equivalents. This is necessary as I don't think Flask can be `async` throughout (or support ASGI) whilst being backwards compatible. Yet this change is good overall, in that users would be able to add some async usage to their existing codebases. I would also like to note that they can switch to Quart if their usage becomes mostly async. 

This allows for async functions to be passed to the Flask class
instance, for example as a view function,

    @app.route("/")
    async def index():
        return "Async hello"

this comes with a cost though of poorer performance than using the
sync equivalent.


miguelgrinberg:
Hey Phil!

I know my views are different than many on this topic, but the main (only?) reason I would use an asynchronous server over a synchronous one is to achieve much higher scaling. This solution obviously does not address this problem, so this should be noted. I think the note about the performance is warranted, but just a partial aspect, the scaling limitations should be noted as well.

A second thought is that it would be nicer if multiple concurrent async view functions could run concurrently in the asyncio thread, all under the same loop. I believe your implementation would allow only one async function to run at any given time. It seems there wouldn't be any problem in using a shared loop to run all of them concurrently, and that would eliminate some of the performance penalties in using this solution, I think.

pgjones:
Hi Miguel!

I agree with you on both points (except that I think this does allow multiple async functions to run at once if the server is threaded or green threaded - each async function would get its own thread). With the latter point, I'd rather not explore how to implement running all functions under the same thread without a feeling that this is the right approach for Flask.

miguelgrinberg:
Phil, I may be missing something. You are starting a thread pool of size 1 hardcoded, which means that you will have only one asyncio thread per Flask process. Even if you used a threaded server, the bottleneck is the single asyncio thread. You could make that size of 1 configurable, and then you can have some control over how many asyncio view functions can run concurrently, but that seems backwards to me. I think running all of them in a single loop is what makes most sense. I agree with you on that there is no point in going through the effort until we decide to go forward with a solution of this type.

davidism:
I've had individual discussions with @mitsuhiko, @andrewgodwin, and @pgjones that lead to this implementation as a possible first step, despite the limitations you point out. The first goal is to enable `@route() async def` at all. Later improvements could build on this.

Please continue to discuss, I wanted to add some context for why this implementation was submitted.

davidism:
@pgjones if it's helpful, I don't mind depending on asgiref, as they have some nice async/sync helpers already.

src/flask/helpers.py
```diff
@@ -1151,3 +1155,30 @@ def is_ip(value):
...
+            try:
+                result = loop.run_until_complete(func(*args, **kwargs))
+            except Exception as error:
+                call_result.set_exception(error)
+            else:
+                call_result.set_result(result)
+            finally:
+                loop.close()
+
+        loop_executor = ThreadPoolExecutor(max_workers=1)
```
pgjones: @miguelgrinberg is this acting as a global? I had thought this would result in a new executor per call, which in the case of a threaded or gthreaded server would mean multiple executors.

miguelgrinberg: Oh, right! Sorry, I understand now. Yes, you are making a brand new executor each time. So yes, concurrency works, but the expense of running an async view function is fairly high, the executor, the thread and the loop need to be created each time.

lepture:
I suggest we create a v2 branch and putting async into v2.

pgjones:
I'd prefer to avoid a v2 branch, and wait till next year and the end of Py2 before merging this. 

I've looked into asgiref, I'm not sure it is a good fit here - will take another look though, thanks.

PsiACE:
Wow, it's exciting. 
In addition, I think we can stop the suitability guarantee for Python2 in advance.

dwsutherland:
This seems like a mere patch-up job (which @pgjones knows & mentioned)... But I'm sure there's a good reason why Flask isn't just merging-in/taking-on-the-direction-of Quart (?).
As that is more inline with the Python3 `asyncio` direction, and this is a breaking change anyway..

davidism:
@pgjones master is Python 3 only now, this can be rebased.

pgjones:
@davidism I think this is good to review and potentially merge.

src/flask/app.py
```diff
@@ -2318,6 +2320,19 @@ def test_request_context(self, *args, **kwargs):
...
 
+    def ensure_sync(self, func):
+        """Ensure that the returned function is sync and calls the async func.
+
+        .. versionadded:: 2?
+
+        Override if you wish to change how asynchronous functions are
+        run.
+        """
+        if asyncio.iscoroutinefunction(func):
```
pgjones, line 2331: @pgjones Fix this with Python < 3.7

davidism: This might be available in asgiref?

It seems like `asyncio.iscoroutinefunction` is only if you want to support the deprecated `@coroutine` decorated generator functions. Maybe use `inspect.iscoroutinefunction` since we only mention supporting `async def`?

pgjones: I've looked again and I don't think the fix is needed here, it is required in Quart due to the way I wrap functions.

pgjones:
I've updated, and think it is good to merge.

Just to note I see this is the start of async support in Flask rather than a complete solution - there are a number of unanswered questions e.g. what about views that return async iterators?

Leechael:
AnyIO is better choice than asgiref?

magiskboy:
I think it hasn't much impact for wsgi app

docs/async_await.rst
```diff
@@ -0,0 +1,46 @@
...
+the Flask using ``async/await`` based on the ASGI standard (Flask is
+based on the WSGI standard).
+
+
+Decorators
+----------
+
+Decorators designed for Flask, such as those in Flask extensions are
+unlikely to work. This is because the decorator will not await the
+coroutine funcion nor will they themselves be awaitable.
```
killthekitten, line 38:
```suggestion
coroutine function nor will they themselves be awaitable.
```

davidism:
Saw @miguelgrinberg tweet about using an async SQLAlchemy demo: https://twitter.com/miguelgrinberg/status/1279894131976921088, might be relevant to consider with this.

miguelgrinberg:
@davidism So it's still somewhat early to call victory on this, but if you want some details on what I have done...

What Mike Bayer's gist made me realize is that you can do greenlet context switches within an asyncio task, and this allows us to interrupt standard functions and send control back to the asyncio loop. To test this out I added a [custom hub](https://eventlet.net/doc/hubs.html) class to eventlet that does all the waiting using asyncio primitives such as asyncio.add_reader, add_writer, etc.

Since Flask runs great with greenlets, I can now run Flask on top of eventlet, who runs on top of asyncio! Here is some of the crazy things I've tested:

- await for an async function in a normal Flask view function (without blocking, obviously):

```python
@app.route('/foo')
def foo():
    await_(asyncio.sleep(1))
    return 'Hello!'
```

The `await_` helper function uses `asyncio.create_task()` to launch the passed coroutine as a normal asyncio function, and then uses a greenlet based event to wait for it to complete w/o blocking.

- use an async function as a Flask view function by decorating it with the same `await_` helper:

```python
@app.route('/foo')
@await_
async def foo():
    await asyncio.sleep(1)
    return 'Hello!'
```

- await a non-blocking standard function inside an async function. Here is requests.get() with eventlet monkey-patching:

```python
@app.route('/baz')
@await_
async def baz():
    rv = await make_awaitable(requests.get, 'https://google.com')
    print(rv)
    return 'Hello!'
```

The `make_awaitable()` helper function starts a greenlet on the requested function and waits for it to complete using a `asyncio.Future`.

None of this is Flask specific, actually. In fact I haven't had to hack on any official packages, this is all running on official Flask, greenlet and eventlet packages.

What I love about this solution is that all of this works on a single OS thread, without executor tricks. It's truly an asyncio application that just has some "under the table" context switches done via `greenlet.switch()`. And the eventlet monkey-patching allows blocking functions in the Python standard library to run on top of asyncio and become awaitables.

I hope I will have something that people can test and poke holes at soon.

jab:
@miguelgrinberg et al. quoting @oremanj in https://twitter.com/oremanj/status/1279994426538905600
> I packaged this a few months ago as https://github.com/oremanj/greenback — it is indeed a useful tool to have in one’s toolbox!

pgjones:
As I understand the greenlet based solutions they work by switching back to a (prime) coroutine in order to await i.e. every `await_` switches to the prime, `await`s normally and then switches back. This then requires an event loop to run and await the prime coroutine. Running the event loop is the problem that this PR solves using a thread. 

It isn't clear to me then how the greenlet solutions help, as Flask has no event loop available to it and running an event loop in a thread is equivalent to this PR. 

@miguelgrinberg would I be correct in thinking that the Eventlet hub acts as an asyncio event loop substitute? I'm thinking it allows the awaiting of coroutines as if the coroutine was awaited in an asyncio event loop?

miguelgrinberg:
> would I be correct in thinking that the Eventlet hub acts as an asyncio event loop substitute

Not really. The eventlet hub is an asyncio task, running in a 100% asyncio application. It's a single OS thread for eventlet's hub with all of its greenlets plus any normal asyncio stuff.

> Actually reading further, I think you are running Flask within an asyncio loop?

Right. So Flask runs as a greenlet, either under Eventlet's WSGI server, or with the monkey-patched Werzeug dev server. So each request is started on a greenlet. If you call this `await_` helper function then there is a switch to the eventlet hub's greenlet, which is an async function that has full access to await on things.

pgjones:
@miguelgrinberg thanks. I think this PR is the right direction still as it allows the same functionality with any WSGI server rather than the Eventlet WSGI server, or the monkey-patched Werzeug dev server. Thereby allowing users to simply add async routes, rather than change their server. Would you agree?

Also note that Quart allows the same as what you are describing without greenlets, using [this](https://gitlab.com/pgjones/quart/-/blob/master/src/quart/flask_patch/_patch.py#L37) or [nested asyncio](https://github.com/erdewit/nest_asyncio). Another option could be to take this code into Flask and add an ASGI interface to the app.

miguelgrinberg:
> I think this PR is the right direction still as it allows the same functionality with any WSGI server .... Would you agree?

Well, that's hard for me to say. I'm not a fan of this approach, since it does not give you scalability, which is often the reason why people go to asynchronous solutions. If you want to handle 1000 concurrent requests you need to allocate for 1000 threads or processes or a combination of them. A greenlet+asyncio solution would run like a proper async application, with just one OS thread. That to me is a huge benefit.

> it allows the same functionality with any WSGI server rather than the Eventlet WSGI server, or the monkey-patched Werzeug dev server.

This I don't understand. The two servers that I mentioned are the ones that I tested so far, but there are really a lot of options, including gunicorn, waitress, maybe even meinheld. It could also work on uvicorn, and others like it once an ASGI wrapper is built for the `Flask` class. You could run the same application on WSGI or ASGI, using a mix of sync and async handlers. Deep down it is always going to be asyncio+greenlet running the show.

> Also note that Quart allows the same as what you are describing without greenlets

I don't think this is accurate. How would you use the Flask-SQLAlchemy extension with Quart to talk to a MySQL db? Someone needs to write an asyncio version of all of that, which is an impossible task (literally the problem that Mike Bayer is trying to solve that started this odyssey). How would you use an API client that uses the requests library internally? Again, you have to wait for someone to do the work and port the code to async.

Basically what I see here is that there is the potential to build a fairly transparent bridge between sync and async code that no other solution provides. If you don't like the greenlet way of doing async then I'm sure you will disagree with me on this, but out of all the options that have been mentioned in all these years this is the first one I'm excited about to build.

davidism:
It looks like both solutions require wrapping `async def` functions. The wrapper in this PR starts an event loop for the request. The wrapper in @zzzeek's solution uses greenlet to switch contexts.

This PR already started adding an overridable API with the new `Flask.ensure_sync` method. Perhaps this could be renamed to a more generic `wrap_async`. Would that be enough to support the user choosing either solution?

pgjones:
> I'm not a fan of this approach, since it does not give you scalability

I think we have differing aims as my aim is to have the ability to write async/await code in Flask as I'm not sure any proposed solution makes Flask more scalable than it is at the moment.

> If you want to handle 1000 concurrent requests you need to allocate for 1000 threads or processes or a combination of them.

I don't think this should be the case, it was my understanding that asgiref played nicely with gevent/eventlet (with one thread) - testing at the moment suggests it is unstable under race conditions. I think this can be fixed and asgiref running within a gevent patched world is feasible.

> but there are really a lot of options, including gunicorn, waitress, maybe even meinheld.

Is this by starting an event loop and then running gunicorn (etc) within it? Or is there an easier solution I'm missing?

> How would you use the Flask-SQLAlchemy extension with Quart to talk to a MySQL db? 

If I understand correctly if an `await_` can be defined and used then these solutions are possible. In Quart (with the flask patch system) an `await_` is,

```python
def await_(coro):
    return loop.sync_wait(coro)
```

I've created an ASGI interface for Flask [here](https://github.com/pgjones/flask/commit/0338f37e3c3c6186cb1a12ad4308ad58b964fcb7) if it helps testing things.

> Would that be enough to support the user choosing either solution? 

I think it is possible to support either but I think if we answer, "should Flask's async/await support be conditional on the server that runs it?" we would have a preference. My view at the moment is that Flask should support async/await without requiring any changes elsewhere (e.g. server), as I think Django is doing.

davidism:
I'm mostly trying to understand if merging this would *prevent* the greenlet solution (or some other solution, such as using Trio) either as a replacement or in cooperation. I think the overridable wrapper method that's applied to each view allows that.

pgjones:
@davidism I agree that it does not prevent another solution. I think an extension, or otherwise, could [override the Flask class `ensure_sync`](https://github.com/pallets/flask/pull/3412/files#diff-412713dfdbad4d0754cb3693b36a4232R44).

miguelgrinberg:
Agreed. The `ensure_async` wrapper is fine. Other implementations can intercept the async function with a decorator before it reaches Flask, or override the wrapper in a `Flask` subclass.

> I think we have differing aims as my aim is to have the ability to write async/await code in Flask

Yes, I understand. I worry that a lot of people will mistakenly think this is actually a full integration and will be disappointed/file bugs when they find the limitations. I think our differences go back to what Mike Bayer was saying the other day. Python has two parallel universes now, the sync one and the async one, and they only communicate through weird mechanisms such as executors and the like. You seem to be happy with this so you look at best options within these rules. I'm extremely unhappy about this situation and would love to change it.

> I don't think this should be the case, it was my understanding that asgiref played nicely with gevent/eventlet (with one thread created by asgiref) - testing at the moment suggests it is unstable under race conditions. I think this can be fixed and asgiref running within a gevent patched world is feasible.

So we are back talking about greenlets. :)

If there was a made-for-greenlets asgiref then this might be a good solution. Monkey-patching in eventlet/gevent is nice, but definitely not something you want to force, since it is incompatible with lots of things. The fact remains that you still have the two parallel universes taking through that little communication pipe between the sync and the async threads. And the sync side would not be using asyncio at all. In my proposed solution when I issue async I/O through an eventlet file descriptor I'm actually using the asyncio loop, so the two worlds are much closer to each other.

> > but there are really a lot of options, including gunicorn, waitress, maybe even meinheld.
>
> Is this by starting an event loop and then running gunicorn (etc) within it? Or is there an easier solution I'm missing?

Let's assume that I get eventlet fully running on top of asyncio. Now I can use any web server that uses eventlet on the asyncio loop. I can also use any standard web server that works with eventlet monkey-patching. In all cases everything runs on a single asyncio loop w/o any tricks outside of context switching with `greenlet.switch()`. I've tested this with the default select-based loop and also with uvloop and both are quite happy allowing greenlets to do their thing.

> If I understand correctly if an await_ can be defined and used then these solutions are possible.

I'm sure you can find more bridging solutions, but that's not what I'm talking about. Back to the parallel universe idea, I want a merged universe, not a passage between the two worlds.

davidism:
> I worry that a lot of people will mistakenly think this is actually a full integration

I've thought about this too. We're being careful to try to explain that in the docs, and I'll talk about it more when I announce it in a blog post. I think no matter what we add, we'll get some confusion, it's just a really complicated topic, and neither proposal addresses request concurrency right now (beyond running a gevent worker, which isn't new).

As cool and simple as the greenlet solution is, I think I'm more comfortable with the thread+loop solution as a built-in. It would be great to see a Flask-greenback extension, or similar. I'm happy to extend the documentation about this to talk about overriding and point to such extensions.

miguelgrinberg:
> I think no matter what we add, we'll get some confusion

Not so much if this was an extension that people install voluntarily if they want to access this functionality. I actually think an extension is the perfect vehicle for what I'm doing as well, I don't believe any of this belongs in the core project until the day we have a perfect solution. Flask could limit itself to be an "enabler" of async solutions by having the `ensure_async` wrapper, which could actually be a more generic `ensure_view` or something like that that takes anything you give it (not only coroutines) and returns a callable.

> and neither proposal addresses request concurrency right now (beyond running a gevent worker, which isn't new).

I understand your reasoning, but in my opinion this isn't fully accurate. You are implying that the only way to achieve concurrency is to add greenlets in some form. This is 100% true for Phil's solution. The greenlet based solution that I'm testing depends on greenlets for everything, and greenlets do their async business on pure asyncio without tricks. So you can't say that you need to "add" greenlets because greenlets are an integral part of this solution, and because of that scalability is not going to be something you need to decide if you want, you'll always have it whether you need it or not (as it should be, in my view).

miguelgrinberg:
FYI, SQLAlchemy support for asyncio based on greenlets landed on master: https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#asynchronous-io-support-for-core-and-orm.

I have been working on https://github.com/miguelgrinberg/aioflask and have it working under a similar solution, but I still have some more work to do. For example, I'd like to enable the async render_template function for Jinja, and see in what other ways the experience for asyncio developers can be improved. But in any case, the core functionality of having async functions as handlers works really well, and I even managed to port one of the asyncio examples from Quart without having to make any changes to the handler logic.

septatrix:
Is there any recent progress on this topic?

davidism:
Just haven't had time to merge it. Want to review the docs a bit more.

Cloud11665:
if this gets merged, its going to be a game changer

src/flask/helpers.py
```diff
@@ -1122,3 +1123,45 @@ def is_ip(value):
...
+        """This function grabs the current context for the inner function.
+
+        This is similar to the copy_current_xxx_context functions in the
+        ctx module, except it has an async inner.
+        """
+        ctx = None
+        if _request_ctx_stack.top is not None:
+            ctx = _request_ctx_stack.top.copy()
+        elif _app_ctx_stack.top is not None:
+            ctx = _app_ctx_stack.top.copy()
```
davidism, line 1148: `AppContext` doesn't have a `copy` method.

src/flask/helpers.py
```diff
@@ -1122,3 +1123,45 @@ def is_ip(value):
...
+
+    @wraps(func)
+    def outer(*args, **kwargs):
+        """This function grabs the current context for the inner function.
+
+        This is similar to the copy_current_xxx_context functions in the
+        ctx module, except it has an async inner.
+        """
+        ctx = None
+        if _request_ctx_stack.top is not None:
```
davidism, line 1145: When will this be `None`?

davidism:
@pgjones I'm not 100% sure, but I think there's an issue with contexts. Are mutations to `request`, `session`, and `g` visible for the entire request? For example, a `before_request` function sets something that a decorator on the view function looks at, such as Flask-Login. Flask-Login also has its own context local, `current_user`. Each async function copies the request context and runs in a separate thread. Not sure how that will change (for better or worse) with `ContextVar` in pallets/werkzeug#1778.

src/flask/helpers.py
pgjones: Ah, it does in Quart... fixed.

src/flask/helpers.py
pgjones: For the before_first_request functions for example, and potentially others if used directly by users.

pgjones:
I think mutation would be fine, isn't it fine when you use `copy_current_request_context`? I believe the contexts are copied, but the request, session, app, and g, objects referenced by the contexts are the same. (Are you thinking about the copy or something else?)

I don't think the Flask-Login context will work, however the extension could update the helper to copy the context as well.

I don't think ContextVars affect either point here.

recepsirin (approved):

recepsirin (approved):

PsiACE (approved):

pgjones:
Note that @davidism and I have discussed that this implementation must support extensions overriding the `Flask.ensure_sync` method to change the usage throughout. This turns out to be difficult given the Scaffold changes, and the best solution we've found so far results in the slightly unpleasant empty `Blueprint.ensure_sync` method.

src/flask/blueprints.py
```diff
@@ -235,20 +236,40 @@ def register(self, app, options, first_registration=False):
...
                     parent_dict[key].extend(values)
 
             def update(bp_dict, parent_dict):
                 for key, value in bp_dict.items():
                     key = self.name if key is None else f"{self.name}.{key}"
+                    value = defaultdict(
+                        dict,
+                        {
+                            code: {
+                                exc_class: app.ensure_sync(func)
```
davidism, line 253: I know this is only used for the error handler spec, but this change makes the `update` method not generic, so it needs a different name.

pgjones:
Updated docs are look good to me.

pgjones:
🎉

PR #3415, opened 2019-11-04
minor update to converter sample code

shuxianc:
So that it can be copy&pasted then work directly.


davidism:
Thanks, but we don't add this to every example.

PR #3421, opened 2019-11-11
Update documentation to reflect change in Gunicorn 20

bilalshaikh42:
Gunicorn 20.0.0 no longer supports the factory pattern. 
A simple change to the documentation will prevent users from trying this method 

Here is the github issue regarding the change:
https://github.com/benoitc/gunicorn/issues/2159

### DIFF
--- a/docs/deploying/wsgi-standalone.rst
+++ b/docs/deploying/wsgi-standalone.rst
@@ -25,9 +25,10 @@ For example, to run a Flask application with 4 worker processes (``-w
 
 The ``gunicorn`` command expects the names of your application module or
 package and the application instance within the module. If you use the
-application factory pattern, you can pass a call to that::
+application factory pattern, must assign a variable to the application instance, 
+such as "app" and then use the command above.
 
-    $ gunicorn "myproject:create_app()"
+    $ gunicorn "myproject:app"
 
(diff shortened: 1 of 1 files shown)


ThiefMaster:
also, I'd rather see gunicorn add explicit support for calling an app factory.

bilalshaikh42:
> also, I'd rather see gunicorn add explicit support for calling an app factory.

I agree with this, but it seems as if the plan is to deprecate this feature.

bilalshaikh42:
> you shouldn't have an app instance there, I'd rather use `myproject.wsgi:app` instead.

I can't get this pattern to work on my project.  Setting a variable to equal the return of the app factory is working. How does the "wsgi" call fit in here?

ThiefMaster:
you add a wsgi.py that imports and calls your app factory

tilgovi:
I agree with the suggestions. Having a separate module that calls the factory and exposes the result is a good way to do this.

tilgovi:
Looks good to me.

davidism:
Given that benoitc/gunicorn#2178 is slated for 20.0.1 and adds this functionality back, I'm going to close this. Thanks for working on it.

PR #3426, opened 2019-11-14
Fix: update a URL to the newest version

RDIL:
Link now goes to `palletsprojects.com` instead of `pocoo.org`


greyli:
Hi, thanks for working on this. The correct link should be https://flask.palletsprojects.com/en/1.1.x/blueprints/#templates. 

I'll handle all the old links then, closing this for now.

PR #3432, opened 2019-11-18
Document lazy loading and eager loading (fixes #3430)

### LINKED ISSUE #3430: eager loading is not documented
I have been annoyed for a while by Flask's lazy loading feature, without even knowing about the feature or why it was annoying me. (The specific problem: when I have a bug in my app factory that should cause an immediate crash on startup, Flask swallowed the exception and only showed it to me with the first request.)

So this morning, I dug into the code and learned that this is not a bug, but clearly a deliberate feature added back in 2014 ago. I also learned about the workaround for this feature: `flask run --eager-loading`. Perfect! Nothing to fix here, right?

So what's the problem? Documentation! I had to dig into the source code to understand what was going on, and to figure out how make Flask behave the way I want. That should be described in the manual.

### Expected Behavior

In my perfect world, a bug in the app factory would cause an immediate crash on startup. But since delaying such crashes was clearly added deliberately to Flask, and has been there for 5 years, I do NOT propose changing this behaviour.

What should happen: the next person who is annoyed or confused by this behaviour should find an explanation and a workaround in the docs.

```python
import flask

app = flask.Flask(__nam__)                     # <<< TYPO!!!

@app.route('/')
def index():
    return flask.make_response({"foo": 42}, 200)
```

### Actual Behavior

I google'ed, I searched the manual, and I could not figure out why Flask was making it hard to see my bugs. I had to read the source code to find out about `--eager-loading`.

### Environment

* Python version: 3.6, 3.7
* Flask version: 1.1.1
* Werkzeug version: 0.16.0

gward:
See https://github.com/pallets/flask/issues/3430

### DIFF
--- a/docs/errorhandling.rst
+++ b/docs/errorhandling.rst
@@ -19,6 +19,7 @@ errors:
 -   a backend server overloaded
 -   a programming error in a library you are using
 -   network connection of the server to another system failed
+-   you have a syntax error in brand-new code
 
 And that's just a small sample of issues you could be facing.  So how do we
 deal with that sort of problem?  By default if your application runs in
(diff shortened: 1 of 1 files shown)


davidism:
Thanks for starting this. I ended up rewriting the entire doc in #3434.

PR #3434, opened 2019-11-18
lazy load app on reload only

### LINKED ISSUE #3430: eager loading is not documented

### LINKED ISSUE #3431: Only lazy load on reload
The default behavior of the `flask run` command is to lazily import the application if the reloader is enabled. This prevents syntax errors from crashing the app on reload, but it also means that such errors are deferred until the first request.

If the reloader is enabled, use eager loading the first time to show errors immediately. This way, initial errors will be shown to the user when they run the command, but errors introduced later will not crash the server.

davidism:
fixes #3431
fixes #3430

If neither `--eager-loading` or `--lazy-loader` are passed to `flask run`, the default is now to eager load on the first run, then lazy load on reloads. This makes the command fail immediately if there is already an error, and the server remains up if errors are introduced later.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Unreleased
 -   Add :meth:`Config.from_file` to load config using arbitrary file
     loaders, such as ``toml.load`` or ``json.load``.
     :meth:`Config.from_json` is deprecated in favor of this. :pr:`3398`
+-   The ``flask run`` command will only defer errors on reload. Errors
+    present during the initial call will cause the server to exit with
+    the traceback immediately. :issue:`3431`
 
 
 Version 1.1.2
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -108,8 +108,8 @@ Old Version of Flask
 Versions of Flask older than 0.11 use to have different ways to start the
 application.  In short, the :command:`flask` command did not exist, and
 neither did :command:`python -m flask`.  In that case you have two options:
-either upgrade to newer Flask versions or have a look at the :ref:`server`
-docs to see the alternative method for running a server.
+either upgrade to newer Flask versions or have a look at :doc:`/server`
+to see the alternative method for running a server.
 
 Invalid Import Name
 ```````````````````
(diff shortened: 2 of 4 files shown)


davidism:
Going to experiment with this, but if we get rid of the `--lazy-loading` option completely and always let the server fail on initial errors, we can match the semantics of `app.run()` and `flask run` at this point. `flask run` can be a call to `app.run()`, reducing a lot of duplication.

cc @miguelgrinberg

miguelgrinberg:
Very nice improvement, and so simple!

On the removing of code duplication between `flask run` and `app.run()`, I would be super happy if we can also expose the CLI through `app.run()`, along the lines of my PR #2781. So I think I would favor the reverse approach, which is to re-implement `app.run()` to use the `flask run` and CLI logic.

davidism:
Well, that's what I get for staying up too late. We can't put this in `app.run()` because it could never be lazy, since the next reload would have to import `app` again. It has made me think about some other things though, so I might be able to change something down the line.

PR #3437, opened 2019-11-22
Updated Issue Template

RDIL:
* Mentioned Pallets Discord as a way of getting help
* Updated it to use the new format, to fix the warning that shows right now


davidism:
Thanks for looking at this. Since most issues opened are not bugs, and the tracker can be used for feature requests as well, I removed the metadata changes and kept the mention of Discord. The issue templates need a much larger refresh, but that's something I want to think about and coordinate across repos.

PR #3439, opened 2019-12-03
simplify the way we get subdomain and url_prefix at blueprints

Mellcap:
Here is an subtle improvement of simplifying the way we get `subdomain` and `url_prefix` at `blueprints.py`


davidism:
Thanks for looking at this. We generally don't accept these types of PRs as they only shuffle code around rather than improving it significantly. Additionally, `x if condition else y` is preferred over the old `or` pattern.

Mellcap:
Okay, thanks for your reply

PR #3453, opened 2019-12-19
fix docstring and remove redundant parentheses

noviluni:
Hi!

This fixes a docstring and remove redundant parentheses, if you want me to submit as separated PR or you need me to do something else just let me know.


noviluni:
can't understand why are the tests failing... :confused: Are they working?

PR #3456, opened 2019-12-26
Fix a bug that the static endpoint will unexpectedly catch all

### LINKED ISSUE #3452: Catch-all route doesn't work when static_folder ends in forwardslash
I am trying to serve a React app with Flask. The React app uses React Router which means some routes should simply be handled by React. To achieve this one can usually use a catch-all route that returns the index.html. I first couldn't get the catch-all route to work at all, and then I got it working by removing the trailing slash for the `static_folder` parameter when calling Flask.

### Expected Behavior
When one enter a path with trailing forward slash in the `static_folder` parameter for flask.Flask() it shouldn't affect whether a catch-all route goes into effect or not.

In the below code remove the trailing slash for the `static_folder` and you are able to reach the catch-all route (home function) by entering a random route like `/something`.

```python
from flask import Flask, render_template

app = Flask(__name__, static_folder="frontend/build/", template_folder="frontend/build")

@app.route("/")
def route1():
    return "route1"


@app.route("/<path:path>")
def home(path):
    return "routeAny"
```

### Actual Behavior
Steps to reproduce
1. Run the above flask app
2. Open browser and verify that it renders the text "route1" for route `/`
3. Browse to `/something` and see that you get 404 Not Found.

### Environment

* Python version: 3.7.
* Flask version: 1.1.1
* Werkzeug version: 0.16.0

frostming:
Fixes #3452

Strip the ending slash when passed to `os.path.basename`, so that the last part can be used for static url path correctly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 -   Work around an issue when running the ``flask`` command with an
     external debugger on Windows. :issue:`3297`
+-   The static route will not catch all URLs if the ``Flask``
+    ``static_folder`` argument ends with a slash. :issue:`3452`
 
 
 Version 1.1.1
--- a/src/flask/helpers.py
+++ b/src/flask/helpers.py
@@ -1000,6 +1000,8 @@ class _PackageBoundObject(object):
 
     @static_folder.setter
     def static_folder(self, value):
+        if value is not None:
+            value = value.rstrip("/\\")
         self._static_folder = value
 
     @property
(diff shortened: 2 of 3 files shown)


frostming:
`test_main_module_paths` fails on both PR and master branch.
`test_static_url_empty_path_default` runs successfully locally

davidism:
I rebased this to 1.1.x since this seems reasonable as a bug fix there. Another test is failing, can you take a look?

Since I rebased, you'll need to fetch and reset your branch to work on it:

```
git fetch origin
git reset --hard origin/bugfix/3452
```

PR #3458, opened 2019-12-29
Fix app.run() when port is not specified and `SERVER_NAME` contains only host name

### LINKED ISSUE #3457: app.run() doesn't work when port is not specified in `SERVER_NAME`
### Expected Behavior
app.run() should work when port is not specified in `SERVER_NAME`.

```python
from flask import Flask

app = Flask(__name__)
app.config['SERVER_NAME'] = 'example.com'

@app.route('/')
def hello_world():
    return 'Hello, World!'

app.run()
```

### Actual Behavior
```
application fails during start
```

```pytb
  File "mini_svr_2.py", line 10, in <module>
    app.run()
  File "/Users/rickwang/Source/flask/src/flask/app.py", line 973, in run
    port = int(next((p for p in (port, sn_port) if p is not None), _port))
ValueError: invalid literal for int() with base 10: '
```

### Environment

* Python version: Python 3.7
* Flask version: latest 1.1.2.dev0
* Werkzeug version: 0.16.0

rickywxg:
Fix app.run() when port is not specified and `SERVER_NAME` contains only host name
Fixes #3457

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -962,6 +962,7 @@ class Flask(_PackageBoundObject):
 
         _host = "127.0.0.1"
         _port = 5000
+
         server_name = self.config.get("SERVER_NAME")
         sn_host, sn_port = None, None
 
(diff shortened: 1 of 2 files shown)


rickywxg:
Closed per the analysis in issue #3457

rickywxg:
Closing the PR per the analysis in issue #3457

PR #3483, opened 2020-02-06
Update on line 203

itisDouglas:
I've added additional code that has been presented before in the quickstart but I included the escape module because it's not mentioned before and when you try to run the code without it you get various errors due to the escape methods being used in this part of the tutorial.


davidism:
Thanks for taking a look at this. We don't add the app creation to every example because that would get noisy. The assumption is that you're mostly reading the quickstart from top to bottom.

PR #3487, opened 2020-02-07
Pins Werkzeug to 0.16.1 due to breaking change

2ZZ:
Pins Werkzeug to 0.16.1 due to breaking change in 1.0.0 release


ThiefMaster:
why would we want to do that in flask? Flask is compatible with 1.0.0; just some flask extensions aren't - and if they use werkzeug directly, they should have pinned it (or you can pin it in your application)

PR #3488, opened 2020-02-07
Forces specific version of Werkzeug to prevent breaking Flask

### LINKED ISSUE #3486: New version of Werkzeug is not compatible with Flask
### Expected Behavior
A new major release of Werkzeug is available as of yesterday. https://pypi.org/project/Werkzeug/1.0.0/

This is not compatible with flask_admin and flask_wtf:

`  File "/home/airflow/miniconda/envs/conda_env/lib/python3.6/site-packages/flask_admin/model/base.py", line 8, in <module>
    from werkzeug import secure_filename
ImportError: cannot import name 'secure_filename'`

Pinning Werkzeug to version 0.16.1 should solve the issue. https://github.com/pallets/flask/blob/master/setup.py#L56

Flask Version: 1.0.4

DavidHernandez:
Fixes #3486 

Werkzeug released a new major version. Flasks does not protect the Werkzeug dependency for major releases with breaking changes. This should force to use a version of Werkzeug that does not break Flask before adapting the code of Flask to use the new version of Werkzeug.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     include_package_data=True,
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
     install_requires=[
-        "Werkzeug>=0.15",
+        "Werkzeug==0.16",
         "Jinja2>=2.10.1",
         "itsdangerous>=0.24",
         "click>=5.1",


ThiefMaster:
The werkzeug update does not break flask but just flask extensions that have not been updated since the deprecation of some top-level imports 5 months ago. Please pin it in your application if there aren't updated versions of the extensions you're using.

gricey432:
@ThiefMaster that's not the case, see this documented property which no longer works
https://flask.palletsprojects.com/en/0.12.x/api/#flask.Request.is_xhr

ThiefMaster:
You are right, I hadn't realized that older versions also pull in the latest Werkzeug.

gricey432:
This PR should be for a 0.12.x release

gricey432:
Also see this line in Flask itself which now throws an exception
https://github.com/pallets/flask/blob/0.12.x/flask/json.py#L298
because https://github.com/pallets/flask/blob/0.12.x/setup.py#L74 doesn't specify an upper limit on werkzeug

gricey432:
I've cleaned up these with some reproducible examples in #3493

PR #2332, opened 2017-05-25
PyCharm Run Configuration Guide

doobeh:
A guide showing how to get the `flask run` cli functionality directly within PyCharm-- addresses issue #2198. Even when PyCharm implements smoother integration, we're probably still going to need a variation of this to address factories/etc.

Issue #2198, opened 2017-03-04
Add instructions for running Flask with PyCharm


davidism:
With the introduction of the `flask` command, setting up PyCharm is not as straightforward. Some screenshots and instructions would be useful. Similar to the section on how to deploy to common platforms.

ThiefMaster:
Maybe PyCharm should have some shortcut to run an app tthrough the `flask` cli since they already have claim to support flask?

pmayer:
I added a feature request in YouTrack: https://youtrack.jetbrains.com/issue/PY-23190

davidism:
While we now document this in Flask, I encourage anyone who comes across this to go vote for the linked PyCharm issue. You'll need to log in, then there's a small checkmark at the bottom of the right-hand sidebar (the star is not the same as the vote button).

PR #3507, opened 2020-02-18
Remove outdated Pycharm instructions

jeffwidman:
This was originally added before Pycharm natively supported this.

Pycharm 2018.1 fixed that issue.

So these instructions are now more than two years out of date so they
are a potential source of confusion rather than helpful for n00bs.

Further context:
* https://github.com/pallets/flask/issues/2198
* https://github.com/pallets/flask/pull/2332
* https://youtrack.jetbrains.com/issue/PY-23190


davidism:
I think the Flask run configuration is only available in PyCharm Professional, so these might still be relevant for Community users. Instructions might need to be updated though.

jeffwidman:
Ah, good thought. I searched a bit but didn't find anything on Jetbrains site stating suppport was/wasn't included... I am not a Pycharm user myself, and the install is pretty heavy-weight so perhaps someone else who already has pycharm CE installed could chime in? If this doesn't get traction in a week I'll try to circle back and test it myself.

RemiDesgrange:
There is no Flask run config in pycharm community. So you could say "Pycharm CE integration" ?

davidism:
Updated the docs to mention PyCharm Professional vs. Community Edition: c6a619af

jeffwidman:
Thanks for handling this!

PR #3510, opened 2020-02-20
Change in return part

Kunal614:


tests/test_templating.py
```diff
@@ -14,7 +14,7 @@
 import werkzeug.serving
 from jinja2 import TemplateNotFound
 
-import flask
+import flask , render_template ,render_template_string
```
davidism:
```python
from flask import render_template
from flask import render_template_string
```

PR #3513, opened 2020-02-25
Make code more easy to read

kevinanew:
When I read the code, I feel something is hard to understand.
So I change it.

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -960,17 +960,24 @@ class Flask(_PackageBoundObject):
         if debug is not None:
             self.debug = bool(debug)
 
-        _host = "127.0.0.1"
-        _port = 5000
         server_name = self.config.get("SERVER_NAME")
-        sn_host, sn_port = None, None
+        sn_host = sn_port = None
 
         if server_name:
             sn_host, _, sn_port = server_name.partition(":")
 
-        host = host or sn_host or _host
-        # pick the first value that's not None (0 is allowed)
-        port = int(next((p for p in (port, sn_port) if p is not None), _port))
+        if not host:
+            if sn_host:
+                host = sn_host
+            else:
+                host = "127.0.0.1"
+
+        if port or port == 0:
+            port = int(port)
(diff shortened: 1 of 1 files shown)


src/flask/app.py
```diff
@@ -1256,11 +1260,11 @@ def index():
             )
 
         if provide_automatic_options is None:
-            if "OPTIONS" not in methods:
+            if "OPTIONS" in methods:
+                provide_automatic_options = False
+            else:
```
sukurcf: why not 
`provide_automatic_options = False if "OPTIONS" in methods else True`

davidism: Because there's other stuff going on in the `else` block.

ThiefMaster: Absolutely not. First of all, this is just adding more complexity than needed as your code is just a more convoluted way of this:

    provide_automatic_options = "OPTIONS" not in methods

But then, if you look a the whole code, you'll notice that the code in the else branch does more than just setting the variable to True...

src/flask/app.py
```diff
@@ -2146,11 +2150,11 @@ def create_url_adapter(self, request):
...
-            subdomain = (
-                (self.url_map.default_subdomain or None)
-                if not self.subdomain_matching
-                else None
-            )
+            if self.subdomain_matching:
+                subdomain = None
+            else:
+                subdomain = self.url_map.default_subdomain or None
+
```
sukurcf: Why not
`subdomain = None if self.subdomain_matching else self.url_map.default_subdomain`

ThiefMaster: That would not convert `''` to `None` as the current code does.

kevinanew: @sukurcf Because it's will too long (more than 78 characters), that's why.

davidism:
I like where you've expanded the code into full `if/else` statements. However, I'm going to back out the wrapping changes, as they don't really match what this PR is about and aren't really consistent.

PR #3524, opened 2020-03-02
Remove deprecated `license_file` option

jeffwidman:
Starting with wheel 0.32.0 (2018-09-29), the `license_file` option is
deprecated in favor of `license_files`.

https://wheel.readthedocs.io/en/stable/news.html

The wheel will continue to include `LICENSE`, it is now included
automatically:

https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file

See https://github.com/FactoryBoy/factory_boy/pull/696 for inspiration and more discussion.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[metadata]
-license_file = LICENSE.rst
-
 [bdist_wheel]
 universal = true
 


jeffwidman:
This may also be applicable to other projects in Pallets, but I thought best to first discuss here before looking into those.

davidism:
Can you confirm that the dist files produced by `python setup.py sdist bdist_wheel` still contain the license?

jeffwidman:
Good question, I hadn't actually done that:

```
$ python setup.py sdist bdist_wheel
...
running install_scripts
adding license file "LICENSE.rst" (matched pattern "LICEN[CS]E*")
...
adding 'Flask-1.2.0.dev0.dist-info/LICENSE.rst'
...
# made a test venv to install Flask into
$ pip install dist/Flask-1.2.0.dev0-py2.py3-none-any.whl
$ tree /Users/jeffwidman/.virtualenvs/test/lib/python3.7/site-packages/Flask-1.2.0.dev0.dist-info
/Users/jeffwidman/.virtualenvs/test/lib/python3.7/site-packages/Flask-1.2.0.dev0.dist-info
├── INSTALLER
├── LICENSE.rst
├── METADATA
├── RECORD
├── WHEEL
├── entry_points.txt
└── top_level.txt

0 directories, 7 files
```

So it looks like the wheel picks it up as expected, but the tar file doesn't... unless I need to somehow install the tarred file for it to be copied to the correct place?

davidism:
Not sure why you got those results, but I just confirmed that the license is present in both archives as well as in dist-info when installed.

jeffwidman:
Weird, I specifically checked the tar files in addition to the output pasted above... maybe I started with the wrong archive somehow? But hey, if it works for you, sounds good to me. 😀

PR #3537, opened 2020-03-18
Update deprecated for config.from_json

shanavas786:


shanavas786:
failures seems to be unrelated.

PR #3541, opened 2020-03-20
Clarify that static_folder must be a relative path

jeenuv:
Also hint at `root_path` parameter rather than mentioning in plain text.


davidism:
Thanks! Changed base to 1.1.x, updated to mention it could also be an absolute path, and to fix some formatting.

PR #3548, opened 2020-03-28
Missing slash in app route

edupo:
The missing `/` in the `@app.route('/hello')` statement makes the example raise a 404 when accessing the page. To fix it I just added it to the `route`.


davidism:
Thanks for looking at this. However, this change isn't required. `/hello` is different than `/hello/`. Since this route is more conceptually a "leaf" node, it doesn't have a trailing slash.

Issue #3108, opened 2019-02-28
config option for "flask run" (e.g. --dotenv/-e)


jab:
Would it make sense to allow passing a `--config/-c` option to `flask run`, that would then be exposed to users somehow for subsequent passing to e.g. `app.config.from_pyfile()`?

(I'm imagining this done in a reusable way so that a "flask gunicorn" custom command (the likes of which I've already implemented) could accept and expose a `-c` value similarly, so that the app could be configured from this same provided path whether running in development or production.)

This would reduce the dependence on an environment variable to specify this. Reason I ask is that many of the Flask users I'm supporting are allergic to environment variables for the reasons mentioned in #3095, and have been missing some way to pass a CLI option to `flask run` to help configure their apps. Thanks 😊

davidism:
I think we might have to disagree about env vars to some degree. We did add support for python-dotenv to read a `.flaskenv` file so that you don't have to export the vars in your terminal, maybe that mitigates some of your concerns.

The issue I see with adding a config option is that it now dictates one way to configure the app that's not portable to how you'd be deploying to production. If users use this and then use mod_wsgi, for example, they'd have to read the config themselves anyway. It lends preference to `from_pyfile` as opposed to other functions in Flask or patterns in tutorials. It wouldn't solve the issue with `FLASK_APP` and `FLASK_ENV`, which aren't set from config.

I think you might be looking for a custom run command that's more appropriate to your (and probably others) use cases. That might be a good fit for an extension that you'd then install for each of your projects. The docs on Flask's CLI document how extensions can register commands automatically. For example, you could register a `run` command that made assumptions about where the app is, and what its env and config are.

jab:
Thanks so much for the helpful reply, @davidism!

I decided to embrace environment variables after all, in a very particular way. After deciding on dotenv as a configuration storage format, and finding Flask CLI's built-in dotenv support only able to get us part of the way there, here's what I did:

I ended up subclassing `FlaskGroup` to add support for passing (one or more) `-e/--dotenv` paths as a Flask CLI option, which my lightly-customized `flask` CLI sources into the environment before then loading a user's app. This allows a user to configure and start their app with a command like `path/to/my/project/bin/flask -e qa.env run`, instead of something like `cd path/to/my/project; set -a; source qa.env; ./bin/flask run` (no more $PWD sensitivity!).

[...]

PR #3560, opened 2020-04-07
Stop change CWD to .env/.flaskenv location

greyli:
By [the current implementation](https://github.com/pallets/flask/blob/master/src/flask/cli.py#L630), if `.env` or `.flaskenv` was found in the top-level directory, Flask will change the current work directory to the directory that contains the `.env` or `.flaskenv` file. 

When the user accidentally put a `.env` or `.flaskenv` in the top-level directory, then executing `flask run` in the current directory (contains `app.py`) will throw out NoAppException.

See more in #3561.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,9 @@ Unreleased
     instead of PyOpenSSL. :pr:`3492`
 -   When specifying a factory function with ``FLASK_APP``, keyword
     argument can be passed. :issue:`3553`
+-   When loading a ``.env`` or ``.flaskenv`` file on top level directory,
+    Flask will not change current work directory to the location of dotenv
+    files, in order to prevent potential confusion. :pr:`3560`
 
 
 Version 1.1.2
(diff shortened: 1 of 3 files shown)


davidism:
@jab any thoughts on this, especially in relation to #3108? I think your projects had setups that were affected by the cd.

jab:
Thanks for the ping on this, @davidism. I can confirm these changes don't break anything in my customized Flask CLI (since it does some more sophisticated detection of .env files that is independent of Flask's built-in detection, which I can explain more in #3108 or elsewhere).

And these changes seem good to me otherwise. Silently changing the CWD based on the presence/location of (hidden) .env files seems like it can break things in a hard-to-debug way for users. I may be missing something, since I'm not clear on what the benefit was in the first place. I'd expect users needing to compute paths relative to their app's modules should be doing so in a CWD-independent way (e.g. using  `__file__`). If they're not, I guess this change could break them. But even if it does, it seems the benefits here outweigh that cost.

The only thing missing before I think this can be merged is an update to the changelog.

jab:
This looks good to me, and no one's raised any concerns, so I'm going to merge this before it stalls any longer.

PR #3562, opened 2020-04-07
Remove simplejson and deprecate encoding options

### LINKED ISSUE #3555: Remove simplejson
In modern Python it's unlikely to be significantly better than the built-in `json`. The module used by `JSONMixin` is overridable, so users can plug it in again if they want.

See pallets/itsdangerous#146 and pallets/werkzeug#1766.

davidism:
* Remove simplejson. Remove UTF encoding guessing, which the built-in module handles now. Simplify `htmlsafe_dumps` to use `str.translate` instead of multiple `str.replace`.
* Issue deprecation warning if `encoding` is passed to `dumps`, `dump`, `loads`, or `load`. This was deprecated in the built-in module back in 3.1. The only valid encodings for JSON are UTF-8 (with/without BOM), UTF-16 (BE/LE), and UTF-32 (BE/LE). `json.loads` accepts bytes and `json.loads` accepts binary files in these encodings without any special handling on our part. `dumps` can be encoded after calling, and `dump` can be passed a binary file wrapped with `io.TextIOWrapper`.
* Rewrite docs for consistency.
* Fix imports and references of `Markup` to be from MarkupSafe instead of Flask or Jinja.

See the individual commits for easier diffs.

closes #3555

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ Version 2.0.0
 Unreleased
 
 -   Drop support for Python 2 and 3.5.
+-   JSON support no longer uses simplejson. To use another JSON module,
+    override ``app.json_encoder`` and ``json_decoder``. :issue:`3555`
+-   The ``encoding`` option to JSON functions is deprecated. :pr:`3562`
 -   Add :meth:`sessions.SessionInterface.get_cookie_name` to allow
     setting the session cookie name dynamically. :pr:`3369`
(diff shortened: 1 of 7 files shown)


anmol-dhingra:
@davidism I see we are no-longer importing json from itsdangerous library. When is the fix for removal of simplejson planned to be released as flask import is failing because of the simplejson raising an error.

PR #3564, opened 2020-04-08
use ast to parse FLASK_APP

### LINKED ISSUE #3553: Use ast.parse to validate FLASK_APP
I wrote this for Gunicorn a while ago: https://github.com/benoitc/gunicorn/pull/2178. It would be good to use it for Flask too, it's the same use case. Right now we use some regex and string manipulation to validate and prepare for `ast.literal_eval`, but `ast.parse` is the correct way to do it.

davidism:
This had the nice side effect of parsing keyword arguments too. Refactored `call_factory` to use `inspect.signature`, even though it will go away in 2.1 because it was easier to figure out the args and kwargs checks.

closes #3553

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@ Unreleased
     200 OK and an empty file. :issue:`3358`
 -   When using ad-hoc certificates, check for the cryptography library
     instead of PyOpenSSL. :pr:`3492`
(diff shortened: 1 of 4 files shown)


jeffwidman:
Very nice, I didn't realize stdlib has an `ast` module.

PR #3567, opened 2020-04-10
Fix flake8 repo url in .pre-commit-config.yaml

Ma233:
When using git version `1.8.3.1`, users cannot use .git-less remote URLs for cloning on gitlab, which will cause failure on pre-commit tool initialization.

Reproduce environment:
CentOS Linux release 7.6.1810 (Core)
git version 1.8.3.1

Logs while running `pre-commit run --all-file` in a newly cloned flask project:
```
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    error: RPC failed; result=22, HTTP code = 404
    fatal: The remote end hung up unexpectedly

Check the log at /home/magine/.cache/pre-commit/pre-commit.log
```

See also:
https://gitlab.com/gitlab-org/gitlab/issues/29629

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
-  - repo: https://gitlab.com/pycqa/flake8
+  - repo: https://gitlab.com/pycqa/flake8.git
     rev: 3.7.9
     hooks:
       - id: flake8


davidism:
[Git 1.8..3.1 is from 2013.](https://github.com/git/git/releases/tag/v1.8.3.1) We're not supporting [*Python* versions](https://devguide.python.org/#status-of-python-branches) that are that old. Is there no way for you to use a more modern version? Flake8 itself closed [an issue about this](https://gitlab.com/pycqa/flake8/-/merge_requests/381).

davidism:
CentOS 8 has a much more modern version. Perhaps it's time to upgrade your dev machine? :-)

Ma233:
@davidism Yes, my bad. I just googled my error message, but forgot to check if the version is newest.

PR #3581, opened 2020-04-22
Another missing import and instantiation

hellowesley:

Missing another Flask module import and app variable instatiation


davidism:
Thanks, but I'm going to close this. I'm sure there's been discussions on this tracker previously about this, but I deliberately don't add boilerplate that's already been shown earlier in the doc because it just gets noisy and is not relevant to the specific thing being shown.

PR #3582, opened 2020-04-23
Add support for gzip compression on json outputs

danilobrinu:
This PR add support for gzip compression on json outputs for large responses compression. Instead of modify the current method for json outputs `flask.json.jsonify`, I created new method called `flask.json.gzonify` with the simple purpose of optimize the size and time of the response for large json responses.

Note: You can find a demo on: https://repl.it/@danilobrinu/flask-jsonify-support-compresion, just click on `Run` button and access to the respective endpoints.  https://honestweepysoftwareagent.danilobrinu.repl.co/without-compression and https://honestweepysoftwareagent.danilobrinu.repl.co/with-compression . Don't forget open the **devtools** and go to the **Network Tab** to select the **Fast 3G** option.

**Without Compression**
image

**With Compression**
image

>  **Network - Fast 3G**: We can optimize the size and time: From 160kb and 1.42s to 540B and 176ms

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,8 @@ Unreleased
     instead of PyOpenSSL. :pr:`3492`
 -   When specifying a factory function with ``FLASK_APP``, keyword
     argument can be passed. :issue:`3553`
-
+-   Add :func:`~flask.json.gzonify` to support gzip compression for better web
+    performance (less size and time). Works like :func:`~flask.json.jsonify`.
 
 Version 1.1.2
 -------------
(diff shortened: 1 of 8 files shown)


davidism:
I'm sorry to do this after you put in so much work, but I don't think this belongs in Flask core. Content encoding is something the [WSGI server]() or [HTTP server](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) should apply, or can be applied generally as a middleware without the framework needing to do it. There's no reason this should apply just to JSON, and the implementation ends up being mostly a copy of `jsonify`.

This might be a better fit as an extension middleware. In fact, a quick search turned up a few, such as https://github.com/shengulong/flask-compress, https://github.com/closeio/Flask-gzip, and https://github.com/cralston0/gzip-encoding.

PR #3587, opened 2020-04-24
Add support for IPython shell when using `flask shell`

jacobstoehr:
When running `flask run -i ipython`, drop the user into an IPython shell with
the `app` instance imported and the context active just like in the regular
shell.

This makes magic methods as well as native autocomplete available to the user.

Update the docs to reflect this change

Commit checklist:

* add tests that fail without the patch
The function was untested before, I dont really know a good way to test this.
* ensure all tests pass with ``pytest``
the tests fail on master as well (on my machine).
* add documentation to the relevant docstrings or pages
Done
* add ``versionadded`` or ``versionchanged`` directives to relevant docstrings
How do I do this when the feature might only be in master at some point? I cant predict the next version number :)
* add a changelog entry if this patch changes code
Done

Tests, coverage, and docs will be run automatically when you submit the pull
request, but running them yourself can save time.

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -863,9 +863,19 @@ def run_command(
     )
 
 
+shell_choices = ("python", "ipython")
+
+
 @click.command("shell", short_help="Run a shell in the app context.")
+@click.option(
+    "-i",
+    "--interface",
+    type=click.Choice(shell_choices),
+    default="python",
+    help="Which interactive shell to use.",
+)
 @with_appcontext
-def shell_command():
+def shell_command(interface):
     """Run an interactive Python shell in the context of a given
     Flask application.  The application will populate the default
     namespace of this shell according to it's configuration.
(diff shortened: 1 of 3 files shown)


src/flask/cli.py
```diff
@@ -863,37 +863,66 @@ def run_command(
...
+    ctx.update(app.make_shell_context())
 
-    # Support the regular Python interpreter startup script if someone
-    # is using it.
-    startup = os.environ.get("PYTHONSTARTUP")
-    if startup and os.path.isfile(startup):
-        with open(startup) as f:
-            eval(compile(f.read(), startup, "exec"), ctx)
+    if interface == shell_choices[0]:
+        banner = (
```
jacobstoehr, line 893: I decided to split up the banner variable into 2 different strings, to be able to also display the IPython version easily. If you think that is not necessary this could be simplified.

ThiefMaster:
Thanks, but there's already [flask-shell-ipython](https://pypi.org/project/flask-shell-ipython/) which works fine as far as I know.

jacobstoehr:
I am aware of the package, however I though it would be nice if that behavior was not an external dependency.

ThiefMaster:
The problem is that this adds more maintenance work for us - and at some point someone will want e.g. bypthon or other REPLs added as well.

jacobstoehr:
Ok, I see, that makes sense to have it as an external dep then. Thank you for taking the time to expain it!

Issue #3599, opened 2020-05-06
Error when apply tojson or flask.json.dumps on dict with None as key


XD-DENG:
#### 1. Example Application

- `app.py`

```python
from flask import Flask, render_template

app = Flask(__name__)


@app.route("/")
def hello_world():
    data = {"X": 1,
            "Y": True,
            None: "abc"}

    return render_template("index.html", data=data)


if __name__ == "__main__":
    app.run(debug=True)

```

-  `templates/index.html`

```html
<html>

<script>
 const a = {{ data|tojson }};
</script>

<p>Hi</p>

</html>
```

#### 2. Example Single-Line Code

The same issue will be triggered if you run code below

```python
flask.json.dumps({None: 1, "X": 2})
```

### Actual Behavior

```pytb
Traceback (most recent call last):
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/XD/Library/Python/3.7/lib/python/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
[...]

PR #3600, opened 2020-05-06
Fix issue arising when tojson or flask.json.dumps is applied on dict with key None

XD-DENG:
This PR aims for fixing #3599


davidism:
This is an issue with the data, not Flask. JSON should only have string keys. I don't think this is something Flask should allow.

PR #3601, opened 2020-05-07
Fix function name in example

jjconti:
The function returned by the decorator should be called "wrapper_view" instead of "wrapped_view" because it wraps the original view. It's not wrapped by it.

The example in Python docs use the same wording: https://docs.python.org/2/library/functools.html#functools.wraps


jjconti:
Maybe `view_wrapper` is even a better name.

davidism:
Thanks for looking at this, but I don't think this is significant enough to warrant a change.

PR #3605, opened 2020-05-09
add support to override commands

aprilahijriyan:
This PR is to answer requests about how to override the default command, on the flask command. seen in issue #2020.


davidism:
The linked issue has been closed for some time, it was  not available for a PR. You also left the large comment from right above that code, that explains why it's currently the way it is.

Issue #3608, opened 2020-05-12
Blueprint cli commands cannot provide help for cli groups


Sebi2020:
### Expected Behavior
There should be a way to specify blueprint command group help.
```python
from flask import Blueprint
editUser = Blueprint("editUser",__name__) # cannot add help to this entry (flask --help)

# Will be ignored:
editUser.__doc__ = "maybe this adds help...."

@editUser.cli.command("seed")
def seed():
    """ Seeds user database (will be shown if you type flask editUser --help)"""
    pass
```

### Actual Behavior
Currently flask does not provide a way to display help for blueprint commands.

*`flask --help` Output:*
```
Options:
  --version  Show the flask version
  --help     Show this message and exit.
Commands:
  editUser 
  routes    Show the routes for the app.
  run       Run a development server.
  shell     Run a shell in the app context.
```

### Environment

* Python version: 3.6.9
* Flask version: 1.1.2
* Werkzeug version: 1.0.1

### Suggestion 
Add a cli_group_help keyword to blueprints:
```
editUser = Blueprint("editUser",__name__, cli_group_help="Provides user management commands")
```

davidism:
`bp.cli` is a Click `Group`, assign to `bp.cli.short_help` to set its help text in the list.

PR #3609, opened 2020-05-13
Add cli_group_help parameter for Blueprint

greyli:
This PR will add a `cli_group_help` parameter for the blueprint. When register blueprint level commands, you can use it to set command group help text.

If not set, the default help text for the blueprint command group will be `A command group for <blueprint's name> blueprint.`.

ref: https://github.com/pallets/flask/issues/3608

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,8 @@ Unreleased
     instead of PyOpenSSL. :pr:`3492`
 -   When specifying a factory function with ``FLASK_APP``, keyword
     argument can be passed. :issue:`3553`
+-   Add ``cli_group_help`` parameter for blueprint when register
+    blueprint level commands.
 
 
 Version 1.1.2
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -362,6 +362,20 @@ commands directly to the application's level:
 
     $ flask create alice
 
+You can set the group help text with the ``cli_group_help`` parameter
+when creating the :class:`Blueprint` object, or later with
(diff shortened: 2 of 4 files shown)


davidism:
I was going to suggest assigning to `bp.cli.help`, I don't think we need a separate parameter for it. Thanks for working on this though.

Sebi2020:
> I was going to suggest assigning to `bp.cli.help`, I don't think we need a separate parameter for it. Thanks for working on this though.

That should be documented! There is no python docstring for bp.cli.help nor can you find any information about this attribute in the official documentation.

davidism:
It's not documented in Flask because it's an attribute of an object from Click.

Sebi2020:
> It's not documented in Flask because it's an attribute of an object from Click.

That may be the case, but you expose the click attribute through the blueprint class and the user does not know that `bp.cli.help` belongs to `Click`.

greyli:
Yes, use `bp.cli.help` will be more simple, then we only need to update the documentation. 

Maybe it's better to set a default value for `bp.cli.help`? Something like `A command group for <blueprint's name> blueprint.`.

davidism:
I don't want to update the docs with this because it's no different than any other attribute on any other object from a separate library. There's no docs about setting help just like there's no docs about using `click.progressbar`, or most other things you *could* do. I think it's already well established that Click is used for the CLI.

davidism:
Setting some default help _might_ be useful, but it's really just repeating the name of the command (the bp name) and saying it's a group.

PR #3622, opened 2020-05-25
add function REGEXP

cjql:
**add function REGEXP for sqlite3 which can support regex search.**

When using such sql 
```sql
SELECT word FROM words WHERE word REGEXP ".*go"
````
the following error will happen
```py
sqlite3.OperationalError: no such function: REGEXP
```
and this PR can fix the error.

References:
https://www.sqlite.org/lang_expr.html#regexp
https://stackoverflow.com/questions/5365451/problem-with-regexp-python-and-sqlite

### DIFF
--- a/examples/tutorial/flaskr/db.py
+++ b/examples/tutorial/flaskr/db.py
@@ -1,3 +1,4 @@
+import re
 import sqlite3
 
 import click
@@ -5,6 +6,9 @@ from flask import current_app
 from flask import g
 from flask.cli import with_appcontext
 
+def regexp(expr, item):
+    reg = re.compile(expr)
+    return reg.search(item) is not None
 
 def get_db():
     """Connect to the application's configured database. The connection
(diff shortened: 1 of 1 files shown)


cjql:
This is my first PR in GitHub. And it seems only the black style fails. 

Can someone help me with this?

Thank you.

davidism:
Thanks, however this doesn't seem to be relevant to the tutorial. You're welcome to add this to your own project if you need it.

cjql:
> Thanks, however this doesn't seem to be relevant to the tutorial. You're welcome to add this to your own project if you need it.

Thanks for your reply. Sqlite supports regex search but flask sqlite does't, by default, as I know. Maybe this will be helpful. Or if there were other simpler way for regex search，please tell me. Thank you.

davidism:
Flask has nothing to do with SQLite beyond the fact that you can use it as a database. The fact that you're using Flask has nothing to do with not having a regex function registered with SQLite. Regex search is not part of the Flask tutorial.

Issue #3612, opened 2020-05-16
Discuss terminal copy/paste issues on security page


davidism:
The [security](https://flask.palletsprojects.com/en/1.1.x/security/) page in the docs discusses various security issues that Flask might not prevent. Add a section discussing things that could happen when copying and pasting code into a terminal. The discussion doesn't need to be exhaustive. Mainly focus on "if you're allowing users to enter code, such as comments on a programming blog, and expect other users to copy it, you might want to apply extra filtering, such as `body.replace("\b", "")`."

See pallets/markupsafe#71, which discussed making `markupsafe.escape` also escape backspace characters. I concluded that it was better for terminals to handle it than for MarkupSafe to expand its scope beyond HTML syntax.

The main issue is the backspace (`\b`, `^H`) character. It is not rendered by browsers, but it does affect text in a terminal. So `import y\bose\bm\bi\bt\be\b ` renders as `import yosemite` but becomes `import os` when pasted in a terminal. Most modern terminals detect invisible control characters and offer to remove them when pasting, but some terminals may still have this issue.

Other issues include newlines causing multi-line blocks to execute automatically, invalid characters being interpreted together as the escape key, commands hidden after other commands with `;`, triggering Vim commands, etc. There are lots of ways to craft commands that look familiar or innocuous and have them cause unexpected behavior.

Another discussion I found was https://security.stackexchange.com/questions/39118/how-can-i-protect-myself-from-this-kind-of-clipboard-abuse.

cyberShaw:
Can I work on this issue?
I'm new to Open Source and interested to contribute 😄

davidism:
No need to ask, if an issue doesn't have an open PR or someone assigned to it, anyone can work on it at any time.

I will say that for more prose-focused docs like this would be, I want the language, grammar, and style to match other sections I've written, so be sure to get a feel for how the other docs are written.

PR #3633, opened 2020-05-31
security.rst: Add section on copy/paste security

kx-chen:
Closes https://github.com/pallets/flask/issues/3612

Adds a section on copy and pasting security.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -258,3 +258,22 @@ certificate key to prevent MITM attacks.
    or upgrade your key incorrectly.
 
 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Public_Key_Pinning
+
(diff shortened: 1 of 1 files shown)


kx-chen:
Hi! Not too sure if this commit is supposed to go into `1.1.x` or `master`. It seems like the contributing [docs](https://flask.palletsprojects.com/en/1.1.x/contributing/) suggest it should be `1.1.x`?

davidism:
Thanks for working on this, this looks like a good start to what I had in mind.

Docs usually go on the latest version branch so that they show up in the default docs build. It looks like you branched off master but tried to open the PR against 1.1.x, which won't work. I rebased to 1.1.x with the following commands:

```
$ git rebase --onto origin/1.1.x origin/master
$ git push -f
```

PR #3643, opened 2020-06-08
simplify and modernize some stuff for the tutorial project

Laurentiu-Andronache:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ htmlcov/
 .coverage
 .coverage.*
 *,cover
+coverage.xml
--- a/examples/tutorial/.gitignore
+++ b/examples/tutorial/.gitignore
@@ -5,6 +5,7 @@ instance/
 .cache/
 .pytest_cache/
 .coverage
+coverage.xml
 htmlcov/
 dist/
 build/
--- a/examples/tutorial/README.rst
+++ b/examples/tutorial/README.rst
@@ -67,7 +67,7 @@ Test
 
 ::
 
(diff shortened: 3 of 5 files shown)


ThiefMaster:
why are you removing the single quotes? they work everywhere and the chance that developers use a decent shell (ie zsh) is rather high...

Laurentiu-Andronache:
@ThiefMaster, for example they give an error in Windows. The test stuff ain't getting installed. And they're not needed in bash, which gets more usage than zsh.

ThiefMaster:
They may not be needed in bash, but they work fine there. And while bash is the default shell everywhere and thus the most commonly installed one, I don't think that's necessarily the case when counting only developer systems.

Maybe double quotes work on windows as well? In any case, Windows is probably the least common OS for web development out there...

Laurentiu-Andronache:
Ok Maybe don't merge that commit then. Do you like the others though?

davidism:
Thanks for looking at this. However, I don't think any of these changes are necessary or better, just different. Going to close this for now.

Laurentiu-Andronache:
OK. But in the HTML code (and in the CSS as well), there's a lot of poor code (like multiple <h1> tags on the same page), non-responsive on mobile, etc. I'll submit a pull request tomorrow, and maybe you reconsider the commits that I did here as well when the time comes.

davidism:
We don't control the HTML or CSS (for the most part), that's rendered by Sphinx. Multiple H1 tags are not poor code, the HTML5 outline algorithm has allowed that for almost a decade.

PR #3652, opened 2020-06-15
Use valid HTML in base template

vpoulailleau:
Add `html`, `head`, `body` tags.


ThiefMaster:
Both `<html>` and `<head>` are optional in HTML5. Your PR also adds inconsistent indentation (4 vs 2 space).

vpoulailleau:
Whoops. Sorry.

Extra information here: https://stackoverflow.com/a/25749523/12199445, Google recommends to remove all optional tags in their style guide.

PR #3662, opened 2020-06-21
Removed misleading DEBUG variable from examples

lgiordani:

The configuragion documentation states that "While it is possible to set :data:`ENV` and :data:`DEBUG` in your config or code, this is strongly discouraged.` So, I believe using the `DEBUG` setting in the examples is misleading, as people are led to believe that that is needed to activate the debug mode. Indeed, basically every tutorial on Flask that you find around has the `DEBUG` variable in the configuration. I propose we get rid of all of them.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -555,7 +555,7 @@ configuration::
     import os
 
     basedir = os.path.abspath(os.path.dirname(__file__))
-    
+
     class Config(object):
         TESTING = False
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -552,17 +552,21 @@ and import different hard-coded files based on that.
 An interesting pattern is also to use classes and inheritance for
 configuration::
 
+    import os
+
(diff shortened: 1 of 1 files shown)


jab:
Yeah, it’s better to demo Flask inferring DEBUG based on FLASK_ENV, as these docs mention earlier. Thanks for the PR. 

While you’re in here: The DATABASE_URI settings in this example is also SQLALCHEMY_DB example

jab:
(Sorry for accidentally hitting the “comment and close” while still editing my comment - never meant to close this. Please see my updated comment.)

lgiordani:
Hey @jab don't worry for the temporary close :joy_cat: Yeah it makes sense. Let me mull it over a bit and I will update that part as well. Thanks!

lgiordani:
@jab can you have a look at this new version of the PR? Thanks!

lgiordani:
Thank you @jab!

PR #3674, opened 2020-07-03
Added very small clarification to quickstart.rst

DeepOde:
Added little bit clarification to quickstart.srt so that it'd be easy to understand for persons who don't have English as native language.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -264,15 +264,15 @@ The following two rules differ in their use of a trailing slash. ::
         return 'The about page'
 
 The canonical URL for the ``projects`` endpoint has a trailing slash.
-It's similar to a folder in a file system. If you access ``/projects`` - the URL without
-a trailing slash, Flask redirects you to the canonical URL with the
-trailing slash - ``/projects/``.
+It's similar to a folder in a file system. If you access the URL without
+a trailing slash (``/projects``), Flask redirects you to the canonical URL
+with the trailing slash (``/projects/``).
 
 The canonical URL for the ``about`` endpoint does not have a trailing
-slash. It's similar to the pathname of a file. Accessing ``/about/`` - the URL with a
-trailing slash produces a 404 "Not Found" error. This helps keep URLs
-unique for these resources, which helps search engines avoid indexing
-the same page twice.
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -264,12 +264,12 @@ The following two rules differ in their use of a trailing slash. ::
         return 'The about page'
 
 The canonical URL for the ``projects`` endpoint has a trailing slash.
-It's similar to a folder in a file system. If you access the URL without
+It's similar to a folder in a file system. If you access ``/projects`` - the URL without
```
teymour-aldridge, line 267:
```suggestion
It's similar to a folder in a file system. If you access the URL (``/projects``) without
```

docs/quickstart.rst
```diff
@@ -264,12 +264,12 @@ The following two rules differ in their use of a trailing slash. ::
...
 a trailing slash, Flask redirects you to the canonical URL with the
-trailing slash.
+trailing slash - ``/projects/``.
```
teymour-aldridge, line 269:
```suggestion
trailing slash (``/projects/``).
```

docs/quickstart.rst
```diff
@@ -264,12 +264,12 @@ The following two rules differ in their use of a trailing slash. ::
...
 
 The canonical URL for the ``about`` endpoint does not have a trailing
-slash. It's similar to the pathname of a file. Accessing the URL with a
+slash. It's similar to the pathname of a file. Accessing ``/about/`` - the URL with a
```
teymour-aldridge, line 272:
```suggestion
slash. It's similar to the pathname of a file. Accessing the URL (``/about/``) with a
```

teymour-aldridge:
IMO this is stylistically better.

docs/quickstart.rst
DeepOde: > IMO this is stylistically better.
Yes, agreed.

docs/quickstart.rst
greyli, line 269: It will be better to make the other tips' position consistent with this one. For example, use `Accessing the URL with a trailing slash (``/about/``)` instead of `Accessing the URL (``/about/``) with a trailing slash`.

PR #3683, opened 2020-07-09
Fixing small html issues in flashing patterns docs

jackwardell:
Fixing some simple syntax in flashing patterns.

The examples in https://flask.palletsprojects.com/en/1.1.x/patterns/flashing/ are shown as:
`<ul class=flashes>`
but rather should be:
`<ul class="flashes">`

More examples of this throughout the page. Just fixing it up.

Also touching up some HTML by removing unnecessary p tag and swapping dl tags for input/label tags as they're better for screen readers and improve accessability.

### DIFF
--- a/docs/patterns/flashing.rst
+++ b/docs/patterns/flashing.rst
@@ -110,7 +110,7 @@ categories.  The loop looks slightly different in that situation then:
 
 .. sourcecode:: html+jinja
 
-   {% with messages = get_flashed_messages(with_categories=true) %}
+   {% with messages = get_flashed_messages(with_categories=True) %}
      {% if messages %}
        <ul class="flashes">
        {% for category, message in messages %}
--- a/docs/patterns/flashing.rst
+++ b/docs/patterns/flashing.rst
@@ -79,14 +79,13 @@ And here is the :file:`login.html` template which also inherits from
        <p class="error"><strong>Error:</strong> {{ error }}
      {% endif %}
      <form method="post">
-       <dl>
-         <dt>Username:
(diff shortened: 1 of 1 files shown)


jackwardell:
should I have raised an issue first?

docs/patterns/flashing.rst
```diff
@@ -48,7 +48,7 @@ And here is the :file:`layout.html` template which does the magic:
    <title>My Application</title>
    {% with messages = get_flashed_messages() %}
      {% if messages %}
-       <ul class=flashes>
+       <ul class="flashes">
```
ThiefMaster, line 51: quotes are optional

docs/patterns/flashing.rst
```diff
@@ -110,7 +110,7 @@ categories.  The loop looks slightly different in that situation then:
 
 .. sourcecode:: html+jinja
 
-   {% with messages = get_flashed_messages(with_categories=true) %}
+   {% with messages = get_flashed_messages(with_categories=True) %}
```
ThiefMaster, line 113: lowercase versions of true/false/none are the default in jinja, the python-cased variants are just there as an alias.

jackwardell: thanks for this, I didn't realise! True had always worked for me and I never realised lowercase was the default

docs/patterns/flashing.rst
jackwardell: true, and it's probably my pedanticism, but it's inconsistent across the example and I would argue double quotes are preferred, but fair enough

PR #3684, opened 2020-07-11
make_response uses headers.update

### LINKED ISSUE #3628: Returning Response and headers causes duplicate headers
### Expected Behavior

```
from flask import Flask
app = Flask(__name__)
@app.route('/')
def issue():
    return {'test': 'test'}, {'Content-Type': 'test'}
```
Using `curl -v http://127.0.0.1:5000/` to query the view I expect only one `Content-Type` header > `Content-Type: test`

### Actual Behavior

Duplicate headers are returned

```
< Content-Type: application/json
< Content-Type: test
```

### Environment

* Python version: 3.8.2
* Flask version: 1.1.2
* Werkzeug version: 1.0.1

### Context

This issue also effects responses created with make_response when using a dict or jsonify body + the headers argument with a 'Content-Type':

```
from flask import Flask, make_response
app = Flask(__name__)
@app.route('/')
def issue():
    return make_response({'test': 'test'}, {'Content-Type': 'test'})
```

This issue is caused by jsonify adding a 'Content-Type' header then make_response uses `extent` to add the additional headers, thus leading to the duplicate.

Returning a str/bytes body does not have this problem as no 'Content-Type' is added by flask, if one is missing it is added by werkzeug.

The reason I came across this issue is we have older code which does `return json.dumps(data), 200, {'Content-Type': 'application/json+somecustomtype'}` and I assumed based on the flask docs that just returning the data and letting flask do the jsonify would be better.

chrisngyn:
fixes #3628

Changed `make_response` to use `headers.update` instead of `headers.extend`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,9 +27,14 @@ Unreleased
     instead of PyOpenSSL. :pr:`3492`
 -   When specifying a factory function with ``FLASK_APP``, keyword
     argument can be passed. :issue:`3553`
--   When loading a ``.env`` or ``.flaskenv`` file on top level directory,
-    Flask will not change current work directory to the location of dotenv
-    files, in order to prevent potential confusion. :pr:`3560`
+-   When loading a ``.env`` or ``.flaskenv`` file, the current working
+    directory is no longer changed to the location of the file.
+    :pr:`3560`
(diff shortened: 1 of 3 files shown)


chrisngyn:
#3690 addresses this issue!

PR #3693, opened 2020-07-22
Common API between Flask + Blueprint

### LINKED ISSUE #3215: Identify the common API between Flask and Blueprint classes

YKo20010:
Resolves #3215 

**Similar functions between Flask and Blueprint:**
**Chris:**
- [x] route
- [ ] add_url_rule
- [ ] endpoint
- [x] before_request
- [x] after_request
- [ ] teardown_request

**Yanlam:**
- [x] init
- [x] context_processor
- [x] url_value_preprocessor
- [x] url_defaults
- [x] errorhandler
- [x] register_error_handler

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -2045,7 +2045,7 @@ class Flask(_PackageBoundObject):
 
         # extend existing headers with provided headers
         if headers:
-            rv.headers.extend(headers)
+            rv.headers.update(headers)
 
         return rv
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Hey, thanks a lot for working on this, just a small comment regarding your commit messages: Ideally the first line should be no longer than 50 chars, and never longer than 72 chars (github ellipsizes after that). [This article](https://chris.beams.io/posts/git-commit/) is a great read about commit message formatting.

davidism:
I haven't checked the methods you listed, but along with things that are in common right now, if there's anything obvious that's on `Flask` but not on `Blueprint` for some reason, we can add those too.

PR #3606, opened 2020-05-10
update /docs/testing.rst based on new tutorial tests

isdanni:
# Update the /docs/testing.rst

Update the`/docs/testing.rst`. Remove the incompatible code snippet like this [one](https://flask.palletsprojects.com/en/1.0.x/testing/#the-testing-skeleton) since the `flaskr` application has been rewritten and does not contain a `flaskr` object.

### a. Sections removed from /docs/testing.rst:

1. The Testing Skeleton;
2. The First Test;
3. Logging In and Out;
4. Test Adding Messages;

### b. Replaced with:
1. The Testing Setup and Fixtures;
2. Database;
3. Authentication;
4. Blog;

### c. Added Section:
- Factory; (The previous version referred to 0.12, which does not contain an app factory, could cause confusions)

Most description is directly taken from `/docs/tutorial/tests.rst` since  @davidism  mentioned [here](https://github.com/pallets/flask/issues/2831#issuecomment-399702709) that the tutorial is update to date;

Relevant issues: #2831,  #3545,  and #2833 .
Possible overlaps: #3349 , #3311 .

**NOTE**: updating the testing docs seems like a long-term job and would take some effort to entirely make the testing docs compatible with the current flask version; I will continue study and update more outdated contents; Meanwhile, any suggestions and help are welcome! :)

PR #3704, opened 2020-07-28
swapping self.assert_equal for assert

jackwardell:
`self.assert_equal(data['username'], my_user.username)` is not in a class and is ambiguous. 

I think this line is technically incorrect in the context of the example snippet: `self.assert_equal` means nothing without a greater context (which isn't shared) whereas `assert` is explicit and clear.

I know that in the grand scheme of things this detail is inconsequential and that the docs aren't supposed to production code, but rather a design aid. However, as someone who has leant flask from the docs, I know that detail matters and it's better to be explicit and clear rather than ambiguous and confusing.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -312,7 +312,7 @@ And then to use it::
         with app.test_client() as c:
             resp = c.get('/users/me')
             data = json.loads(resp.data)
-            self.assert_equal(data['username'], my_user.username)
+            assert data['username'] == my_user.username
 
 
 Keeping the Context Around


davidism:
See also #3606

jackwardell:
Thank you, I will see if I can contribute to that issue. 

Also, in general, what are the pallets members appetite for small changes in the docs, etc via PRs?

Occasionally I see things I don't love (e.g. #3683) and occasionally I see things like the above and I wonder if I should put in a PR, is there explicit policy for what deserves a PR?

davidism:
If you find typos or awkward wording in the docs, those are fine. 

Common things that don't get merged:

* Changing code style.
* Adding imports that are shown earlier in a page.
* Making code "runnable" that's not intended to be standalone.
* Shuffle things around rather than actually making them objectively clearer.
* A change that isn't actually correct, or subtly changes the meaning of something.

jackwardell:
Thank you for that feedback

davidism:
We have a Discord server if you'd like to ask questions about contributing: https://discord.gg/pallets

jackwardell:
Very cool, I will join

PR #3708, opened 2020-07-29
Common API Between Blueprints + Flask

### LINKED ISSUE #3215: Identify the common API between Flask and Blueprint classes

YKo20010:
Resolves #3215 

**Similar functions between Flask and Blueprint:**
**Chris:**
- [x] route
- [ ] add_url_rule
- [x] endpoint
- [x] before_request
- [x] after_request
- [ ] teardown_request

**Yanlam:**
- [x] init
- [x] context_processor
- [x] url_value_preprocessor
- [x] url_defaults
- [x] errorhandler
- [x] register_error_handler


src/flask/blueprints.py
```diff
@@ -241,6 +244,35 @@ def register(self, app, options, first_registration=False):
                 endpoint="static",
             )
 
+        # Merge app and self dictionaries.
+        def merge_dict_lists(self_dict, app_dict):
+            """Merges self_dict into app_dict. Replaces None keys with self.name.
+            Values of dict must be lists.
+            """
+            for key, values in self_dict.items():
+                key = self.name if key is None else f"{self.name}"
```
ThiefMaster, line 253: This line doesn't seem to make much sense. It's basically equivalent to `key = str(self.name`)`

YKo20010: I changed it to `else key` for now, but I'm unsure if that's the right approach. Logically, I don't see a reason for changing the key if it exists, but the tests seem to pass either way, so I'd like to double check first. 
@davidism @ThiefMaster What do you think?

src/flask/blueprints.py
```diff
@@ -167,14 +167,17 @@ def __init__(
         root_path=None,
         cli_group=_sentinel,
     ):
-        _PackageBoundObject.__init__(
-            self, import_name, template_folder, root_path=root_path
+        Skeleton.__init__(
```
davidism, line 170: Use `super()` here

src/flask/blueprints.py
davidism: The implementation is supposed to be: if None, use the blueprint name, otherwise prepend the blueprint name to the key.

```python
key = self.name if key is None else f"{self.name}.{key}"
```

src/flask/app.py
```diff
@@ -84,132 +84,46 @@ def wrapper_func(self, *args, **kwargs):
...
-        information and a lot more.
-
-        So it's important what you provide there.  If you are using a single
-        module, `__name__` is always the correct value.  If you however are
-        using a package, it's usually recommended to hardcode the name of
-        your package there.
-
-        For example if your application is defined in :file:`yourapplication/app.py`
-        you should create it with one of the two versions below::
+class Skeleton(_PackageBoundObject):
```
davidism, line 87: Move this (probably along with `_PackageBoundObject`) to a separate file, like `skeleton.py` (final name still pending). Right now it's making the diff really confusing.

src/flask/app.py
```diff
@@ -335,79 +170,33 @@ class Flask(_PackageBoundObject):
...
-    #: .. versionadded:: 0.7
-    url_rule_class = Rule
+    def __init__(
+        self,
+        import_name,
+        template_folder,
+        root_path=None,
+        static_url_path=None,
+        static_folder="static",
+        instance_relative_config=False,
```
davidism, line 180: Blueprints don't have instance config or instance folders, so these shouldn't be in the skeleton.

src/flask/app.py
```diff
@@ -84,132 +84,46 @@ def wrapper_func(self, *args, **kwargs):
...
-        automatically, such as for namespace packages.
-    """
+    #: Absolute path to the package on the filesystem. Used to look up
+    #: resources contained in the package.
+    root_path = None
 
-    #: The class that is used for request objects.  See :class:`~flask.Request`
-    #: for more information.
-    request_class = Request
+    # MARK: additional vars added from Flask.
```
davidism, line 109: Most of these should probably stay in `Flask`, blueprints don't have things like a Jinja environment or config.

src/flask/blueprints.py
YKo20010: @davidism Thanks! I'll update that!

src/flask/app.py
YKo20010: Should some of the Skeleton functions in Blueprints be overridden?

My workflow has been like this:
```
1. Comment out / Delete the function declaration (ie. def before_request) in Blueprints.
2. pytest - See what error comes up (usually function or variable definition undefined in Blueprints).
3. Move that func/var from Flask to Skeleton.
4. Repeat [2-3] until reached an **AssertionError**.
5. Update Blueprints register to merge self's dict into app's dict.
```

For context, I moved `debug` to Skeleton when I removed `before_request` (this was the first function we refactored) from Blueprints because pytest gave me an error that it wasn't implemented. I moved all the Jinja functions because I saw that `debug` sets `self.jinja_env.auto_reload` (though the tests pass without moving the Jinja functions over).

YKo20010: **Update:**
I experimented a bit with our old branch and found that the `debug` error doesn't show if I remove `@setupmethod` from the respective function in `Skeleton`. Would this be the right approach to take?

Also in this case, how would we add `@setupmethod` for `Flask`?

davidism: `debug` is only a property of `Flask`, it shouldn't be in `Skeleton`. Same with at least the stuff related to the Jinja environment and Config.

YKo20010: Thanks for clarifying! 
I realized that `Blueprint` doesn't have any `@setupmethod`, which is where my `debug` error was coming from. Hence, if `Skeleton` contains the "common" function of `Flask` without the `@setupmethod`, then the tests work as expected.

Since `Flask` still uses `@setupmethod`, should `Flask` override the respective `Skeleton` functions somewhat like this?
```
 @setupmethod
 def url_defaults(self, f):
        super().url_defaults(...)
```

PR #3709, opened 2020-07-30
Common API Between Flask + Blueprint

### LINKED ISSUE #3215: Identify the common API between Flask and Blueprint classes

YKo20010:
Resolves #3215

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1,7 +1,6 @@
 import os
 import sys
 from datetime import timedelta
-from functools import update_wrapper
 from itertools import chain
 from threading import Lock
 
@@ -9,7 +8,6 @@ from werkzeug.datastructures import Headers
 from werkzeug.datastructures import ImmutableDict
 from werkzeug.exceptions import BadRequest
 from werkzeug.exceptions import BadRequestKeyError
-from werkzeug.exceptions import default_exceptions
 from werkzeug.exceptions import HTTPException
 from werkzeug.exceptions import InternalServerError
 from werkzeug.routing import BuildError
@@ -30,8 +28,6 @@ from .globals import _request_ctx_stack
 from .globals import g
 from .globals import request
 from .globals import session
-from .helpers import _endpoint_from_view_func
-from .helpers import _PackageBoundObject
 from .helpers import find_package
 from .helpers import get_debug_flag
 from .helpers import get_env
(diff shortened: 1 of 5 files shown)


YKo20010:
From the previous draft PR on our experimental branch:
image

davidism:
`Skeleton`, `Flask`, and `Blueprint` should use `setupmethod` consistently. The fact that blueprints weren't using it (or some form of it) is probably an oversight. The point of `setupmethod` is to ensure those methods aren't called after the application starts because it would lead to inconsitent results. That seems even more true for blueprints, where adding a route, etc. to a blueprint after it's registered means it won't have been added to the app.

davidism:
Since `setupmethod` is specific to `Flask` right now (uses `debug` and `_got_first_request`, which are `Flask`-only), this could be extracted to a helper method that both `Flask` and `Blueprint` implement.

```python
def setupmethod(f):
    @wraps(f)
    def wrapped(self, *args, **kwargs):
        if self._is_setup_finished():
            raise AssertionError(...)

        return f(*args, **kwargs)

    return wrapper

class Skeleton:
    def _is_setup_finished(self):
        raise NotImplementedError

class Flask:
    def _is_setup_finished(self):
        return self.debug and self._got_first_request

class Blueprint:
    def _is_setup_finished(self):
        return self.warn_on_modifications and self._got_registered_once
```

davidism:
This does change the behavior a bit, since `Blueprint.warn_on_modifications` shows a warning instead of raising an error. For some reason, `warn_on_modifications` is disabled by default as well, and I'm not really clear why an `AssertionError` was used instead of a `RuntimeError`. I'll address that in a later issue though.

davidism:
Regarding naming, a few of us in Discord liked the name `Scaffold`, it fits with Pallets and `Blueprint`. I'll let it sit a while longer, no need to rename it yet.

chrisngyn:
I saw the messages in the Discord about switching from `skeleton.py` and `class Skeleton` to `scaffold.py` and `class Scaffold` respectively, just scrolled up and saw your comment about not needing to change it just yet -- sorry about that! 😅 We can always revert if we decide against it. 👍

src/flask/app.py
```diff
@@ -371,7 +371,7 @@ def __init__(
         instance_relative_config=False,
         root_path=None,
     ):
-        Skeleton.__init__(
+        Scaffold.__init__(
```
davidism, line 374: Use `super()`

src/flask/blueprints.py
```diff
@@ -167,7 +167,7 @@ def __init__(
         root_path=None,
         cli_group=_sentinel,
     ):
-        Skeleton.__init__(
+        Scaffold.__init__(
```
davidism, line 170: Use `super()`

src/flask/blueprints.py
```diff
@@ -258,17 +293,16 @@ def register(self, app, options, first_registration=False):
...
-    def route(self, rule, **options):
-        """Like :meth:`Flask.route` but for a blueprint.  The endpoint for the
-        :func:`url_for` function is prefixed with the name of the blueprint.
-        """
-
-        def decorator(f):
-            endpoint = options.pop("endpoint", f.__name__)
-            self.add_url_rule(rule, endpoint, f, **options)
-            return f
+    # def route(self, rule, **options):
```
davidism, line 296: leftover commented code

davidism:
Moved a couple minor things into `scaffold`, added a changelog.

PR #3726, opened 2020-08-10
Include samesite and secure keys when removing session cookie

darcatron:
The `samesite` key is required on empty cookies. When a flask `session` is cleared, the delete cookie logic needs to set the configured `samesite` and `secure` keys.

samesite missing on cookie

Requires that https://github.com/pallets/werkzeug/pull/1889 is merged first.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,8 @@ Unreleased
     attributes just like ``Flask``, rather than opaque lambda functions.
     This is intended to improve consistency and maintainability.
     :issue:`3215`
+-   Include ``samesite`` and ``secure`` options when removing the
+    session cookie. :pr:`3726`
 
 
 Version 1.1.x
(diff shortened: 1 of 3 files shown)


darcatron:
Running `pytest` succeded but the full suite via `tox` is failing on the `delete_cookie` method I'm utilizing. Seems like I'll need to first update https://github.com/pallets/werkzeug/blob/b45ac05b7feb30d4611d6b754bd94334ece4b1cd/src/werkzeug/test.py#L886 to support `samesite` and `secure`

pgjones:
See, https://github.com/pallets/werkzeug/pull/1889 which I think is a blocker for this.

darcatron:
@pgjones thanks for the reference. It definitely looks like that PR will be necessary for this functionality. I'll also need to update the [werkzeug test session class](https://github.com/pallets/werkzeug/blob/b45ac05b7feb30d4611d6b754bd94334ece4b1cd/src/werkzeug/test.py#L886) to have nearly identical changes to your PR.

PR #3728, opened 2020-08-13
Discover app factory deterministically

nphilipp:
Commit 2ae740dd4903bbe5f3864384bcdf2ab2d4638bce probably inadvertently
changed to iterate over a set of "create_app", "make_app" to discover
application factory functions. As sets don't guarantee order,
differences in the implementation of the type can make it
non-deterministic which factory function is used if a project implements
both. Revert to using a tuple.

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -60,7 +60,7 @@ def find_best_app(script_info, module):
         )
 
     # Search for app factory functions.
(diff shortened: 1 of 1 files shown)


davidism:
Seems to be a mistake, must have been thinking of a `contains` operation at some point.

PR #3733, opened 2020-08-17
Disable required attributes overriding error output

matthkang:
Error messages were being overriden with a popup, "Please fill out this field", because of the required attribute from the respective templates. Preserved custom error messages by disabling the required attribute.


ThiefMaster:
I don't think we want to fail only after submission. The point of server-side checks is so you cannot send bad data even if you tinker with the client (e.g. removing required attributes, or using a different client altogether).

PR #3740, opened 2020-08-23
Fix pre-commit cache key in GitHub Actions

jab:
Fix pre-commit cache key to include full Python version as per https://github.com/pre-commit/pre-commit/issues/1575#issuecomment-678792700 and https://pre-commit.com/#github-actions-example.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,11 +43,14 @@ jobs:
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key: pip|${{ runner.os }}|${{ matrix.python }}|${{ hashFiles('setup.py') }}|${{ hashFiles('requirements/*.txt') }}
+      - name: set full Python version in PY env var
+        # See https://pre-commit.com/#github-actions-example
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -43,11 +43,15 @@ jobs:
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key: pip|${{ runner.os }}|${{ matrix.python }}|${{ hashFiles('setup.py') }}|${{ hashFiles('requirements/*.txt') }}
+      - name: set full Python version
+        run: echo ::set-env name=PY_VER_FULL::$(python --version)
```
asottile, line 47: you want `--version --version` as the docs indicate: https://pre-commit.com/#github-actions-example

jab: Thanks for reviewing this! Updated in the latest revision (to use the exact `python -VV | sha256sum | cut -d' ' -f1` spelling documented). (And now I've learned what the extra `-V` does, which before I mistakenly thought was a no-op because I was looking at the `python -VV` and the `pypy3 -VV` behavior, not realizing this is a CPython 3 thing. TIL!)

PR #3755, opened 2020-09-11
Use content-encoding from mimetypes check in `helpers.py:send_file`

arabidopsis:
in `helpers.py/send_file` the `mimetypes` check ignores the content encoding.

Use case: Usually static content is delivered by the frontend server but it's annoying that the debug server
can't understand files such as compressed svg (viz: svgz). mimetypes *does* understand this but flask
ignores the output. This patch just checks the encoding and add extra headers.

I haven't added any tests but would be happy to if requested (and instructed where to put it!).


davidism:
The implementation of `send_file` was moved to Werkzeug, so this won't be merged here. I'll move it to Werkzeug.

davidism:
Moved to Werkzeug in pallets/werkzeug#2029, listed you as a co-author on the commit. I didn't set the `Vary` header as it wasn't clear that was correct.

PR #3757, opened 2020-09-14
DOC: Simplified loading of favicon

gondo:
Simplified loading of favicon from static folder

### DIFF
--- a/docs/patterns/favicon.rst
+++ b/docs/patterns/favicon.rst
@@ -28,15 +28,13 @@ however your application is the root you can simply route a redirect::
                      redirect_to=url_for('static', filename='favicon.ico'))
 
 If you want to save the extra redirect request you can also write a view
-using :func:`~flask.send_from_directory`::
+using :func:`~flask.send_static_file`::
 
-    import os
-    from flask import send_from_directory
+    from flask import current_app
 
     @app.route('/favicon.ico')
     def favicon():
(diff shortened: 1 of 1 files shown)


greyli:
Hi, thank you for working on this. `send_staic_file` is initially for internal use, and it doesn't accept any keyword parameters like `mimetype`, but we do need to pass the `mimetype` parameter (see paragraph under the code snippet).

gondo:
@greyli does it really matter what `send_staic_file` was intented for if it is exposed for everyone to use? Also the proposed solution works for me without the need to pass `mimetype`. In any-case, the whole issues around favicon is purely in cases when flask is handling static requests and that should only be an issue during development as in production static requests should be handled by f.e. nginx

greyli:
It doesn't matter, but your change is incompatible with the description below the code (also the description explained why need to pass the mimetype):

> We can leave out the explicit mimetype and it will be guessed, but we may as well specify it to avoid the extra guessing, as it will always be the same.

Yes, normally you will use web server to serve static files, the description below the code also address it: 

> The above will serve the icon via your application and if possible it's better to configure your dedicated web server to serve it; refer to the web server's documentation.

In a word, this example is just for older browsers that can't recognize `<link rel="shortcut icon">` element, as described in the docs, you only need this with modern browsers:

```html
<link rel="shortcut icon" href="{{ url_for('static', filename='favicon.ico') }}">
```

PR #3762, opened 2020-09-23
Breaking a ref cycle created by default in Flask()

### LINKED ISSUE #3761: A Flask app holds a reference to itself
```python
from flask import Flask
import objgraph

def make_app():
    app = Flask(__name__)

    @app.route('/')
    def hello_world():
        return 'Hello, World!'

    return app

def test():
    myapp = make_app()
    objgraph.show_backrefs([myapp], filename='backrefs.png', max_depth=6)

test()
```

Here's what the produced reference graph looks like:
backrefs

Is it the intended behavior? In our library a Flask app is created dynamically in a closure, and as a result it retains anything else that was created in that closure. We work around this by having a trampoline function and passing heavy objects as weakrefs, but it was still unexpected.

### Environment

* Python version: 3.7.5
* Flask version: 1.1.2
* Werkzeug version: 1.0.1

fjarri:
Fixes #3761

Breaks the ref cycle `Flask` -> `view_functions` -> `self.send_static_file` -> `Flask` that is created in the constructor.

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import weakref
 from datetime import timedelta
 from itertools import chain
 from threading import Lock
@@ -474,6 +475,7 @@ class Flask(Scaffold):
         # Note we do this without checking if static_folder exists.
         # For one, it might be created while the server is running (e.g. during
         # development). Also, Google App Engine stores static files somewhere
+        self_ref = weakref.ref(self)
         if self.has_static_folder:
             assert (
                 bool(static_host) == host_matching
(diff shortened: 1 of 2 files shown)


jab (commented):
Thanks for the PR! I’m not sure if we’ll end up accepting this change (defer to @davidism), but in the meantime it wouldn’t hurt to add some test coverage. Here’s an example of some test coverage I added to one of my projects to prove my weakrefs are working as intended, in case it’s helpful: https://github.com/jab/bidict/blob/master/tests/properties/test_properties.py#L273

fjarri:
Thanks for the pointer, I added a test (shamelessly most of it from your project).

Issue #3768, opened 2020-09-27
Typo in Flask pallets fileuploads


miquelvir:
### Expected Behavior
"if user does not select file, browsers also submit an empty part without filename" [missing plural s in 'browser']
or
"if user does not select file, browser also submits an empty part without filename" [missing present simple s in 'submit']

### Actual Behavior
"if user does not select file, browser also submit an empty part without filename"

Sorry if it should have been a pull request.

ThiefMaster:
Yes, a PR would have been much better for this :)

Also, there should be a "the" before "user".

miquelvir:
Okey, I'm going to close the issue and create a PR then.

PR #3769, opened 2020-09-27
grammar mistake fix

miquelvir:
Grammar mistake in comment from the fileuploads patterns example in the docs.

### DIFF
--- a/docs/patterns/fileuploads.rst
+++ b/docs/patterns/fileuploads.rst
@@ -55,8 +55,8 @@ the file and redirects the user to the URL for the uploaded file::
                 flash('No file part')
                 return redirect(request.url)
             file = request.files['file']
-            # if user does not select file, browser also
-            # submit an empty part without filename
+            # If the user does not select a file, the browser submits an
+            # empty file without a filename.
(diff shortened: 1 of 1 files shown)


davidism:
There are still grammatical issues with this. If fixing a top is important, please clean up the entire area.

miquelvir:
Hi @davidism , it is my first PR and I am not a native English speaker. I have tried correcting all the mistakes in the file, but please let me know if I have done something wrong.

davidism:
Sorry, I guess I wasn't clear. I was referring to the specific sentence you were editing, not the entire document.

davidism:
Note that the original fix wasn't particularly necessary either.

> # if user does not select file, browser also
> # submit an empty part without filename

This sort of style, without the "a" and "the" before nouns, is common in code comments. The only thing wrong here is "browser also submit" should be "browser also submits".

miquelvir:
Thank's for the clarification. It was not my original intention to change the 'the'; I only wanted to add the missing present simple 's', but was told to also add the 'the' before user: https://github.com/pallets/flask/issues/3768.

Issue #3780, opened 2020-10-07
@app.route should either trim urls or throw an error when route ending with space is provided


lalnuo:
### Expected Behavior
If user writes code like this, @app.route should either throw an exception or trim the given path.
```python
from flask import Flask

app = Flask(__name__)

@app.route("/foobar ")
def hello():
    return "Hello, World!"
```

### Actual Behavior
@app.route silently accepts the url that ends with space and navigating to `/foobar` doesnt work. 

### Environment

* Python version: Python 3.8.0
* Flask version: 1.1.2
* Werkzeug version:

davidism:
While this is probably not a good URL to use, there is nothing invalid about it. This would be a good test to add to a linter, such as a flake8 plugin for flask-related checks. But it is not an error, and would be incorrect for Flask to change the URL.

```python
from flask import Flask, url_for

app = Flask(__name__)

@app.route("/")
def index():
    return url_for("space")

@app.route("/space ")
def space():
    return "space"
```

```
/space%20
```

PR #3781, opened 2020-10-07
Raise an error if rule starts or ends with a space

lalnuo:
This PR solves https://github.com/pallets/flask/issues/3780. After this PR is merged creating routes like this will throw an error:

`@bp.route("/bar ")`
`@bp.route(" /bar")`


davidism:
See #3780 for close reason.

PR #3782, opened 2020-10-08
Ready flask app for beginners

DhyanTD:
create a basic blogging website with this module.
note: you have to use html templates for main interface , 
do little modification has per your needs


davidism:
Thanks, but we already have an example paired with the tutorial, and Flask-SQLAlchemy has the same example adapted to use SQLAlchemy.

PR #3786, opened 2020-10-10
Parmetrizing tests

sanders41:
Removing variable loop from within test cases and moving them to pytest parameters.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -895,9 +895,9 @@ class TestStreaming:
 
 
 class TestSafeJoin:
-    def test_safe_join(self):
-        # Valid combinations of *args and expected joined paths.
-        passing = (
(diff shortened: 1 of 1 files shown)


tests/test_helpers.py
```diff
@@ -912,14 +912,16 @@ def test_safe_join(self):
...
             (("../", "a/b/c"), "../a/b/c"),
             (("/..",), "/.."),
-        )
-
-        for args, expected in passing:
-            assert flask.safe_join(*args) == expected
+        ),
+    )
+    def test_safe_join(self, passing):
+        args = passing[0]
```
davidism, line 918: Split this into multiple parameters. `("args", "expected")` instead of `"passing"`.

sanders41: I updated them to args and expected as requested.

PR #3793, opened 2020-10-16
Add signed int and float to docs

adithyay328:
Adds signed int and float converter types to the docs.

Signed ints and floats were added in this PR to werkzeug, but were never added to the flask docs:
https://github.com/pallets/werkzeug/pull/1355

I checked to make sure that these URL paramater types were available in flask, and after testing the following code, I can confirm that they are actually supported.

```
from flask import Flask, request, Response
app = Flask(__name__)
    
@app.route("/int/<int:intParam>", methods=["GET"])
def normalInt(intParam):
    return Response("normalInt: " + str(intParam))

@app.route("/int/<int(signed=True):intParam>", methods=["GET"])
def signedInt(intParam):
    return Response("signedInt: " + str(intParam))
    
@app.route("/float/<float:floatParam>", methods=["GET"])
def normalFloat(floatParam):
    return Response("normalFloat: " + str(floatParam))

@app.route("/float/<float(signed=True):floatParam>", methods=["GET"])
def signedFloat(floatParam):
    return Response("signedFloat: " + str(floatParam))
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -608,14 +608,15 @@ Variable parts are passed to the view function as keyword arguments.
 
 The following converters are available:
 
-=========== ===============================================
-`string`    accepts any text without a slash (the default)
-`int`       accepts integers
-`float`     like `int` but for floating point values
-`path`      like the default but also accepts slashes
-`any`       matches one of the items provided
-`uuid`      accepts UUID strings
-=========== ===============================================
+======================   ====================================================
+``string``               (default) accepts any text without a slash
+``int``                  accepts positive integers
+``int(signed=True)``     accepts positive and negative integers
+``float``                accepts positive floating point values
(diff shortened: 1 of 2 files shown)


davidism:
Thanks, but I don't think I want to add this right now. These are not different types, they're types with arguments. Note that the `any` converter is not included either, nor are the options for `str` or the other options for `int` and `float`. I'd rather find a better way (probably a pallets-sphinx-theme domain, similar to Jinja) to link the converters to their docs in Werkzeug.

davidism:
For now, if you want to make a new PR that makes these names link to their docs in Werkzeug, that would be good.

adithyay328:
@davidism Thanks for the feedback! I'll get on that soon.

PR #3794, opened 2020-10-17
Ensure session_interface.open_session is called after URL matching

### LINKED ISSUE #3776: request.endpoint always None in SessionInterface.open_session
### Expected Behavior

Session Interface should have access to `request.endpoint` from `open_session`. In Versions before 1.1.0, this worked fine.

### Actual Behavior
Our SessionInterface subclass looks at `request.endpoint` to determine if it should open a new session - opening a new is only appropriate on some endpoints. Now this is always `None`, so it cannot make this distinction anymore.

I believe #3104 (issue #3088) broke this.

### Environment

* Python version: 3.8.5
* Flask version: 1.1.2
* Werkzeug version: 0.16.1

mbpreble:
Fixes #3776

This moves a call to `match_request` prior to invocation of `session_interface.open_session`. 
There is a new test which confirms that test application can access `request.endpont` in its `open_session` implementation - this test fails prior to the change since an assertion error triggers a 500 status.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -60,6 +60,9 @@ Unreleased
     conditional requests instead of using a timed cache. :pr:`3828`
 -   ``helpers.safe_join`` is deprecated. Use
     ``werkzeug.utils.safe_join`` instead. :pr:`3828`
+-   The request context does route matching before opening the session.
+    This could allow a session interface to change behavior based on
+    ``request.endpoint``. :issue:`3776`
 
 
 Version 1.1.2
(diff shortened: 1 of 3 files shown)


davidism:
I need to review this in more detail, I'm not sure if there was a reason for the ordering. This code was last changed a year ago in #3104. Since I'm not going to get to this before October's over, I'm marking this valid for Hacktoberfest in case you're participating.

davidism:
I can't see any reason not to do this right now. The previous PR I linked moved it from a completely different part of the code, I think we just didn't consider back then whether it made sense to move it anywhere but the end of `push`.

davidism:
Planning to roll this back, see #4053

davidism:
Reverted in #4055

PR #3797, opened 2020-10-18
added numpy requirement

anshroge:


ThiefMaster:
why would we want that? flask doesn't use numpy and neither do the docs. not to mention that this file is generated based on `requirements/docs.in`...

PR #3799, opened 2020-10-20
Small fixes and improvements

proggga:
Small fixes, noting to add in changes

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -284,11 +284,11 @@ class Blueprint(Scaffold):
 
         if cli_resolved_group is None:
             app.cli.commands.update(self.cli.commands)
-        elif cli_resolved_group is _sentinel:
-            self.cli.name = self.name
-            app.cli.add_command(self.cli)
         else:
-            self.cli.name = cli_resolved_group
+            if cli_resolved_group is _sentinel:
+                self.cli.name = self.name
+            else:
+                self.cli.name = cli_resolved_group
             app.cli.add_command(self.cli)
 
     def add_url_rule(self, rule, endpoint=None, view_func=None, **options):
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -640,7 +640,7 @@ def load_dotenv(path=None):
 
         return False
 
-    new_dir = None
+    env_loaded = False
 
     for name in (".env", ".flaskenv"):
         path = dotenv.find_dotenv(name, usecwd=True)
@@ -648,12 +648,9 @@ def load_dotenv(path=None):
         if not path:
             continue
 
-        if new_dir is None:
-            new_dir = os.path.dirname(path)
+        env_loaded = dotenv.load_dotenv(path) or env_loaded
 
-        dotenv.load_dotenv(path)
-
-    return new_dir is not None  # at least one file was located and loaded
+    return env_loaded
 
 
 def show_server_banner(env, debug, app_import_path, eager_loading):
@@ -709,7 +706,7 @@ class CertParamType(click.ParamType):
         except click.BadParameter:
             value = click.STRING(value, param, ctx).lower()
 
-            if value == "adhoc":
+            if _isAdHoc(value):
                 try:
                     import cryptography  # noqa: F401
                 except ImportError:
@@ -734,7 +731,7 @@ def _validate_key(ctx, param, value):
     Modifies the ``cert`` param to be a ``(cert, key)`` pair if needed.
     """
     cert = ctx.params.get("cert")
-    is_adhoc = cert == "adhoc"
+    is_adhoc = _isAdHoc(cert)
     is_context = ssl and isinstance(cert, ssl.SSLContext)
 
     if value is not None:
@@ -758,6 +755,11 @@ def _validate_key(ctx, param, value):
             raise click.BadParameter('Required when using "--cert".', ctx, param)
 
     return value
+
+
+def _isAdHoc(cert):
+    """Check is it adhoc"""
+    return cert == "adhoc"
 
 
 class SeparatedPathType(click.Path):
--- a/src/flask/helpers.py
+++ b/src/flask/helpers.py
@@ -292,7 +292,7 @@ def url_for(endpoint, **values):
         url_adapter = reqctx.url_adapter
         blueprint_name = request.blueprint
 
-        if endpoint[:1] == ".":
+        if endpoint[0] == ".":
             if blueprint_name is not None:
                 endpoint = f"{blueprint_name}{endpoint}"
             else:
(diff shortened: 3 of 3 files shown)


davidism:
What does this fix? Many of these changes seem unrelated to each other. Many of them are incorrect, or shuffle things around for no apparent reason.

Looking through your other recent PRs, it seems you're making these sorts of changes to a lot of projects. Please be respectful of maintainers time by not making unnecessary changes and clearly explaining the reasons for the changes you do make.

proggga:
> What does this fix? Many of these changes seem unrelated to each other. Many of them are incorrect, or shuffle things around for no apparent reason.
> 
> Looking through your other recent PRs, it seems you're making these sorts of changes to a lot of projects. Please be respectful of maintainers time by not making unnecessary changes and clearly explaining the reasons for the changes you do make.

I really like to improve readablity of code. This is my fetish. All my code is correct. Look carefully. Most of projects like this one. Please recheck:

this one will fix logic and improve speed (bool is faster then work with string)
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R651

this code incapsulate ad-hoc constant to the one place in code (ye better to create const somewhere, but I don't know the right place)
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R760

access to right char is faster then slice
https://github.com/pallets/flask/pull/3799/files#diff-1f51c8ded4d4ff7e13badab599ef22436c529c2b5f9c25dc6250c1f9fd985440R295

multiple update faster then many access, will be optimized by python
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R651

style changes, else and elif after return is not very good
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R651
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R651
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R651

here I removed also 'else'
https://github.com/pallets/flask/pull/3799/files#diff-fed8939d1905b99bba605ad91e9ccdf5ede6223e03ccbc3f0121853035051e62R651

proggga:
> Many of them are incorrect
If I've made incorrect change, please say where, I don't want to break something. If I've made this, only by mistake.

I really try help with style, small mistakes, improve readability and code support (remove unneeded tabs when you have return + else/elif

may be can fix this PR in some case

src/flask/blueprints.py
```diff
@@ -284,11 +284,11 @@ def merge_dict_nested(self_dict, app_dict):
 
         if cli_resolved_group is None:
             app.cli.commands.update(self.cli.commands)
-        elif cli_resolved_group is _sentinel:
-            self.cli.name = self.name
-            app.cli.add_command(self.cli)
         else:
-            self.cli.name = cli_resolved_group
+            if cli_resolved_group is _sentinel:
```
ThiefMaster, line 288: This change does not improve readability. Repeating a line once (or even twice) is perfectly fine; being overly DRY actually makes code harder to read here.

src/flask/cli.py
```diff
@@ -709,7 +706,7 @@ def convert(self, value, param, ctx):
         except click.BadParameter:
             value = click.STRING(value, param, ctx).lower()
 
-            if value == "adhoc":
+            if _isAdHoc(value):
```
ThiefMaster, line 709: camelCase has no place in Python code (see PEP8). Also, you replace a very simple check with a function call where you now need to check the function first to see what it does!

src/flask/cli.py
```diff
@@ -760,6 +757,11 @@ def _validate_key(ctx, param, value):
     return value
 
 
+def _isAdHoc(cert):
+    """Check is it adhoc"""
```
ThiefMaster, line 761: This docstring is the type of documentation that nobody really needs :) It's like "loop over list" above a loop...

src/flask/helpers.py
```diff
@@ -292,7 +292,7 @@ def external_url_handler(error, endpoint, values):
         url_adapter = reqctx.url_adapter
         blueprint_name = request.blueprint
 
-        if endpoint[:1] == ".":
+        if endpoint[0] == ".":
```
ThiefMaster, line 295: You are changing behavior here:

```python
>>> empty = ''
>>> empty[:1]
''
>>> empty[0]
  File "<stdin>", line 1, in <module>
IndexError: string index out of range
```

src/flask/helpers.py
```diff
@@ -340,10 +340,14 @@ def external_url_handler(error, endpoint, values):
...
-        values["_method"] = method
-        values["_scheme"] = scheme
+        values.update(
+            {
+                "_external": external,
+                "_anchor": anchor,
+                "_method": method,
+                "_scheme": scheme,
+            }
+        )
```
ThiefMaster, line 350: This is longer and doesn't look better. The usecase for `dict.update()` is when you already have a sequence...

src/flask/helpers.py
```diff
@@ -841,38 +845,37 @@ def _find_package_path(root_mod_name):
         if spec.origin in {"namespace", None}:
             return os.path.dirname(next(iter(spec.submodule_search_locations)))
         # a package (with __init__.py)
-        elif spec.submodule_search_locations:
+        if spec.submodule_search_locations:
             return os.path.dirname(os.path.dirname(spec.origin))
         # just a normal module
-        else:
-            return os.path.dirname(spec.origin)
+        return os.path.dirname(spec.origin)
```
ThiefMaster, line 851: With the old code every return was on the same levl - now the last one is on a different level. Which option to choose is completely personal preference though, but I don't think sending PRs to projects you are not related to do make changes that are purely based on personal preference is helpful.

ThiefMaster (commented):
I'm someone who cares a lot about clean and readable code. Your PR does not really help in that regard - I added some comments explaining why that's the case.

src/flask/helpers.py
proggga: Hm, thats true, made a mistake

src/flask/cli.py
proggga: Yea, Its my problem from go, start write everywhere camecase, can fx this

proggga:
Please tell, what about other changes.
I can revert some changes

ThiefMaster:
I don't think it makes much sense to do anything with this PR - none of the changes are really worth a PR...

proggga:
> I don't think it makes much sense to do anything with this PR - none of the changes are really worth a PR...

Thanks, Ive really tried.

davidism:
If you're interested in contributing, there are plenty of open issues with stuff we're actually asking for. However, we won't accept random changes like this.

PR #3802, opened 2020-10-21
[Minor changes] Update documentation by removing outdated information

bartbroere:
I noticed something about the market share of Internet Explorer and thought "that can't be true". After fixing that I also found out that Mark Pilgrim's book has been published, so I updated that as well.

### DIFF
--- a/docs/htmlfaq.rst
+++ b/docs/htmlfaq.rst
@@ -183,9 +183,8 @@ and to use.
     geolocation, and offline applications.
 
 Many other features have been added, as well. A good guide to new features
-in HTML5 is Mark Pilgrim's book, `Dive Into HTML5`_ ( `HTML5 Up & Running`_
-as published by O'Reilly).  Not all of them are supported in browsers yet,
-however, so use caution.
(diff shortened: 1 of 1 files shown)


davidism:
Thanks. I'd rather not include the link to the book, we don't really include links to paid resources. This whole page should probably be removed, it doesn't seem particularly relevant anymore.

bartbroere:
> Thanks. I'd rather not include the link to the book, we don't really include links to paid resources. This whole page should probably be removed, it doesn't seem particularly relevant anymore.

I agree that it seemed a bit outdated to me while reading it. Should I remove this pull request, or change it to omit the O'Reilly link?

davidism:
You can either remove the link or remove the page, I'll merge either one.

davidism:
Looks like there's some leftover trailing spaces, need to remove those to get the style test to pass.

bartbroere:
Okay, I have removed the O'Reilly link. Thanks for the quick reply!

bartbroere:
> Looks like there's some leftover trailing spaces, need to remove those to get the style test to pass. (Don't worry about the range test, that's expected to fail right now.)

I got them out of there (finally).

PR #3804, opened 2020-10-23
fix html in example 404/500 templates

gpwclark:


ThiefMaster:
Those end tags are optional in HTML5.

PR #3805, opened 2020-10-24
cli.load_dotenv: when path provided and python-dotenv not found, extra message

proggga:
cli.load_dotenv showed wrong message, when path to cli.load_dotenv provided,
now it show right message

make small refactor of this func

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -626,21 +626,21 @@ def load_dotenv(path=None):
     FLASK_ENV_FILE = ".flaskenv"
 
     if dotenv is None:
+        msg = ""
         if path:
-            click.secho(
-                f" Env file {path} will not loaded, module python-dotenv not found"
-                ' Do "pip install python-dotenv" to use it.',
-                fg="yellow",
-                err=True,
+            msg = (
+                f" Env file {path} will not be loaded, "
+                "module python-dotenv is not installed."
             )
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, but this seems to be mostly moving words around, I don't find the refactor particularly useful. I agree the message could be more precise with regards to the path vs the default names, but that can be done in a much shorter way than what you've submitted. Your new message also has a different format and grammatical mistakes.

Tangential: try to keep the subject line of commit messages to <= 50 characters.

proggga:
> Thanks, but this seems to be mostly moving words around, I don't find the refactor particularly useful. I agree the message could be more precise with regards to the path vs the default names, but that can be done in a much shorter way than what you've submitted. Your new message also has a different format and grammatical mistakes.
> 
> Tangential: try to keep the subject line of commit messages to <= 50 characters.

Okey, what I can do? keep only new message?
Constants bad?
I know how it looks, but this code better then previous one:
- it doesn't create useless new_dir variable
- it add comment, that describe dotenv.load_dotenv behaviour
- add extra warning
- add constants, that reusable

What I can leave? or not

proggga:
Fixed message

davidism:
After thinking about this more, it probably doesn't make sense to be calling this manually from a project if the project isn't going to depend on python-dotenv, so adding a special message for this doesn't make much sense.

proggga:
May be I can remove param 'file', if it isnt used?

PR #3806, opened 2020-10-25
Link SQL injection to python sqlite3 library doc

hvnsweeting:
The replacement is done by the library, link to the
python sqlite3 library document instead of a generic document
as different libraries use different placeholders.


davidism:

PR #3807, opened 2020-10-25
Clarify it is talking about HTML tags

hvnsweeting:
In the page, there are mentions of Jinja2 tags (block start/end tag),
clarify here to avoid confusion.


davidism:

PR #3808, opened 2020-10-25
Use list in methods instead of tuple

hvnsweeting:
To be consistent with the rest of documents.

In API doc, `route` function docstring states that
`methods is a list of methods this rule should be limited to (GET, POST etc.)`

Using list also avoid confusing when using a tuple of 1 element, when new
user tries to remove a method from ('GET', 'POST')
and removed the `,` would make it no longer a tuple.

### DIFF
--- a/docs/tutorial/blog.rst
+++ b/docs/tutorial/blog.rst
@@ -142,7 +142,7 @@ will be redirected to the login page.
 .. code-block:: python
     :caption: ``flaskr/blog.py``
 
-    @bp.route('/create', methods=('GET', 'POST'))
+    @bp.route('/create', methods=['GET', 'POST'])
     @login_required
     def create():
         if request.method == 'POST':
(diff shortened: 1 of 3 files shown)


ThiefMaster:
It doesn't really matter what you use there - you can use a tuple or list just fine. A set would probably work too.
Personally I think using a tuple is nicer (even if it does require a trailing comma in case of just one element).

docs/tutorial/blog.rst
```diff
@@ -320,7 +320,7 @@ to the ``index`` view.
 .. code-block:: python
     :caption: ``flaskr/blog.py``
 
-    @bp.route('/<int:id>/delete', methods=('POST',))
+    @bp.route('/<int:id>/delete', methods=['POST',])
```
ThiefMaster, line 323: If we stick with lists here the trailing comma should go away for sure.

davidism:
Thanks, but I don't think this is necessary. The docs use the word "list" for convenience, but this can be any iterable (and is converted to a `set` internally). There is a specific check for passing a string instead of an iterable. Tuples are used interchangeably with lists in this case.

PR #3811, opened 2020-10-26
create hello world added file

DairoCantillo:
It seems to me that it was necessary to specify the step of creating the hello.py file, because some newbie may be confused, even if the creation of the file is implicit


davidism:
Thanks, however the readme is meant to be a *very* brief example. See the tutorial and docs for full information.

PR #3812, opened 2020-10-26
Add separate and more detailed issue templates

filips123:
Previously, there was just a single issue template with "bug report" format. This was good for reporting bugs, but not so convenient for feature requests.

However, GitHub allows multiple issue templates and 'config.yml' file for configuring them. So I added new issue template for feature requests and links to Stack Overflow and Discord to issue reporting menu. Template for bug reports is mostly the same as the original template, and template for feature requests is based on default GitHub feature request template but modified a bit to be consistent with the original template. This will also automatically label issues either as 'bug' or as 'enhancement' depending on which template is chosen. Links to Stack Overflow and Discord were added in similar way as in Werkzeug and Jinja repositories.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,3 +1,9 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: bug
+---
+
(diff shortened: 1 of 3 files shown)


davidism:
Thanks, however I have a specific format I've written, I just haven't copied them over from another project yet.

PR #3813, opened 2020-10-27
Added link to deploying on Divio how-to guide

evildmp:
The guide shows how to create or migrate a Flask application using
Docker on Divio. It discusses using docker-compose.yml to create
a Dockerised development environment, environment variables for
configuration and a gateway server.


davidism:
Thanks for looking at this. However, I'm not currently adding more links to this section. I'm working on a general rewrite of these docs to emphasize software and configuration over specific services. The current list represents the most common services (at least that I've heard of), but I'll be evaluating that as well.

PR #3814, opened 2020-10-28
JSONEncoder: Add test for objects with __html__ method

lielfr:
Just added a test for the `default` method in JSONEncoder, specifically for objects with `__html__` method.


davidism:
Having JSON tests in a separate file is good, but all the other tests are in `test_helpers.py::TestJSON`. Either move the test back there, or make another commit that moves everything to `test_json.py` instead.

tests/test_json_encoder.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_encoder_html():
+    class ObjectWithHTML:
+        TEST_HTML = "<p>test</p>"
+
+        def __html__(self):
+            return self.TEST_HTML
+
+    encoder = JSONEncoder()
+    assert isinstance(encoder, JSONEncoder)
```
davidism, line 12: This assert doesn't seem necessary.

tests/test_json_encoder.py
```diff
@@ -0,0 +1,15 @@
...
+    encoder_default = encoder.default(ObjectWithHTML())
+    assert isinstance(encoder_default, str)
+    assert encoder_default == ObjectWithHTML().TEST_HTML
```
davidism, line 15: Just use a string, no need for a class attribute. (Also, since it's a class attribute you wouldn't need to make a new instance to get it.)

lielfr:
> Having JSON tests in a separate file is good, but all the other tests are in `test_helpers.py::TestJSON`. Either move the test back there, or make another commit that moves everything to `test_json.py` instead.
Does it also mean moving the tests from `test_json_encoder.py` to `test_json.py`?

davidism:
Yes.

tests/test_json_encoder.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_encoder_html():
+    class ObjectWithHTML:
+        TEST_HTML = "<p>test</p>"
+
+        def __html__(self):
+            return self.TEST_HTML
+
+    encoder = JSONEncoder()
```
davidism, line 11: Use `json.dumps` rather than using the encoder directly, as that's not the typical use case.

lielfr:
Fixed the PR according to your input

tests/test_json.py
```diff
@@ -0,0 +1,444 @@
...
+from werkzeug.http import http_date
+
+import flask
+from flask import Markup
+from flask.json import JSONEncoder
+from flask.json.tag import JSONTag
+from flask.json.tag import TaggedJSONSerializer
+
+
+class TestJSON:
```
davidism, line 19: Having a class is redundant now, these can all be module-level functions.

tests/test_json_tag.py
```diff
@@ -1,85 +0,0 @@
...
-from uuid import uuid4
-
-import pytest
-
-from flask import Markup
-from flask.json.tag import JSONTag
-from flask.json.tag import TaggedJSONSerializer
-
-
-@pytest.mark.parametrize(
```
davidism, line 11: This file shouldn't have been removed, it's testing other behavior.

davidism (approved):

davidism:
Rebased the commits to split up the move and the test.

PR #3816, opened 2020-10-29
Updating external_url_handler example to Python 3

shoeffner:
The external_url_handler in the documentation for `flask.url_for`
example [1] now uses Python 3 compatible `E(V).with_traceback(T)`
instead of the old `raise E, V, T`.

[1]: https://flask.palletsprojects.com/en/1.1.x/api/#flask.url_for

See also
https://docs.python.org/3/library/2to3.html?highlight=raise#2to3fixer-raise


shoeffner:
The "Tests / Version Range" seems to be broken on the master branch already, so I guess it is not affected by this documentation change?

PR #3835, opened 2020-11-17
Update deprecated set-env ci command

plourenco:
I noticed that recent CI test runs were [failing](https://github.com/pallets/flask/actions/runs/366551315) due to the deprecated `set-env` command. 

These changes update the deprecated command according to [this changelog](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,7 +44,7 @@ jobs:
           key: pip|${{ runner.os }}|${{ matrix.python }}|${{ hashFiles('setup.py') }}|${{ hashFiles('requirements/*.txt') }}
       - name: set full Python version in PY env var
         # See https://pre-commit.com/#github-actions-example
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, I got worried when that pre-commit PR failed, hadn't seen that error message yet.

plourenco:
Hi @davidism , it has just come to my attention that I mistakenly committed the incorrect new syntax:
```
run: echo "name=PY::$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_EN
```
should become
```
run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_EN
```

I was about to submit a new commit after testing but unfortunately didn't catch you in time! 
Any chance we can revert this?

PR #3836, opened 2020-11-17
Fix ci environment command syntax

plourenco:
This change fixes the incorrect syntax introduced by #3835. Sorry about the mixup!
It has just come to my attention that I mistakenly committed the incorrect new syntax:
```
run: echo "name=PY::$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
```
should become
```
run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
```


davidism:
I missed this too, thanks for following up.

plourenco:
Thanks @davidism !

PR #3844, opened 2020-11-23
Update cli.rst, repeated word "that" in line 258

diegogarromolina:


davidism:
Thanks for looking at this. However, it's correct as-is, even if it looks weird. It sounds less correct to me when it's removed.

diegogarromolina:
Looking it and reading with care, it's good like it is now. Sorry for the PR :P

PR #3848, opened 2020-12-03
fix explanation of wheel name

jordivandooren:
A small self-explanatory documentation change.


ThiefMaster:
Not sure if going that in-depth is needed, but replacing the "and some tags about the file can install" engrish is a good thing for sure :)

PR #3866, opened 2021-01-06
Update the example in README

greyli:
I don't know if this change is necessary, I just thought that it will be better to keep the example as simple as possible, and make it OS-compatible (the `env` command not available on Windows CMD or Powershell). The new user will learn the details of how to run the application or how the application discovery works in the documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,7 @@ Install and update using `pip`_:
 
 .. code-block:: text
 
-    pip install -U Flask
+    $ pip install -U Flask
 
 
 A Simple Example
@@ -28,6 +28,7 @@ A Simple Example
 
 .. code-block:: python
 
+    # save this as app.py
     from flask import Flask
 
     app = Flask(__name__)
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -20,14 +20,15 @@ Install and update using `pip`_:
...
-    pip install -U Flask
+    $ pip install -U Flask
 
 
 A Simple Example
 ----------------
 
 .. code-block:: python
 
+    # app.py
```
jab, line 31: Maybe worth adding something to this comment to make it more explicit that this is required? Or maybe just something like `# put this in a file named app.py`?

jab (commented):
Just chiming in to say that something like these changes would better match how I've always demo'ed Flask to people who've never seen it before: I always use the filename "app.py" to avoid having to complicate their first impression with the incidentals of the FLASK_APP env var.

README.rst
greyli: Good point. How about using `# save this as app.py`?

greyli: Updated. Since you and David agreed, now I'm going to merge it.

PR #3868, opened 2021-01-08
Update views.rst: fix the routes sample code

berviantoleo:
Update the sample code at `views.rst`, update the methods sample.


davidism:
Thanks, however this isn't required.

PR #3883, opened 2021-01-30
Doc update: Clarify got_request_exception signal docs

### LINKED ISSUE #3851: raising HTTPException does not cause got_request_exception signal to be emitted
Raising a `HTTPException` (for example by calling `abort()`) does not cause registered signal handlers for `got_request_exception` to be called.
https://flask.palletsprojects.com/en/1.1.x/api/#flask.got_request_exception does not state that some Exceptions
are forwarded to the handler and some are not.
Changing this behaviour could potentially break quite a few application, but at least the signal documentation should include this behaviour.

### Expected Behavior
raising a HTTPException in a view function should cause the signal handler to be called.

```python
from flask import Flask, got_request_exception, abort
app = Flask(__name__)

def signal_handler(sender, exception, **extra):
    print("handled %s" % str(type(exception)))

got_request_exception.connect(signal_handler)

@app.route('/1')
def view_1():
    # calling this view prints "handled <class 'RuntimeError'>" to the console
    raise RuntimeError()

@app.route('/2')
def view_2():
    # calling this view prints nothing
    abort(400)
    # raise werkzeug.exception.TooManyRequests
    # e.g. also does nothing
```

### Actual Behavior
the signal handler is not called

### Environment

* Python version: 3.8.6 (also tested with 3.7)
* Flask version: 1.1.2
* Werkzeug version: 1.0.1

tonydelanuez:
`got_request_exception` documentation did not explain that the signal is only sent for unhandled exceptions, not `HTTPException` or other handled exceptions.

closes #3851

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -440,18 +440,32 @@ The following signals exist in Flask:
 
 .. data:: got_request_exception
 
-   This signal is sent when an exception happens during request processing.
-   It is sent *before* the standard exception handling kicks in and even
-   in debug mode, where no exception handling happens.  The exception
-   itself is passed to the subscriber as `exception`.
+    This signal is sent when an unhandled exception happens during
+    request processing, including when debugging. The exception is
+    passed to the subscriber as ``exception``.
 
-   Example subscriber::
+    This signal is not sent for
+    :exc:`~werkzeug.exceptions.HTTPException`, or other exceptions that
+    have error handlers registered, unless the exception was raised from
+    an error handler.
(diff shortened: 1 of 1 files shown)


davidism:
The existing section was already bit confusing, so I did a further rewrite based on your changes. It describes that it's sent for unhandled exceptions, and not sent for `HTTPExcpetion` and handled exceptions. Also rewrote the example to be more interesting, since exceptions are already logged anyway.

PR #3897, opened 2021-02-02
Added multi-line imports to app.py file

manuelIsCoding:
I added multi-line imports because it is more pythonic and clean when you want to find some dependency.


ThiefMaster:
Since our CI formatter actually reverts your change, I'm quite sure that the current style is intentional...

PR #3907, opened 2021-02-14
Add syntatic sugar for route registration

pgjones:
I know we briefly discussed this, but I'd like to make the case with a concrete implementation. My argument is that this is actually quite a nice API and requires little implementation (especially now with the nice Scaffold). There is also precedent, in my view, that the test client also has this API. (Also, everyone copies Flask so lets copy them :p). 

**Commit message**

This takes a popular API whereby instead of passing the HTTP method as
an argument to route it is instead used as the method name i.e.

    @app.route("/", methods=["POST"])

is now writeable as,

    @app.post("/")

This is simply syntatic sugar, it doesn't do anything else, but makes
it slightly easier for users.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,6 +64,9 @@ Unreleased
     This could allow a session interface to change behavior based on
     ``request.endpoint``. :issue:`3776`
 -   Use Jinja's implementation of the ``|tojson`` filter. :issue:`3881`
+-   Add route decorators for common HTTP methods. For example,
+    ``@app.post("/login")`` is a shortcut for
+    ``@app.route("/login", methods=["POST"])``. :pr:`3907`
 
 
 Version 1.1.2
--- a/src/flask/scaffold.py
+++ b/src/flask/scaffold.py
@@ -146,6 +146,47 @@ class Scaffold(_PackageBoundObject):
     def _is_setup_finished(self):
         raise NotImplementedError
 
+    def _method_route(self, method, rule, options):
+        if "methods" in options:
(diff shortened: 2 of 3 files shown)


ThiefMaster:
What I dislike with this approach is that - at least when you aren't writing a RESTful API - there is a very good chance that you need endpoints handling both GET and POST (render form if not submitted or validation failed, otherwise do stuff). In fact, in a classic webapp with lots of forms it's probably the most common thing after GET endpoints...

However, this case is not covered by the shortcuts and I can't think of any decent name for a method that does cover it... And I think beginners may think mixed-method routes are discouraged because of this, making them do weird stuff with the session for the basic web form usecase dscribed above...

pgjones:
I believe the test failures are unrelated.

I think the existing `route` method covers the GET & POST case you describe really well. There is (and will still be) a large corpus of tutorials and documentation for beginners that uses `route`, so I don't think we need worry about them going any further astray. As a counter I think it will make it easier for beginners writing RESTful APIs with this syntax.

recepsirin (approved):

davidism:
@recepsirin please stop adding reviews, it generates notifications for the actual maintainers and isn't helping.

davidism:
Refactored to avoid repeating the error message in each method. Changed to a `TypeError` since that's what's used for issues with arguments. Added tests for all methods and for the error raised.

greyli:
Glad to see this finally merged! After I proposed this idea and was declined in the first pallets meeting, I started to make the REST API extension I'm developing to become a framework so that I can inherit the `Flask` class and add these route shortcuts, now I may consider making it back to an extension...

PR #3910, opened 2021-02-16
Fix CLI for latest Werkzeug (2.0.0rc1)

### LINKED ISSUE #3909: CLI incompatibility with Werkzeug 2 LocalProxy
LocalProxy has been changed in Werkzeug 2:

> LocalProxy matches the current Python data model special methods, including all r-ops, in-place ops, and async. __class__ is proxied, so the proxy will look like the object in more cases, including isinstance. Use issubclass(type(obj), LocalProxy) to check if an object is actually a proxy. :issue:`1754`

The Flask CLI can fail to run if the following conditions are met:

1. A `LocalProxy` is defined that requires an app context.
2. `FLASK_APP` is not defined.
3. The app is not named `app` or `application` or an app factory is used.

A simple `app.py` to reproduce:
```
from flask import current_app, Flask
from werkzeug.local import LocalProxy

log = LocalProxy(lambda: current_app.logger)

def create_app():
    app = Flask(__name__)
    return app
```

Then, running `flask shell` should drop nicely into the console. Instead, the result is `RuntimeError: Working outside of application context.`

Environment:

- Python version: 3.9
- Flask version: 1.1.2
- **Werkzeug version: 2.0.0rc1**

The issue is with this line: 
https://github.com/pallets/flask/blob/adeaf27e767f7bd37996d999eba135d63e890408/src/flask/cli.py#L51

and the aforementioned "the proxy will look like the object in more cases, including isinstance."

We can wrap the `isinstance` check in a try/except to ignore failing objects.

ziddey:
Wrap the `isinstance` check in a try/except to ignore failing objects (LocalProxy outside app context).

"LocalProxy matches the current Python data model special methods, including all r-ops, in-place ops, and async. class is proxied, so the proxy will look like the object in more cases, including isinstance. Use issubclass(type(obj), LocalProxy) to check if an object is actually a proxy. :issue:1754"

- fixes #3909 

Checklist:

- [X] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -48,7 +48,13 @@ def find_best_app(script_info, module):
             return app
 
     # Otherwise find the only object that is a Flask instance.
-    matches = [v for v in module.__dict__.values() if isinstance(v, Flask)]
+    matches = []
+    for v in module.__dict__.values():
+        try:
+            if isinstance(v, Flask):
+                matches.append(v)
+        except Exception:
+            pass
 
     if len(matches) == 1:
         return matches[0]
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -44,6 +44,8 @@ def test_cli_name(test_apps):
 
 
 def test_find_best_app(test_apps):
+    from werkzeug.local import LocalProxy
+
     script_info = ScriptInfo()
 
     class Module:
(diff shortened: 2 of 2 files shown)


davidism:
I think we came up with a better solution to this by telling `LocalProxy` about its type even when unbound. I'll leave this open until I make sure that works.

src/flask/cli.py
```diff
@@ -48,7 +48,13 @@ def find_best_app(script_info, module):
...
 
     # Otherwise find the only object that is a Flask instance.
-    matches = [v for v in module.__dict__.values() if isinstance(v, Flask)]
+    matches = []
+    for v in module.__dict__.values():
+        try:
+            if isinstance(v, Flask):
+                matches.append(v)
+        except Exception:
+            pass
```
recepsirin, line 57: Please, handle the possible exceptions because unless you handle them both explicitly or implicitly are meaningless.

recepsirin (changes requested):
Please, handle the possible exceptions because unless you handle them both explicitly or implicitly are meaningless.

davidism:
Fixed in https://github.com/pallets/werkzeug/pull/2042, specifically 352cd7167984f805daf226966f38f8dfeb991313.

PR #3917, opened 2021-02-20
update json.dumps for http_date changes

greyli:
Update for https://github.com/pallets/werkzeug/pull/2040

### DIFF
--- a/src/flask/json/__init__.py
+++ b/src/flask/json/__init__.py
@@ -3,7 +3,6 @@ import json as _json
 import uuid
 import warnings
 from datetime import date
-from datetime import datetime
 
 from jinja2.utils import htmlsafe_json_dumps as _jinja_htmlsafe_dumps
 from werkzeug.http import http_date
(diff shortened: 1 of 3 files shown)


pgjones (approved):

recepsirin (approved):

davidism:
We should probably add support for `date` to `http_date`.

greyli:
@davidism Is it OK to update `http_date` to this:

```py
def http_date(timestamp: t.Optional[t.Union[datetime, int, float]] = None) -> str:
    if isinstance(timestamp, datetime):
        timestamp = _dt_as_utc(timestamp)
        return email.utils.format_datetime(timestamp, usegmt=True)

    if isinstance(timestamp, date):
        timestamp = mktime(timestamp.timetuple())

    return email.utils.formatdate(timestamp, usegmt=True)
```

I'm a bit confused with `time.mktime` and `calander.timegm`, which one should be used to get the timestamp?

davidism:
I added back support in `http_date` for `struct_time`, but also added support for `date`.

PR #3918, opened 2021-02-21
Utilise defaultdicts

pgjones:
This code originates from the Python 2.4 supporting version of Flask,
with defaultdicts being added in 2.5. Using defaultdict makes the
intentional usage clearer, and slightly simplifies the code.

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -430,13 +430,13 @@ class Flask(Scaffold):
         #: .. versionadded:: 0.11
         self.shell_context_processors = []
 
-        #: all the attached blueprints in a dictionary by name.  Blueprints
-        #: can be attached multiple times so this dictionary does not tell
-        #: you how often they got attached.
+        #: Maps registered blueprint names to blueprint objects. The
+        #: dict retains the order the blueprints were registered in.
+        #: Blueprints can be registered multiple times, this dict does
+        #: not track how often they were attached.
         #:
         #: .. versionadded:: 0.7
         self.blueprints = {}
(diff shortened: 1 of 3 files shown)


recepsirin (approved):

src/flask/app.py
```diff
@@ -435,8 +436,7 @@ def __init__(
         #: you how often they got attached.
         #:
         #: .. versionadded:: 0.7
-        self.blueprints = {}
-        self._blueprint_order = []
+        self.blueprints = OrderedDict()
```
davidism, line 439: I think we can assume dicts are ordered for all Pythons we support. Dicts are ordered in PyPy 3.6 and CPython 3.6 as part of the implementation, and this became part of the language in 3.7. Neither Jython nor IronPython support Python 3.

davidism: The commit message says you're keeping `OrderedDict` to indicate that order matters, but given that it's a fairly internal API and we're not taking advantage of any of the differences enumerated in the docs, I don't think it's worth it right now. We can call out in the docstring that the registration order is retained.

src/flask/scaffold.py
```diff
@@ -375,7 +378,7 @@ def _register_error_handler(self, key, code_or_exception, f):
                 " instead."
             )
 
-        handlers = self.error_handler_spec.setdefault(key, {}).setdefault(code, {})
+        handlers = self.error_handler_spec[key].setdefault(code, {})
```
davidism, line 381: The second `setdefault` can also be replaced.

PR #3920, opened 2021-02-22
Small typo correction

tonysepia:
I understand that it's just a vague example and probably isn't meant to be perfect, but the conditional won't hit if user_id is a mandatory argument


davidism:
Look at the `add_url_rule` line to see how this can be `None`. Mandatory argument isn't the same as `not None`.

tonysepia:
Thank you for explaining. I've tried to recreate in a separate app and can now see that all works correctly

PR #3924, opened 2021-02-28
One more attribute required in form file input field

CosmicSage:
without the attribute -- name=file -- being present on the file input element in the form, the request object's files attribute was an empty immutablemultidict. The uploaded file was not being found and consequently the "no file part" message was being flashed.

With the addition of name attribute in the template/html form the upload worked as intended.

Thus I suggest this change in the docs

Thank you

- fixes #<issue number>

Checklist:


davidism:
Thanks, but this is just a description of the full code below, which contains all the necessary parts.

CosmicSage:
> Thanks, but this is just a description of the full code below, which contains all the necessary parts.

Yes but it might waste some individuals time like it did with me. Internally I was like "I don't need to read some demo html" thus I had to lookup other people's implementation etc.

That sounds reasonable right?? @davidism

PR #3932, opened 2021-03-10
Set default encoding to UTF-8 for load_dotenv

### LINKED ISSUE #3931: Default encoding of .env and .flaskenv
I maintain python-dotenv, which Flask uses for loading .env and .flaskenv files, and we've received reports from users confused about the default encoding used by our library: https://github.com/theskumar/python-dotenv/issues/300, which brings me here because it also involves Flask.

Currently, python-dotenv uses the system's preferred encoding to load and store .env files.  Since Flask uses `dotenv.load_dotenv` without the `encoding` parameter, it inherits that behavior. This means that the encoding of a user's `.env` or `.flaskenv` depends on their platform, if I understand Flask's code correctly.

Hence my suggestion: What about making Flask use UTF-8 encoding with the `encoding="utf-8"` argument?

I don't know if the current behavior was deliberately implemented, and you might want to keep it anyway, but I suspect that it wasn't.

I'm considering making UTF-8 the default in python-dotenv as well, but I'd like your opinion on this since Flask has so many users and relies on python-dotenv's current behavior with respect to encoding.

Here's what it would likely imply for us, and probably also for Flask if it also chooses to use UTF-8:

- Pro: It would facilitate sharing .env files between projects since the default encoding would be independent of the platform.  Those files aren't really meant to be committed in a repository but it's quite typical for developers to have a sample .env in their repository.
- Pro: It would make .env files use the same encoding, by default, as those of Python source files.
- Con: It would be a breaking change, but only for users with a non-UTF-8 system encoding and non-UTF-8 strings in their .env files.
- Con: It would diverge from Python's handling of the `encoding` parameter in functions like [`open`](https://docs.python.org/3/library/functions.html#open).

Does that make sense?  I've only started thinking about this and I'm not an expert with regards to Python's handling of encodings so I may well have missed an important point.

greyli:
This PR will set encoding to UTF-8 when calling `dotenv.load_dotenv()`.

Is it necessary to add a `encoding` argument for `cli.load_dotenv`?

fixes #3931 

Checklist:

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -619,6 +619,9 @@ def load_dotenv(path=None):
         Returns ``False`` when python-dotenv is not installed, or when
         the given path isn't a file.
 
+    .. versionchanged:: 2.0
+        When loading the env files, set the default encoding to UTF-8.
+
     .. versionadded:: 1.0
     """
     if dotenv is None:
@@ -636,7 +639,7 @@ def load_dotenv(path=None):
     # else False
     if path is not None:
         if os.path.isfile(path):
-            return dotenv.load_dotenv(path)
+            return dotenv.load_dotenv(path, encoding="utf-8")
 
         return False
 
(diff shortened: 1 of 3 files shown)


greyli:
@davidism Is it needed to update the changelog and the cli docs for this change?  (I was just about to push the commit...)

greyli:
No worries, I also think it's not quite necessary, just tried to fill the checkboxs :p

PR #3940, opened 2021-03-16
Fixture url updated

GlennToms:
URL no longer works.  New link = https://docs.pytest.org/en/latest/reference/reference.html#pytest.fixture


davidism:
Thanks, but this whole page is getting rewritten so I'm going to close this since it will just be replaced.

PR #3942, opened 2021-03-21
shell autocomplete

### LINKED ISSUE #3941: Add auto-completion to shell command
in often times we use the shell command to debug or ensure something, but there's a small problem which is we don't have auto-complete when we press the tab.
so I decided to make a pull request to add this feature.

The auto-complete it's very helpful to speed our development.

awbx:
in often times we use the shell command to debug or ensure something, but there's a small problem which is we don't have auto-complete when we press the tab.
so I decided to make a pull request to add this feature

- fixes #3941

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,6 +67,7 @@ Unreleased
 -   Add route decorators for common HTTP methods. For example,
     ``@app.post("/login")`` is a shortcut for
     ``@app.route("/login", methods=["POST"])``. :pr:`3907`
+-   Add autocompletion to shell command.
 
 
 Version 1.1.2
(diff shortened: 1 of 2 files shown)


davidism:
Readline isn't guaranteed to be installed, so you need to account for that.

awbx:
I think the readline it's a built-in module?

davidism:
Yes, but it's only available on some platforms, and depends on how Python was compiled.

awbx:
I wish you add this feature if possible, it's very helpful in our development env.
Regards, Abdelhadi

davidism:
Ok, but you need to address the comment I made. Paying that your want me to merge it isn't really helpful on its own.

awbx:
@davidism, I'm not talking to merge my own, but wish if you add this feature, it's very helpful.

PR #3957, opened 2021-04-11
Bugfix iscoroutinefunction with Python3.7

pgjones:
See this Python bug https://bugs.python.org/issue33261. The
iscoroutinefunction doesn't recognise partially wrapped coroutine
functions as coroutine functions - which is problematic as the
coroutines will be called as if they are sync, which results in
un-awaited coroutines.

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1,8 +1,9 @@
+import functools
+import inspect
 import os
 import sys
 import weakref
 from datetime import timedelta
-from inspect import iscoroutinefunction
 from itertools import chain
 from threading import Lock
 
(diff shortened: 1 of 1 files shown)


src/flask/_compat.py
```diff
@@ -0,0 +1,19 @@
+import asyncio
+import functools
+import inspect
+import sys
+
+
+def iscoroutinefunction(func):
```
davidism, line 7: I don't want to bring back the `_compat` module, I killed that when dropping Python 2. :-)

Instead, I would this after the imports in `app.py`:

```python
import inspect

if sys.version_info >= (3, 8):
    iscoroutinefunction = inspect.iscoroutinefunction
else:
    def iscoroutinefunction(func):
        while inspect.ismethod(func):
            func = func.__func__

        while isinstance(func, functools.partial):
            func = func.func

        return inspect.iscoroutinefunction(func)
```

src/flask/_compat.py
```diff
@@ -0,0 +1,19 @@
...
+    # Python < 3.8 does not correctly determine partially wrapped
+    # coroutine functions are coroutine functions, hence the need for
+    # this to exist. Code taken from CPython.
+    if sys.version_info >= (3, 8):
+        return asyncio.iscoroutinefunction(func)
```
davidism, line 12: Should be `inspect.iscoroutinefunction` still. Same with below.

pgjones:
Changes made.

PR #3967, opened 2021-04-17
Async Docs

pgjones:
See commits


docs/async-await.rst
```diff
@@ -39,6 +39,21 @@ most use cases, but Flask's async support enables writing and using
...
 
 
+Background tasks
+----------------
+
+Async functions will run in an event loop until they complete, at
+which stage the event loop will stop. This means any additional
+spawned tasks that haven't completed when the async function completes
+will be cancelled. Therefore you cannot spawn background tasks, for
+example via ``asyncio.ensure_future``.
```
graingert, line 49:
```suggestion
example via ``asyncio.create_task``
```

docs/async-await.rst
```diff
@@ -39,6 +39,21 @@ most use cases, but Flask's async support enables writing and using
 code that wasn't possible natively before.
 
 
+Background tasks
+----------------
+
+Async functions will run in an event loop until they complete, at
+which stage the event loop will stop. This means any additional
+spawned tasks that haven't completed when the async function completes
+will be cancelled. Therefore you cannot spawn background tasks, for
```
graingert, line 48: Will they be cancelled or just stopped?

graingert: ah yes they are cancelled https://github.com/django/asgiref/blob/main/asgiref/sync.py#L240

docs/async-await.rst
pgjones: Ta

jab:
Thanks for the great docs!

I tried to preview them rendered as html, but the readthedocs.org CI failed, looks like due to the following?
```
WARNING: autodoc: failed to import class 'Flask' from module 'flask'; the following exception was raised:
cannot import name 'ContextVar' from 'werkzeug.local' (/home/docs/checkouts/readthedocs.org/user_builds/flask/envs/3967/lib/python3.7/site-packages/werkzeug/local.py)
```

davidism:
It should be fixed because it depends on Werkzeug rc now, maybe I need to clear the cache on rtd. It should build locally though

jab:
Thanks @davidism, confirmed the docs build just fine for me locally. And looks like other builds are working just fine on rtd too (see e.g. https://flask.palletsprojects.com/en/master/async-await/). I can submit [a trivial docs PR](https://github.com/pallets/flask/compare/master...jab:docs?expand=1) if it'd help to confirm that the rtd PR docs builds are working again.

PR #3969, opened 2021-04-21
Update README.rst

arihant2math:
Enforce forward slash at the end of all urls.


davidism:
Thanks, but this is not necessary.

PR #3970, opened 2021-04-23
Update old Read Me

Yourun-proger:
I wrote via Markdown, added badges and a link in case pip is not installed. I want to change all .rst to .md


ThiefMaster:
Thanks, but why do you think we are looking for someone to rewrite our readme in markdown?

Also, besides the fact that we do not need this change: changing both content and format in the same commit makes this impossible to review properly since there is no useful diff.

Yourun-proger:
Okey

PR #3972, opened 2021-04-24
Allow using Click 7 with a DeprecationWarning

ThiefMaster:
As long as popular libraries (e.g. Celery) require click 7, depending on Click 8 in Flask makes it hard to test the latest version (and its other dependencies) in existing applications.

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -979,6 +979,13 @@ debug mode.
 
 
 def main() -> None:
+    if int(click.__version__[0]) < 8:
+        warnings.warn(
+            "Using the `flask` cli with Click 7 is deprecated and"
+            " will not be supported starting with Flask 2.1."
+            " Please upgrade to Click 8 as soon as possible.",
+            DeprecationWarning,
+        )
     # TODO omit sys.argv once https://github.com/pallets/click/issues/536 is fixed
     cli.main(args=sys.argv[1:])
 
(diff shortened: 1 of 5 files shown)


src/flask/cli.py
```diff
@@ -976,6 +976,12 @@ def routes_command(sort, all_methods):
 
 
 def main():
+    if int(click.__version__[0]) < 8:
+        warnings.warn(
+            "Using the `flask` cli with Click 7 is deprecated and"
+            " will not be supported starting with Flask 2.1.",
```
davidism, line 985: Should add something about upgrading to Click 8.

tox.ini
```diff
@@ -14,7 +15,9 @@ deps =
     https://github.com/pallets/markupsafe/archive/master.tar.gz
     https://github.com/pallets/jinja/archive/master.tar.gz
     https://github.com/pallets/itsdangerous/archive/master.tar.gz
-    https://github.com/pallets/click/archive/master.tar.gz
+
+    click8: https://github.com/pallets/click/archive/master.tar.gz
```
davidism, line 19: Remove the click8 tag so that `tox -e py` still works.

ThiefMaster: I didn't know `!click7` works (I need to specify something or pip will choke on the dependency conflict) since the documentation on those markers is VERY poor... but apparently it works just fine so I'm using that now.

src/flask/cli.py
ThiefMaster: done

davidism (approved):

PR #3977, opened 2021-04-28
Update docs changelog URL to "/changes"

greyli:
[This commit](https://github.com/pallets/flask/commit/9e7d3a6b694b27f997266889ae463e06dd83a189) changes the URL of the changelog page from `/changelog` to `/changes`, this will break the URL in the [version warning banner](https://flask.palletsprojects.com/en/master/changes/#version-2-0-0) for non-stable versions. This PR will update the URL in old docs. It will be ideal to add a redirect entry on [readthedocs's redirects setting page](https://readthedocs.org/dashboard/flask/redirects/).

PR #3978, opened 2021-04-28
Update docs changelog URL to "/changes" for 1.0.x

greyli:
See #3977 for the details.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,4 +1,4 @@
-Changelog
-=========
+Changes
+=======
 
 .. include:: ../CHANGES.rst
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I'm not sure if updating the old branches helps - AFAIK the RTD builds are based on tags, so unless we tag a new version the docs won't be updated.

Maybe redirecting from the old to the new changelog url would be better (also for not breaking links on Google - it's REALLY annoying if you google for "$project changelog" and get a 404!).

greyli:
> Maybe redirecting from the old to the new changelog url would be better

Ah, I can't believe I didn't think of this, it's absolutely a better solution.

PR #3982, opened 2021-04-28
mini update

Yourun-proger:
i added information


pgjones:
Thanks, but I don't think we need to highlight this in the quickstart (it is mentioned elsewhere in the docs).

PR #3989, opened 2021-05-02
Async improvements

pgjones:

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -35,12 +35,12 @@ from .globals import _request_ctx_stack
 from .globals import g
 from .globals import request
 from .globals import session
+from .helpers import async_to_sync
 from .helpers import get_debug_flag
 from .helpers import get_env
 from .helpers import get_flashed_messages
 from .helpers import get_load_dotenv
 from .helpers import locked_cached_property
-from .helpers import run_async
 from .helpers import url_for
 from .json import jsonify
 from .logging import create_logger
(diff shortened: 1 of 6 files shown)


src/flask/app.py
```diff
@@ -1581,7 +1579,7 @@ def ensure_sync(self, func: t.Callable) -> t.Callable:
         .. versionadded:: 2.0
         """
         if iscoroutinefunction(func):
-            return run_async(func)
+            return async_to_sync(func)
```
jab, line 1582: It still seems like the `async_to_sync` implementation is the only thing users will want to customize, rather than this entire `ensure_sync` method. The current design still requires users to duplicate this `iscoroutinefunction(...)` boilerplate. But it would be just as easy to parameterize the `async_to_sync` function instead, no?

pgjones: Done, added `async_to_sync` as an overrideable method.

src/flask/app.py
```diff
@@ -1581,10 +1579,41 @@ def ensure_sync(self, func: t.Callable) -> t.Callable:
...
         return func
 
+    def async_to_sync(
+        self, func: t.Callable[..., t.Coroutine]
+    ) -> t.Callable[..., t.Any]:
+        """Return a sync function that will run the coroutine function *func*.
+
+        This can be used as so,::
+
+            result = app.async_to_async(func)(*args, **kwargs)
```
mozartilize, line 1593:
```suggestion
            result = app.async_to_sync(func)(*args, **kwargs)
```
typo here

PR #3997, opened 2021-05-05
info about running flask app

Yourun-proger:


greyli:
Thank you for working on this, but the section under is already addressed this:

> If `FLASK_APP` is not set, the command will try to import “app” or “wsgi” (as a “.py” file, or package) and try to detect an application instance or factory.

PR #4006, opened 2021-05-11
Change misleading warning

anitakar:
"This is a development server. Do not use it in a production deployment." was shown if the env was set to production instead of development.

- fixes #<issue number>

Checklist:


anitakar:
Sorry. My fault. The message is correct. This is a warning that you are using Flask without WSGI container. I saw this message in Airflow that uses gunicorn to serve web server. But the message came from a side log server that did not use WSGI server.
Sorry for engaging you.

PR #4016, opened 2021-05-12
Fix typo in the example of nesting bp docs

pgjones:


greyli:
I think we could cherry-pick commits from the `.x` branch to the `main` branch instead of creating new PR.

davidism:
I usually do `git merge origin/2.0.x` and push it directly instead of making a PR, but you'll probably hit branch protection. Don't worry about immediately syncronizing main with 2.0.x, I can do that periodically. Maybe we need a backport/merge bot like CPython has.

PR #4018, opened 2021-05-12
Update changelog for send_from_directory change

greyli:
The parameter `filename` has been renamed to `path` (accidentally?).


davidism:
Closing in favor of #4019

greyli:
I think we still need to add the `versionchanged` entry in #4019 and update the change entry in 2.0.1 to something like:

- Re-add the `filename` parameter in `send_from_directory`. The `filename` parameter has been renamed to `path`, the old name is deprecated.

What do you think?

PR #4019, opened 2021-05-12
Re-add filename parameter for send_from_directory

greyli:
Re-add the `filename` parameter with a deprecation warning, then we can remove it in 2.1.0.

### DIFF
--- a/src/flask/helpers.py
+++ b/src/flask/helpers.py
@@ -642,7 +642,9 @@ def safe_join(directory: str, *pathnames: str) -> str:
     return path
 
 
-def send_from_directory(directory: str, path: str, **kwargs: t.Any) -> "Response":
+def send_from_directory(
+    directory: str, path: str, filename: t.Optional[str] = None, **kwargs: t.Any
+) -> "Response":
     """Send a file from within a directory using :func:`send_file`.
 
     .. code-block:: python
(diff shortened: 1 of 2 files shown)


src/flask/helpers.py
```diff
@@ -642,7 +642,9 @@ def safe_join(directory: str, *pathnames: str) -> str:
     return path
 
 
-def send_from_directory(directory: str, path: str, **kwargs: t.Any) -> "Response":
+def send_from_directory(
+    directory: str, path: str, filename: str, **kwargs: t.Any
```
ThiefMaster, line 646: This is missing a `= None`

greyli: Thanks!

src/flask/helpers.py
```diff
@@ -666,12 +668,24 @@ def download_file(name):
...
 
     .. versionadded:: 0.5
     """
+    if filename is not None:
+        warnings.warn(
+            "The 'filename' parameter has been renamed to"
+            " 'path'. The old name will be removed in Flask"
+            " 2.1.",
+            DeprecationWarning,
+            stacklevel=3,
```
davidism, line 686: Was `stacklevel=3` intentional? Seems like it should be `2`.

davidism: I changed it to 2.

greyli: In the beginning, I tried to write a wrapper function like the one for `send_file`, then find it is unnecessary, so I remove the wrapper, but forgot to update the `stacklevel`.

PR #4021, opened 2021-05-12
Added Google Cloud Run as a Hosted option

andrewjroth:
Minor Document update to add Google Cloud Run as another (possibly lower cost and more portable) alternative to App Engine on Google Cloud Platform.


davidism:
Thanks, I'll merge this since we already mention Google Cloud. However, in general I close PRs adding more services to this list, as it's not possible for me to evaluate them. In the future we may make this list require a sponsorship/donation to add to.

PR #4026, opened 2021-05-12
Improve type hints for send_file and send_from_directory

### LINKED ISSUE #4044: fix typing for get_send_file_max_age
`get_send_file_max_age` accepts an optional string and returns an optional int. See https://github.com/pallets/werkzeug/pull/2121.

greyli:
- Add missing type hints for send_file params.
- Improve type hints for send_from_directory params.

fixes #4044


anitakar:
Sorry. My fault. The message is correct. This is a warning that you are using Flask without WSGI container. I saw this message in Airflow that uses gunicorn to serve web server. But the message came from a side log server.
Sorry for engaging you.
Closing the item.

PR #4028, opened 2021-05-12
Update env to venv for consistency with the rest docs

Yuxiaoy1:


ThiefMaster:
What about `.venv`?

davidism:
I've been using the name `venv`, as it's more visible especially for new contributors.

davidism:
Wait, this is already consistent with the other contributing guides, they use "env" too.

PR #4029, opened 2021-05-12
Update env to venv for consistency with the rest docs

Yuxiaoy1:
`venv` is used in `docs/cli.rst`, `docs/installation.rst`, the tutorial part and example part.

This PR makes it consistent with the rest Flask docs.

Supplement #4028

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -102,15 +102,15 @@ First time setup
 
           .. code-block:: text
 
-             $ python3 -m venv env
-             $ . env/bin/activate
+             $ python3 -m venv venv
+             $ . venv/bin/activate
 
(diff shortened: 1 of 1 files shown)


davidism:
This is the same PR, closing for the same reason.

davidism:
Anyway, thanks for your interest, but this isn't a necessary change, the name of the venv isn't very important.

Yuxiaoy1:
Sorry I'm a little confused here, but why don't we make the `venv` name consistent in `CONTRIBUTING.rst` and the `docs/cli.rst`, `docs/installation.rst` file?

Yuxiaoy1:
@davidism If you have concerns for other pallets projects, I can make PR for them as well.

PR #4036, opened 2021-05-13
Fix url_prefix argument for nesting blueprint

### LINKED ISSUE #4037: The Blueprint(url_prefix) parameter is not working when using nesting blueprint
Minimal example:

```python
from flask import Flask, Blueprint

app = Flask(__name__)
parent = Blueprint('parent', __name__, url_prefix='/child')
child = Blueprint('child', __name__, url_prefix='/parent')


@child.get('/')
def index():
    return 'hello'


app.register_blueprint(parent)
```

The output of `flask routes`:

```
$ flask routes
Endpoint            Methods  Rule
------------------  -------  -----------------------
parent.child.index  GET      /
static              GET      /static/<path:filename>
```

The URL rule should be `/parent/child/` instead of `/`.

P.S. The URL prefix works fine if set the prefix by using `register_blueprint(url_prefix)`.

Environment:

- Flask version: 2.0

greyli:
fixes #4037

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -260,7 +260,7 @@ class Blueprint(Scaffold):
         """Called by :meth:`Flask.register_blueprint` to register all
         views and callbacks registered on the blueprint with the
         application. Creates a :class:`.BlueprintSetupState` and calls
-        each :meth:`record` callbackwith it.
+        each :meth:`record` callback with it.
 
         :param app: The application this blueprint is being registered
             with.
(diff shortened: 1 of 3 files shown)


tests/test_blueprints.py
```diff
@@ -899,3 +899,51 @@ def grandchild_no():
     assert client.get("/parent/no").data == b"Parent no"
     assert client.get("/parent/child/no").data == b"Parent no"
     assert client.get("/parent/child/grandchild/no").data == b"Grandchild no"
+
+
+def test_nested_blueprint_url_prefix(app, client):
```
pgjones, line 904: This test seems too convoluted for what it tests. Could you simplify it to test the prefix alone - for example using the snippet in #4037?

pgjones (changes requested):

tests/test_blueprints.py
greyli: Sure. Just updated.

pgjones (approved):

src/flask/blueprints.py
```diff
@@ -340,11 +340,10 @@ def extend(bp_dict, parent_dict):
                 app.cli.add_command(self.cli)
 
         for blueprint, bp_options in self._blueprints:
-            url_prefix = options.get("url_prefix", "")
-            if "url_prefix" in bp_options:
-                url_prefix = (
-                    url_prefix.rstrip("/") + "/" + bp_options["url_prefix"].lstrip("/")
-                )
+            url_prefix = options.get("url_prefix", self.url_prefix)
```
pgjones, line 343: I think this should be,
```python
url_prefix = options.get("url_prefix") or self.url_prefix                                                                                                                                       
bp_url_prefix = bp_options.get("url_prefix") or blueprint.url_prefix                                                                                                                            
if bp_url_prefix and url_prefix:                                                                                                                                                                
    url_prefix = url_prefix.rstrip("/") + "/" + bp_url_prefix.lstrip("/")
```
As it is possible for `url_prefix` to be in the options with a value of `None` and the net url_prefix to also be `None`. What do you think? (This is my [fix](https://gitlab.com/pgjones/quart/-/commit/29e4157583fd10bd504346fc3f319908339acc8d) for Quart).

davidism: That makes sense, currently rebasing this PR, I'll make that change.

davidism:
@pgjones I noticed while working on this that we're registering blueprints differently in blueprints vs apps. I would expect `Blueprint.register_blueprint` to immediately register all its stuff on the parent blueprint, just like `Flask.register_blueprint` works. I think we should look into refactoring that later.

PR #4042, opened 2021-05-13
Update typing

### LINKED ISSUE #4020: Assigning to attribute flask.g.[...] not defined in class slots (assigning-non-slot)
Thanks for a recent release of `flask` which looks great. :tada: Just started testing it out.

This is not directly a bug afaik, but our CI started failing on the new release of `flask` and `werkzeug`.

Following the flask documentation, it appears the recommended way of setting new attributes to `flask.g` is simply doing `flask.g.some_attribute = some_value`.

This example code
```python
from flask import g

g.some_attribute = 42
```
gives no `pylint` warning when used together with `pip install "flask<2" "werkzeug<2"`, however after `pip install --upgrade flask werkzeug` the same code gives the following `pylint` warning:
```
test_pylint.py:3:0: E0237: Assigning to attribute 'some_attribute' not defined in class slots (assigning-non-slot)
```

Any recommended code change that should be done related to setting attributes on the namespace object `flask.g`, or would the best recommendation be to silence `pylint` when used with newest `flask` + `werkzeug`?

### LINKED ISSUE #4024: Flask 2.0 uses implicit reexports which mypy complains about
Now that flask 2.0.0 has a `py.typed` file, the flask types are preferred over the typeshed hints. When running mypy on my code, I now see errors like the following:

```
 app.py:6: error: Module 'flask' does not explicitly export attribute 'Flask'; implicit reexport disabled
 app.py:6: error: Module 'flask' does not explicitly export attribute 'render_template'; implicit reexport disabled
 app.py:6: error: Module 'flask' does not explicitly export attribute 'request'; implicit reexport disabled
 app.py:6: error: Module 'flask' does not explicitly export attribute 'flash'; implicit reexport disabled
 app.py:6: error: Module 'flask' does not explicitly export attribute 'send_file'; implicit reexport disabled
```

In `flask/__init__.py` there are lines like this:

https://github.com/pallets/flask/blob/d575de5159a6e40944275763c9ada2801214058b/src/flask/__init__.py#L7

However, mypy expects that these reexports should either have an `__all__` to explicitly say which variables are being reexported, or to change the import to `from .app import Flask as Flask`

In typeshed there is an example of how this was done before flask including typing: https://github.com/python/typeshed/blob/6bb1d885c696914e5900329c3d0b6bf2bfefee81/stubs/Flask/flask/__init__.pyi

Environment:

- Python version: 3.9
- Flask version: 2.0.0

### LINKED ISSUE #4040: some types aren't available in 3.6.0
For example, `NoReturn` wasn't added until 3.6.2. Replace any such types with `typing_extensions`. See https://github.com/asottile/flake8-typing-imports/blob/d1f8e340dcc2c22d844e1f5eba3248629185c426/flake8_typing_imports.py#L234-L248 for a list of supported types. Unfortunately, can't use flake8-typing-imports since we alias `import typing as t` and it doesn't understand that.

davidism:
- Mark top-level names as exported. fixes #4024
- Fix type annotation for ``g`` and tell mypy it's a namespace. fixes #4020 
- Fix some types that weren't available in Python 3.6.0. fixes #4040

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Unreleased
     imports in user projects. :issue:`4024`
 -   Fix type annotation for ``g`` and inform mypy that it is a namespace
     object that has arbitrary attributes. :issue:`4020`
+-   Fix some types that weren't available in Python 3.6.0. :issue:`4040`
 
 
 Version 2.0.0
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Unreleased
     is deprecated. :pr:`4019`
 -   Mark top-level names as exported so type checking understands
     imports in user projects. :issue:`4024`
+-   Fix type annotation for ``g`` and inform mypy that it is a namespace
+    object that has arbitrary attributes. :issue:`4020`
 
 
 Version 2.0.0
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Unreleased
 -   Re-add the ``filename`` parameter in ``send_from_directory``. The
     ``filename`` parameter has been renamed to ``path``, the old name
     is deprecated. :pr:`4019`
+-   Mark top-level names as exported so type checking understands
+    imports in user projects. :issue:`4024`
 
 
 Version 2.0.0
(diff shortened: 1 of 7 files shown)


pgjones (approved):

hughsie:
This fixes the CI failure on the LVFS. Thanks!

PR #4049, opened 2021-05-14
Use compat fspath instead of os.fspath in static_folder

### LINKED ISSUE #4050: AttributeError: 'module' object has no attribute 'fspath'
`AttributeError: 'module' object has no attribute 'fspath'` exception is thrown when running `app = flask.Flask(__name__)` for flask v1.13.

Version is Python 2.7 and 3.5. I realize `fspath` is only available for Python version >= 3.6 but there shouldn't be any breaking changes from Flask 1.1.2 => 1.1.3, so the classifiers [here](https://pypi.org/project/Flask/1.1.3/) are actually incorrect.

brettlangdon:
When #3678 was merged it introduced the usage of os.fspath which is not supported on Python <3.6

This PR updates `static_folder` to use the `fspath` from `_compat` and un-skips the test case.

- fixes #4050 

Checklist:

- ~[ ] Add or update relevant docs, in the docs folder and in code.~
- ~[ ] Add `.. versionchanged::` entries in any relevant code docs.~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: flask
 
+Version 1.1.4
+-------------
+
+Unreleased
+
+-   Update ``static_folder`` to use ``_compat.fspath`` instead of
+    ``os.fspath`` to continue supporting Python <3.6 :issue:`4050`
 
 
 Version 1.1.3
--- a/src/flask/__init__.py
+++ b/src/flask/__init__.py
@@ -57,4 +57,4 @@ from .signals import template_rendered
 from .templating import render_template
 from .templating import render_template_string
 
-__version__ = "1.1.3"
+__version__ = "1.1.4.dev0"
--- a/src/flask/helpers.py
+++ b/src/flask/helpers.py
@@ -1001,7 +1001,7 @@ class _PackageBoundObject(object):
     @static_folder.setter
     def static_folder(self, value):
         if value is not None:
-            value = os.fspath(value).rstrip(r"\/")
+            value = fspath(value).rstrip(r"\/")
         self._static_folder = value
 
     @property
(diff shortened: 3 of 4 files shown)


davidism:
We do not support < 3.6 any longer (and neither does CPython).

brettlangdon:
`setup.py` in 1.1.x says `python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"`

https://github.com/pallets/flask/blob/c04b0de558fe8e1ccb8edb4525d40e725ae9a24d/setup.py#L54

Is it documented somewhere that 1.1.x does not support 3.6?

davidism:
We no longer support 1.1.x either, due to the magnitude of the 2.x release. Unfortunately, this snuck in when we made an unplanned release to satisfy people who did not pin their dependencies.

brettlangdon:
Yeah, I understand not wanting to support EOL versions of Python, just seems like oversight that 1.1.3 introduced this breaking change, especially when there is a compat `fspath` available.

This is fine, it is your decision on how you want to support this, I should be able to work around this on my side without too much hassle, I'll just pin our 1.1.x testing to 1.1.2. If there are no more planned released on 1.1.x then should be perfectly fine for our use case.

davidism:
I'll make another release.

I'd highly suggest telling clients who are still using EOL Python versions that they should consider upgrading. This seems appropriate for a monitoring tool that's designed to alert users about issues with their stack. ;-)

brettlangdon:
@davidism thanks, I really appreciate it.

davidism:
1.1.4 is available on PyPI

brettlangdon:
@davidism thank you for the quick turn around time on this.

PR #4058, opened 2021-05-15
Lucas/add head short route

LucasAndrad:
Inspired by this awesome change #3907 I believe the http method HEAD also deserve to be represented as a shortcut.

Checklist:


davidism:
We decided when adding the methods that internal things like head and options were better handled internally, or if really necessary then with the regular `route` decorator. The shortcuts are intended for common methods that most people would be writing APIs with.

PR #4059, opened 2021-05-15
Fix the request blueprint method

pgjones:
Previously for a nested blueprint it would return the parent_name.name
rather than the name of the current blueprint. This is fixed and
another blueprints method is added in order to get the parent_name and
name. The latter is made use of in the app.

I've used a cached_property as this method is called a few times per
request (by the app methods).

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,9 @@ Unreleased
 -   Roll back a change to the order that URL matching was done. The
     URL is again matched after the session is loaded, so the session is
     available in custom URL converters. :issue:`4053`
-
+-   Fix the request ``blueprint`` method to ensure it returns the
+    current blueprint name. A ``blueprints`` method is added to
+    replace any functionality lost in this fix. :pr:`4059`
 
 Version 2.0.0
 -------------
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -747,7 +747,7 @@ class Flask(Scaffold):
         ] = self.template_context_processors[None]
         reqctx = _request_ctx_stack.top
         if reqctx is not None:
-            for bp in self._request_blueprints():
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -23,7 +23,9 @@ Unreleased
 -   Roll back a change to the order that URL matching was done. The
     URL is again matched after the session is loaded, so the session is
     available in custom URL converters. :issue:`4053`
-
+-   Fix the request blueprint method to ensure it returns the current
+    blueprint. A blueprints method is added to replace any
```
ThiefMaster, line 27:
```suggestion
    blueprint. A ``blueprints`` property is added to replace any
```

pgjones: Ta

davidism:
Based on #4069, I think we need some more thought about what a blueprint's name is. Right now `request.blueprint` is correct for what would be needed for #4069: a blueprint's name should be "unique" in that it's the full path it's registered on the app with. That also feels adjacent.

Additionally, this implementation of `blueprints` reverses the names, so that it's in the opposite order of the endpoint name. Perhaps the user-facing property needs to return in the original order, so `request.blueprint == request.blueprints[-1]`, and an `_iter_blueprints` method is available for internal use.

pgjones:
At the moment though the blueprint is registered on the app with its name, which is what `request.blueprint` will return if this is merged i.e. I thought this should continue work `app.blueprints[request.blueprint]` for nested blueprints.

Yea, I'd wondered about the ordering. I thought that given this method (and `request.blueprint`) mostly exist for flask internal usage it was ok to default this way. Happy to switch though.

pgjones:
An alternative will be proposed, based on recent issue reports.

Issue #4062, opened 2021-05-16
Url prefix duplicating when instanticiating multiple apps with nested blueprints


reritom:
When using an app factory approach to create an app that has a blueprint with a url prefix, and a nested blueprint, the more times the app factory is called, the more times the parent url_prefix gets appended.

```
from flask import Flask, jsonify, Blueprint
from flask.views import MethodView


# Create a child blueprint with a simple view
class UsersView(MethodView):
    def get(self):
        return jsonify([]), 200

child_blueprint = Blueprint("child", __name__)
child_blueprint.add_url_rule("/user", view_func=UsersView.as_view("accounts"))

# Create a parent blueprint that registers the child blueprint
parent_blueprint = Blueprint("parent", __name__)
parent_blueprint.register_blueprint(child_blueprint)

# App factory
def create_app():
    app = Flask(__name__)
    app.register_blueprint(parent_blueprint, url_prefix="/api")
    return app

print("First app")
app = create_app()
print(app.url_map)

print("Second app")
app = create_app()
print(app.url_map)

print("Third app")
app = create_app()
print(app.url_map)
```
Output:
```
First app
Map([<Rule '/api/user' (HEAD, GET, OPTIONS) -> parent.child.accounts>,
 <Rule '/static/<filename>' (HEAD, GET, OPTIONS) -> static>])
Second app
Map([<Rule '/api/api/user' (HEAD, GET, OPTIONS) -> parent.child.accounts>,
 <Rule '/static/<filename>' (HEAD, GET, OPTIONS) -> static>])
Third app
Map([<Rule '/api/api/api/user' (HEAD, GET, OPTIONS) -> parent.child.accounts>,
 <Rule '/static/<filename>' (HEAD, GET, OPTIONS) -> static>])
```
In the above output you can see that the url_prefix gets appended an additional time each time the app factory is called

I would expect that the app factory could be called as many times as wanted, with the resulting url_map being the same each time

Environment:

- Python version: 3.8.1
- Flask version: 2.0.0

davidism:
Fixed by #4036

reritom:
Running the same code on branch 2.0.x has a different problem with the output being:
```
First app
Map([<Rule '/user' (HEAD, OPTIONS, GET) -> parent.child.accounts>,
 <Rule '/static/<filename>' (HEAD, OPTIONS, GET) -> static>])
Second app
Map([<Rule '/user' (HEAD, OPTIONS, GET) -> parent.child.accounts>,
[...]

PR #4063, opened 2021-05-16
Bp fix

pgjones:
Note for the Changelog the existing entry will suffice.

See #4062

Checklist:


src/flask/blueprints.py
```diff
@@ -344,17 +344,12 @@ def extend(bp_dict, parent_dict):
                 app.cli.add_command(self.cli)
 
         for blueprint, bp_options in self._blueprints:
-            bp_options = bp_options.copy()
```
pgjones, line 347: Is this required? I'm wasn't sure and the code added works as expected in Quart.

davidism: Yes, otherwise you're mutating the shared state, so each parent registration will affect future registrations.

PR #4070, opened 2021-05-17
blueprints: fix nested blueprint url_prefix

### LINKED ISSUE #4062: Url prefix duplicating when instanticiating multiple apps with nested blueprints

MaicoTimmerman:
The url_prefix would only be propagated if the parent blueprint or overwrite had an url_prefix.

Fixes #4062

Checklist:

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -354,7 +354,9 @@ class Blueprint(Scaffold):
                 bp_options["url_prefix"] = (
                     state.url_prefix.rstrip("/") + "/" + bp_url_prefix.lstrip("/")
                 )
-            else:
+            elif bp_url_prefix is not None:
+                bp_options["url_prefix"] = bp_url_prefix
+            elif state.url_prefix is not None:
                 bp_options["url_prefix"] = state.url_prefix
 
             bp_options["name_prefix"] = options.get("name_prefix", "") + self.name + "."
--- a/tests/test_blueprints.py
+++ b/tests/test_blueprints.py
@@ -870,7 +870,7 @@ def test_nested_blueprint_url_prefix(app, client):
     assert client.get("/parent/child/orange/").data == b"Apple"
 
 
-def test_nested_blueprint_url_prefix_only_parent_prefix(app, client):
+def test_nested_blueprint_url_prefix_only_parent_overwrite_prefix(app, client):
     parent = flask.Blueprint("parent", __name__)
     child = flask.Blueprint("child", __name__)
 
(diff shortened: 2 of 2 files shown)


pgjones (approved):

pgjones:
This needs no additional docs, it is covered under the existing fixes.

davidism:
These tests all feel very similar. Can you refactor this to use `pytest.mark.parametrize`?

MaicoTimmerman:
Sure, will do together with the existing 3 tests

davidism:
@pgjones seems you made this change as part of #4074?

pgjones:
Hopefully all the outstanding blueprint naming issues are fixed in #4074. I pulled this in as well.

PR #4071, opened 2021-05-17
Improve decorator factory typing

### LINKED ISSUE #4060: Untyped decorator makes function untyped
When I run `mypy --strict` on the following file (named `flask_types.py`),

```python
from typing import Tuple

from flask import Flask

app = Flask(__name__)

@app.errorhandler(404)
def page_not_found(error: Exception) -> Tuple[str, int]:
    return str(error), 404
```

I get the following error:

```console
flask_types.py:7: error: Untyped decorator makes function "page_not_found" untyped
```

Instead, mypy should return no errors.

Using `--strict` (or the equivalent option in `setup.cfg`) in your tests might have caught an error like this, but given how new your type annotations are, I understand why you aren't using the setting yet.

The root cause appears to be that the implemented decorators often use `typing.Callable` in type annotations without any type parameters, which causes them to lose type information.

The fix for this specific issue is only a couple of lines (I have a prototype implemented locally), but I will try to add appropriate type parameters for `Callable` in as many places as possible. I expect to have a PR out for this in a couple of days.
Once I got a fix working for `errorhandler()`, mypy was able to find an error in my type annotations of `page_not_found()` as well as some internal issues in Flask itself, which I feel demonstrates how better typing would be useful here.

Environment:

- Python version: 3.7.9
- Flask version: 2.0.0
- mypy version: 0.812

---

PS: I really appreciate the new typing annotations in Flask! They seem like they took a lot of effort, and I have already found them useful. Thank you very much!

aphedges:
I added type parameters to `Callable` types in decorator factories. This required changing the type annotations for `setupmethod()`, which was removing type information from the functions it was called on. I also fixed the signatures for `add_url_rule()` because my changes to `setupmethod()` revealed some errors there.

There is still more room for improvement in usage of `Callable` elsewhere throughout the codebase, but this is at least a step forward.

- fixes #4060

Checklist:

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -387,7 +387,9 @@ class Blueprint(Scaffold):
             )
         )
 
-    def app_template_filter(self, name: t.Optional[str] = None) -> t.Callable:
+    def app_template_filter(
+        self, name: t.Optional[str] = None
+    ) -> t.Callable[[TemplateFilterCallable], TemplateFilterCallable]:
         """Register a custom template filter, available application wide.  Like
         :meth:`Flask.template_filter` but for a blueprint.
 
@@ -417,7 +419,9 @@ class Blueprint(Scaffold):
 
         self.record_once(register_template)
 
-    def app_template_test(self, name: t.Optional[str] = None) -> t.Callable:
+    def app_template_test(
+        self, name: t.Optional[str] = None
+    ) -> t.Callable[[TemplateTestCallable], TemplateTestCallable]:
         """Register a custom template test, available application wide.  Like
         :meth:`Flask.template_test` but for a blueprint.
 
@@ -451,7 +455,9 @@ class Blueprint(Scaffold):
 
         self.record_once(register_template)
 
-    def app_template_global(self, name: t.Optional[str] = None) -> t.Callable:
+    def app_template_global(
+        self, name: t.Optional[str] = None
+    ) -> t.Callable[[TemplateGlobalCallable], TemplateGlobalCallable]:
         """Register a custom template global, available application wide.  Like
         :meth:`Flask.template_global` but for a blueprint.
 
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -365,6 +365,7 @@ class Blueprint(Scaffold):
         rule: str,
         endpoint: t.Optional[str] = None,
         view_func: t.Optional[t.Callable] = None,
+        provide_automatic_options: t.Optional[bool] = None,
         **options: t.Any,
     ) -> None:
         """Like :meth:`Flask.add_url_rule` but for a blueprint.  The endpoint for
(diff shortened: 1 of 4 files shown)


src/flask/blueprints.py
```diff
@@ -365,6 +365,7 @@ def add_url_rule(
         rule: str,
         endpoint: t.Optional[str] = None,
         view_func: t.Optional[t.Callable] = None,
+        provide_automatic_options: t.Optional[bool] = None,
```
pgjones, line 368: `self.record(lambda s: s.add_url_rule(rule, endpoint, view_func, provide_automatic_options=provide_automatic_options, **options))` is also required

pgjones (changes requested):
Looks good bar a potential bug

src/flask/blueprints.py
aphedges: Good catch! I don't know enough about blueprints to know if this is the desired behavior, but it seems right to pass it properly.

It turns out the `add_url_rule()` in the lambda is a different function than the function I had already modified. It seems there is room for improvement in the type annotations for `record()`, but that is too complicated for me to handle.

aphedges: Should anything about these changes be added to the changelog as well?

src/flask/blueprints.py
```diff
@@ -376,9 +379,15 @@ def add_url_rule(
         if view_func and hasattr(view_func, "__name__") and "." in view_func.__name__:
             raise ValueError("'view_func' name may not contain a dot '.' character.")
 
-        self.record(lambda s: s.add_url_rule(rule, endpoint, view_func, **options))
+        self.record(
+            lambda s: s.add_url_rule(
+                rule, endpoint, view_func, provide_automatic_options, **options
```
pgjones, line 384: `provide_automatic_options` was a keyword argument before (via options) and will need to be here as well.

src/flask/blueprints.py
pgjones: I think the changelog is clear enough. 

The record is less concerning (typing wise) as it is perhaps solely used internally by Flask.

src/flask/blueprints.py
aphedges: I took a different approach of adding it as a positional argument, but passing it as a keyword argument seems much safer, given that I'm unfamiliar with this code and that `provide_automatic_options` doesn't seem to be well tested.

I just pushed a commit with the requested change. Should I squash the 3 commits for `add_url_rule()` together to reduce churn? And should I rebase against the branch, or do you not care? `CONTRIBUTING.rst` doesn't really go into the preferred Git workflow besides which branch to target.

pgjones: Please squash the commits - a cleaner history is nice. This is targeting the correct branch, and I should be able to merge without the rebase, but go ahead if you'd like.

aphedges:
I have squashed the `add_url_rule()` commits and rebased against `2.0.x`, fixing a merge conflict in `CHANGES.rst`. Hopefully this PR looks good now?

davidism:
Yes, looks fine, I've just been dealing with other PRs so haven't been able to merge this yet.

pgjones (approved):

Issue #4078, opened 2021-05-20
`Config.from_json` should currently be deprecated, not removed


slokhorst:
After upgrading to Flask 2.0, using `app.config.from_json()` gives the following error:

```
AttributeError: 'Config' object has no attribute 'from_json'
```

[The changelog](https://flask.palletsprojects.com/en/2.0.x/changes/#version-2-0-0) only says that it has been deprecated, but not removed:

> Add `Config.from_file()` to load config using arbitrary file loaders, such as `toml.load` or `json.load`. `Config.from_json()` is deprecated in favor of this. #3398

PR #4080, opened 2021-05-20
updated text of changes due to #4078

Yourun-proger:


davidism:
Thanks, but it is in fact deprecated, I've already got a branch for 2.0.1 that will (temporarily) add it back.

PR #4081, opened 2021-05-20
re-add deprecated `Config.from_json` method

### LINKED ISSUE #4078: `Config.from_json` should currently be deprecated, not removed

davidism:
This was removed accidentally, it's deprecated until Flask 2.1.

- fixes #4078 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,8 @@ Unreleased
 -   Roll back a change to the order that URL matching was done. The
     URL is again matched after the session is loaded, so the session is
     available in custom URL converters. :issue:`4053`
+-   Re-add deprecated ``Config.from_json``, which was accidentally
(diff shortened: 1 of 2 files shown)


ThiefMaster:
somewhat unrelated, but shouldn't all those function also document their `:return:` value?

davidism:
@ThiefMaster sure, want to create a separate issue for that?

PR #4085, opened 2021-05-21
Improve commands in the contributing docs

greyli:
- Split `pip install -r requirements/dev.txt && pip install -e .` into separate commands to make it work on Windows.
- Add commands group for the command to build docs.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -121,9 +121,26 @@ First time setup
 -   Install the development dependencies, then install Flask in editable
     mode.
 
-    .. code-block:: text
+    .. tabs::
 
-        $ pip install -r requirements/dev.txt && pip install -e .
+        .. group-tab:: Bash
+
+            .. code-block:: text
+
+                $ pip install -r requirements/dev.txt && pip install -e .
+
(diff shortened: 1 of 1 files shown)


davidism:
`&&` should work on Windows. I didn't split it up because I wanted contributors to treat it as one install command, not multiple steps that they could miss.

greyli:
I think it's not working on PowerShell...Not sure if it's due to the version, will check it tomorrow. Maybe we could add a commands group here?

greyli:
The pipe operators only supported in [PowerShell 7.0+](https://github.com/PowerShell/PowerShell/releases/tag/v7.0.0-preview.5), and it's not supported in Windows PowerShell. So I think it's better to add a command group for it.

davidism:
Just confirmed the following on Windows CMD:

* `pip install -r requirements/dev.txt && pip install -e .` I don't want to add all this extra markup/explanation just for powershell, let's just leave that out unless someone runs into an issue.
* `make html` works, you don't need to do `.\make.bat`.

greyli:
`make html` not working on PowerShell... Ok, I will close this PR, then update related PRs to remove the explanation for `make html`.

PR #4087, opened 2021-05-21
Add venv and .venv to .gitignore

greyli:
Although we use `python -m venv env` in the contributing guide as the example, we don't need to force the contributor to use `env`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.pyc
 *.pyo
 env/
+venv/
+.venv/
 env*
 dist/
 build/


ThiefMaster:
People can use `.git/info/exclude` to add ignores locally, but I'd like to have `.venv` on gitignore as well since it's a very common name for virtualenvs, and it saves you one small step after setting up your environment...

greyli:
I just learned the local gitignore thing, thanks. I normally use `venv` as the name of the virtual environment.

PR #4094, opened 2021-05-21
improve typing for `teardown_request` 

### LINKED ISSUE #4093: Wrong typing for teardown_request decorator
According to the [doc](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.teardown_request), the return values of teardown functions are ignored. However in `flask 2.0.1` the decorator only accept functions returning a `Response`.

It should accept to decorate functions returning any type.

Environment:

- Python version: 3.9
- Flask version: 2.0.1

pcorpet:
- fixes #4093

Checklist:


pgjones:
I think returning None might be more informative to end users. This way they'll be warned if they add a return expecting it to be used.

pcorpet:
@pgjones yes, I prefer returning `None` as well. I didn't want to break the type of other code (especially for the ones that typed this file by setting `Response`), but I'll follow you gladly.

hynek:
Maybe the changelog should be expanded? I've run into the same problem with `teardown_appcontext` and it seems like this PR fixes it for all `teardown_*` methods?

pgjones:
Yep, minor change in 63893a427bd022d192febadad0b31ae06fa80776.

PR #4097, opened 2021-05-22
improve typing for `app.errorhandler` decorator

### LINKED ISSUE #4095: Improve typing for `errorhandler` decorator.
Today the `errorhandler` decorator does not have consistency between the type of error in its parameter and the parameter of the function it is wrapping although those should be the same.

The following code should be properly typed:
```py
@app.errorhandler(auth.ExpiredTokenException)
def expired_token(error: auth.ExpiredTokenException) -> Tuple[str, int]:
    """Handle the 498 error."""

    return error.description or i18n.flask_translate("Expired token"), 498
```

the current version enforces it to be typed as
```py
@app.errorhandler(auth.ExpiredTokenException)
def expired_token(untyped_error: Exception) -> Tuple[str, int]:
    """Handle the 498 error."""

    error = typing.cast(auth.ExpiredTokenException, untyped_error)
    return error.description or i18n.flask_translate("Expired token"), 498
```

pcorpet:
- fixes #4095

Checklist:

### DIFF
--- a/src/flask/typing.py
+++ b/src/flask/typing.py
@@ -33,11 +33,12 @@ ResponseReturnValue = t.Union[
     "WSGIApplication",
 ]
 
+GenericException = t.TypeVar("GenericException", bound=Exception, contravariant=True)
+
 AppOrBlueprintKey = t.Optional[str]  # The App key is None, whereas blueprints are named
 AfterRequestCallable = t.Callable[["Response"], "Response"]
 BeforeFirstRequestCallable = t.Callable[[], None]
 BeforeRequestCallable = t.Callable[[], t.Optional[ResponseReturnValue]]
-ErrorHandlerCallable = t.Callable[[Exception], ResponseReturnValue]
 TeardownCallable = t.Callable[[t.Optional[BaseException]], None]
 TemplateContextProcessorCallable = t.Callable[[], t.Dict[str, t.Any]]
 TemplateFilterCallable = t.Callable[..., t.Any]
@@ -45,3 +46,11 @@ TemplateGlobalCallable = t.Callable[..., t.Any]
 TemplateTestCallable = t.Callable[..., bool]
 URLDefaultCallable = t.Callable[[str, dict], None]
 URLValuePreprocessorCallable = t.Callable[[t.Optional[str], t.Optional[dict]], None]
+
+
+if t.TYPE_CHECKING:
+    import typing_extensions as te
+
+    class ErrorHandlerCallable(te.Protocol[GenericException]):
+        def __call__(self, error: GenericException) -> ResponseReturnValue:
+            ...
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Unreleased
     decorators. :issue:`4104`
 -   Fixed the issue where typing requires template global
     decorators to accept functions with no arguments. :issue:`4098`
+-   Enhance typing of ``app.errorhandler`` decorator. :issue:`4095`
 
 
 Version 2.0.1
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -61,7 +61,6 @@ from .templating import Environment
 from .typing import AfterRequestCallable
 from .typing import BeforeFirstRequestCallable
 from .typing import BeforeRequestCallable
-from .typing import ErrorHandlerCallable
 from .typing import ResponseReturnValue
 from .typing import TeardownCallable
 from .typing import TemplateContextProcessorCallable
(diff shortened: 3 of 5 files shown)


src/flask/typing.py
```diff
@@ -33,14 +33,20 @@
...
 TemplateContextProcessorCallable = t.Callable[[], t.Dict[str, t.Any]]
 TemplateFilterCallable = t.Callable[[t.Any], str]
 TemplateGlobalCallable = t.Callable[[], t.Any]
 TemplateTestCallable = t.Callable[[t.Any], bool]
 URLDefaultCallable = t.Callable[[str, dict], None]
 URLValuePreprocessorCallable = t.Callable[[t.Optional[str], t.Optional[dict]], None]
+
+
+class ErrorHandlerCallable(t.Protocol[GenericException]):
+    def __call__(self, error: GenericException, /) -> ResponseReturnValue:
```
pcorpet, line 51: Positional-only arguments (with the use of `/`) has been introduced in Python 3.8 so this code is not compatible with earlier interpreters. However it gives the benefit of not enforcing the name of the argument `error`.

If that's preferable I can also introduce some code that depends on the current version of `Python` and give another line before `3.8`

Yourun-proger: Apparently readthedocs doesn't accept this ...

ThiefMaster: I don't see how RTD is related here.

Anyway, I think introducing syntax that doesn't work on all supported Python versions is not a good idea (I guess type checking would fail on those versions?)

davidism: We need to wait until we drop Python 3.7 in 2023. Typing supports the same versions the code supports.

pcorpet: Yes, Read The Docs is parsing the python using python 3.7 thus choking on the `/`. Let me try to use a python version check before using the syntax and fallback on imposing the `error` name for the argument for earlier versions.

ThiefMaster: fwiw we could use 3.9 for the RTD build. but if we need a version check anyway there's probably no need for that

pcorpet: It look like Python still wants to be able to parse the protected code anyway, and I don't blame it: it could hide some open parenthesis. I'll drop my idea of letting the users choose their argument name: they'll have to stick to `error`

pcorpet:
Grrmmf, it seems that Python 3.7 doesn't have `typing.Protocol` which actually makes it quite hard to type properly.

I think I'm just going to leave it like this and drop this PR unless someone has a better idea.

src/flask/typing.py
```diff
@@ -33,15 +33,21 @@
     "WSGIApplication",
 ]
 
+GenericException = t.TypeVar("GenericException", bound=Exception, contravariant=True)
```
pgjones, line 36: @davidism I vaguely remember we had performance concerns about Generics, which I think doesn't apply here. Could you confirm this is fine?

davidism: This isn't a `Generic`, so it's fine in that regard. I have no idea what this type does though.

davidism:
@pcorpet `typing_extensions` is available, so you can do the following:

```python
if t.TYPE_CHECKING:
    import typing_extensions as te

    class MyProtocol(te.Protocol):
        ...
```

We use this in a few other places in Pallets.

pgjones:
Could you rebase this against the 2.0.x branch?

pgjones:
The web based conflict fixer sadly didn't work, so I've merged this manually. Thanks.

davidism:
Could I get an explanation for what `GenericException` is?

pcorpet:
@davidism, `GenericExcpetion` is a [TypeVar](https://docs.python.org/3/library/typing.html#typing.TypeVar), it is only useful for deining generic on top of it. In TypeScript, one would express it like this:
```ts
type ErrorHandlerCallable<T extends Exception> = ...
```
You can see that `T` is only defined for the definition of the generic type. In Python they have taken another route where you first define a generic type var that you bound properly (the equivalent of `extends Exception` above) and that you can then use inside generic expressions.

pcorpet:
In `scaffold.py` we want to check that the **same** exception type is used as a parameter of the decorator and the type of the parameter of the function:

For instance, the following code is not valid and MyPy will know about that thanks to the use of `GenericException`.

```py
@app.error_handler(ValueError)
def handle_value_error(error: AttributeError) -> str:
  return str(error)
```

davidism:
This fixed matching the decorator argument to the function argument, but broke being able to use the decorator multiple times on top of the same function. #4297

davidism:
In #4487, I relaxed the type for the callable argument to `Any`.

PR #4105, opened 2021-05-25
Fix type annotation for `before_request` and `before_app_request`

### LINKED ISSUE #4104: Wrong typing for `before_request` decorator
The [doc](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.before_request) says the following:
```
The function will be called without any arguments. 
If it returns a non-None value, the value is handled 
as if it was the return value from the view, and further 
request handling is stopped.
```
However the function type hint is not correct:
https://github.com/pallets/flask/blob/7161776824734fc2797fe2b4fc974d183487ebf8/src/flask/typing.py#L38

I guess it's because this type was used not only for `before_request` but for `before_first_request`, `before_app_request` and `before_app_first_request` decorators too. But that's what `before_first_request` [doc](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.before_first_request) says:
```
The function will be called without any arguments and its return value is ignored.
```

Maybe existing `BeforeRequestCallable` type should be renamed to `BeforeFirstRequestCallable` and used for only `before_first_request` and `before_app_first_request` decorators and there should be separate type `BeforeRequestCallable` for `before_request` and `before_app_request` decorators?

Environment:

- Python version: 3.9.5
- Flask version: 2.0.1

decaz:
- fixes #4104 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@ Version 2.0.2
 
 Unreleased
 
--  Fix type annotation for ``teardown_request``. :issue:`4093`
+-   Fix type annotation for ``teardown_request``. :issue:`4093`
+-   Fix type annotation for ``before_request`` and ``before_app_request``
+    decorators. :issue:`4104`
 
 
 Version 2.0.1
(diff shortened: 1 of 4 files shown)


pgjones:
Thank you.

PR #4108, opened 2021-05-26
Fix type annotation for `__exit__` methods

decaz:
From [docs](https://docs.python.org/3/reference/datamodel.html#object.__exit__):
```
If the context was exited without an exception, all three arguments will be None.
```

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Unreleased
 -   Fix type annotation for ``teardown_request``. :issue:`4093`
 -   Fix type annotation for ``before_request`` and ``before_app_request``
     decorators. :issue:`4104`
+-   Fix type annotation for ``__exit__`` methods. :pr:`4108`
 
 
 Version 2.0.1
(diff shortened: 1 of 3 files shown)


davidism:
Typing these isn't useful. I've just been omitting types on `__exit__` with a `# type: ignore` comment.

decaz:
But typing of `__exit__` methods exists and it conflicts with my inherited `FlaskClient.__exit__` method:
```
tests/helpers.py:65: error: Argument 1 to "__exit__" of "FlaskClient" has incompatible type "Optional[Type[BaseException]]"; expected "type"  [arg-type]
tests/helpers.py:65: error: Argument 2 to "__exit__" of "FlaskClient" has incompatible type "Optional[BaseException]"; expected "BaseException"  [arg-type]
tests/helpers.py:65: error: Argument 3 to "__exit__" of "FlaskClient" has incompatible type "Optional[TracebackType]"; expected "TracebackType"  [arg-type]
```

davidism:
Yes, I'm saying we should remove the types, since they're verbose and don't have much use in this case.

decaz:
Ok, will close it in favor of future PR with removal of this typing.

PR #4111, opened 2021-05-27
Clarification added to the tutorial

Yourun-proger:
This is due to the new version of Flask, where when serving static files, browsers will cache based on content rather than based on a 12 hour timer. This means changes to static content such as CSS styles will be reflected immediately on reload without having to clear the cache.


ThiefMaster:
I'm not sure whether this is actually useful... Most developers know that they can just CTRL+F5 to reload without cache.

davidism:
The original statement "If the change doesn’t show up, try clearing your browser’s cache." is still accurate, the fact that the user won't have to in most cases doesn't change the recommendation in case they *do* have to for some reason.

PR #4113, opened 2021-05-27
Support View and MethodView instances with async handlers

### LINKED ISSUE #4112: View and MethodView classes do not support async handlers
Regular views can now be written as async functions, but instances of the `View` and `MethodView` classes did not get updated to support this functionality. Consider the following example:

```python
import asyncio
from flask import Flask

app = Flask(__name__)

class Test(View):
    async def dispatch_request(self):
        await asyncio.sleep(1)
        return 'test'

class Test2(MethodView):
    async def get(self):
        await asyncio.sleep(1)
        return 'test2'

app.add_url_rule('/test', view_func=Test.as_view('test'))
app.add_url_rule('/test2', view_func=Test2.as_view('test2'))
```

The expectation is that you should be able to send requests to `/test` and `/test2` and that their respective handlers should be wrapped with the `app.ensure_sync()` method, as function-based handlers are.

Instead sending requests to these URLs cause the following error:

```
TypeError: The view function did not return a valid response. The return type must be a string, dict, tuple, Response instance, or WSGI callable, but it was a coroutine.
```

Environment:

- Python version: >= 3.7
- Flask version: >= 2

miguelgrinberg:
This change runs handlers in `View` and `MethodView` instances through the `app.ensure_sync()` wrapper, so that async handlers can be used.

- fixes #4112 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Unreleased
     decorators. :issue:`4104`
 -   Fixed the issue where typing requires template global
     decorators to accept functions with no arguments. :issue:`4098`
+-   Support View and MethodView instances with async handlers. :issue:`4112`
 
 
 Version 2.0.1
--- a/docs/async-await.rst
+++ b/docs/async-await.rst
@@ -18,6 +18,12 @@ defined with ``async def`` and use ``await``.
         data = await async_db_query(...)
         return jsonify(data)
 
+Pluggable class-based views also support handlers that are implemented as
+coroutines. This applies to the :meth:`~flask.views.View.dispatch_request`
+method in views that inherit from the :class:`flask.views.View` class, as
+well as all the HTTP method handlers in views that inherit from the
+:class:`flask.views.MethodView` class.
+
 .. admonition:: Using ``async`` on Windows on Python 3.8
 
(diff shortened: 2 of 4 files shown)


miguelgrinberg:
Let me know if this change looks reasonable and I'll update the docs and change log. Thanks!

pgjones (approved):
I take the view that this is a bug fix, given the intention was to support this in 2.0.

pgjones:
There is something a little weird with the commits in this PR - why is 613b46b35c38cd1a2a6941e60c0ee5ab0132d502 present? I'm otherwise happy to merge.

davidism:
I've seen that happen if you don't use `--onto` when changing base.

```
git rebase --onto origin/2.0.x origin/main
git push -f
```

miguelgrinberg:
Sorry about the merge, I did not even realize I somehow dragged that extra commit. It should be fixed now.

hmilkovi:
Hi just a quick question, is this deployed to pypi package?

miguelgrinberg:
@hmilkovi Not yet. Keep an eye out for the 2.0.2 release, that's the one that has this change.

PR #4115, opened 2021-05-28
Fix: avoid duplicate loop key in err handler search

laggardkernel:
Very minimal change. The code is self-explanation. `code` in `for c in [code, None]` may be `None`. Made a deduplication before do the loop.

Checklist:

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1276,7 +1276,7 @@ class Flask(Scaffold):
         """
         exc_class, code = self._get_exc_class_and_code(type(e))
 
-        for c in [code, None]:
+        for c in [code, None] if code is not None else [None]:
             for name in chain(request.blueprints, [None]):
                 handler_map = self.error_handler_spec[name][c]
 


src/flask/app.py
```diff
@@ -1276,7 +1276,7 @@ def _find_error_handler(self, e: Exception) -> t.Optional[ErrorHandlerCallable]:
         """
         exc_class, code = self._get_exc_class_and_code(type(e))
 
-        for c in [code, None]:
+        for c in [None] if code is None else [code, None]:
```
pgjones, line 1279: What about `for c in {code, None}:`? Reads more clearly clearer to me

davidism: Order isn't guaranteed for sets.

pgjones: Oh yea, the order matters.

laggardkernel:
Adjusted the code a little to avoid typing error, and rebased onto the HEAD of branch '2.0.x'.

PR #4127, opened 2021-06-03
Fix async view handling

### LINKED ISSUE #4126: Async class based view causes crash
## The Bug
* `flask.views.View` does not properly handle conversion of an async `dispatch_request`, resulting in a coroutine that is never awaited.
* `flask.views.MethodView` does not properly handle conversion of async methods, resulting in coroutines that are never awaited.

## Replication

1. Create an async `dispatch_request` method on a `View`, or a `get` method on a `MethodView`.
1. Add the view to the app using the `View.as_view()` and `app.add_url_rule()` functions.
1. Access the route and observe a crash and 500 status code.

The following application reproduces the bug.

```python
from flask import Flask

app = Flask("MyApp")

class TestAsyncView(View):
    async def dispatch_request(self):
        return 'ASYNC VIEW RESPONSE'

class TestAsyncMethodView(MethodView):
    async def get(self):
        return 'ASYNC METHODVIEW GET RESPONSE'

app.add_url_rule("/async_view", view_func=TestAsyncView.as_view("test_async_view"))
app.add_url_rule('/async_methodview', view_func=TestAsyncMethodView.as_view('test_async_methodview'))

if __name__ == "__main__":
    app.run(port=8000)
```

The following error occurs after running `curl localhost:8000/async_view`:

```
  File "/Users/tpansino/NewRelic/Testing/Flask2Bug/flask/src/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/tpansino/NewRelic/Testing/Flask2Bug/flask/src/flask/app.py", line 1519, in full_dispatch_request
    return self.finalize_request(rv)
  File "/Users/tpansino/NewRelic/Testing/Flask2Bug/flask/src/flask/app.py", line 1538, in finalize_request
    response = self.make_response(rv)
  File "/Users/tpansino/NewRelic/Testing/Flask2Bug/flask/src/flask/app.py", line 1730, in make_response
    raise TypeError(
TypeError: The view function did not return a valid response. The return type must be a string, dict, tuple, Response instance, or WSGI callable, but it was a coroutine.
```

```
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
```

## Expected Behavior
* Asynchronous `dispatch_request` functions and methods should either be:
  * Converted using `async_to_sync` before being returned from `as_view()`.
  * Executed or awaited in the application when a route is accessed.

## Environment

- Python version: 3.9.5
- Flask version: 2.0.1

TimPansino:
## Changes

* Move `async_to_sync` and `ensure_sync` off of Flask class and into helpers.
  * This is open to discussion, but the necessity of referencing these functions from outside the context of the application may outweigh the ability to override them using the current implementation. More consideration is probably needed around how to make this behavior overridable while still being accessible from Views.
* Call `ensure_sync` on `dispatch_request` for `Views` and methods for `MethodViews`

## Issue

- fixes #4126

Checklist:

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1,5 +1,3 @@
-import functools
-import inspect
 import logging
 import os
 import sys
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -16,7 +16,6 @@ from werkzeug.exceptions import BadRequest
 from werkzeug.exceptions import BadRequestKeyError
 from werkzeug.exceptions import HTTPException
 from werkzeug.exceptions import InternalServerError
-from werkzeug.local import ContextVar
 from werkzeug.routing import BuildError
 from werkzeug.routing import Map
 from werkzeug.routing import MapAdapter
@@ -37,6 +36,7 @@ from .globals import g
 from .globals import request
 from .globals import session
 from .helpers import _split_blueprint_path
+from .helpers import ensure_sync
 from .helpers import get_debug_flag
 from .helpers import get_env
 from .helpers import get_flashed_messages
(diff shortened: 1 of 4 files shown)


TimPansino:
Tests are still needed, opening as draft until they're completed.

TimPansino:
Will update docs and comments before this merges if this proposed fix is accepted.

miguelgrinberg:
In my view, one of the reasons why `ensure_sync` should stay as a method of the `Flask` class is that this allows applications or Flask extensions to override it and provide a different implementation. I am currently taking advantage of this in my aioflask package. If `ensure_sync` is moved to a static function then it would be impossible to override.

Also, a patch that was merged a few days ago adds support for `View` and `MethodView`. This is coming with the next Flask 2.0.x release. See https://github.com/pallets/flask/issues/4112.

PR #4134, opened 2021-06-05
Add specified HTTP method validation.

iml1111:
I Added a function to check if a method in the input string list is a valid method supported by HTTP when specifying a method using "route".

Sometimes I entered this value incorrectly and suffered incorrect behavior.

Checklist:

[x] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
[x] Add or update relevant docs, in the docs folder and in code.
[x] Add an entry in CHANGES.rst summarizing the change and linking to the issue.
[x]  Add .. versionchanged:: entries in any relevant code docs.
[x] Run pre-commit hooks and fix any issues.
[x] Run pytest and tox, no tests failed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
 .. currentmodule:: flask
 
+
+Version 2.1.1
+-------------
+
+Unreleased
+
+-   Add specified HTTP method validation.
+
+
 Version 2.1.0
 -------------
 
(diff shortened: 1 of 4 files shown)


davidism:
I think this has come up before, but can't find the discussion. Any value is technically ok, for example there are methods used by webdav, which you've added, but there may be others as well.

This would be good as a linting plugin, but I don't think it should be an error.

iml1111:
Thanks for reviewing my code!

PR #4139, opened 2021-06-07
Avoid race condition in example app

### LINKED ISSUE #4099: Harmless race condition in tutorial
I was browsing the flaskr tutorial when I noticed an (admittedly quite unlikely) race condition in the `register` view, specifically:

```py
if not username:
    error = 'Username is required.'
elif not password:
    error = 'Password is required.'
elif db.execute(
    'SELECT id FROM user WHERE username = ?', (username,)
).fetchone() is not None:
    error = f"User {username} is already registered."

if error is None:
    db.execute(
        'INSERT INTO user (username, password) VALUES (?, ?)',
        (username, generate_password_hash(password))
    )
    db.commit()
    return redirect(url_for('auth.login'))
```

If two requests arrive with the right timing, the following can happen:

```
   Request 1:                                Request 2:
SELECT id
  FROM user
 WHERE username = abc
     |
     v
empty, no such user

                                          SELECT id
                                            FROM user
                                           WHERE username = abc
                                               |
                                               v
                                          empty, no such user

INSERT INTO user (username, password)
     VALUES (abc, 123)
     |
     v
    ok

                                          INSERT INTO user (username, password)
                                               VALUES (abc, 456)
                                               |
                                               v
                                          failed UNIQUE constraint -> 
                                          -> sqlite3.IntegrityError ->
                                          -> user gets HTTP 500
```

While the likelihood of this happening is pretty small and the harm practically zero (user gets HTTP 500 and has to manually login/choose a different username), I feel like this is not really the sort of good practice the tutorial should teach. I also believe it's important the developer understands that it's the UNIQUE constraint that ensures their app works correctly and not the if condition in the application code (the tutorial mentions SQL injection attacks and explains what protects the developer against them, so I don't really feel this is out of scope).

In my own app I've modified the code to the following:
```py
if not username:
    error = 'Username is required.'
elif not password:
    error = 'Password is required.'
else:
    try:
        db.execute(
            'INSERT INTO users (username, password) VALUES (?, ?)',
            (username, generate_password_hash(password))
        )
        db.commit()
    except IntegrityError:
        error = f"User {username} is already registered."
    else:
        return redirect(url_for('auth.login'))
```

I suggest something similar be incorporated into the tutorial, with a short explanation (maybe a comment) of how the UNIQUE constraint does the work for the developer and maybe a note about the principle that one should "ask forgiveness, not permission." I'm not sure on how it's better worded, so I'm making this an issue instead of a pull request.

Cheers, and thank you for your great work!

awijaya22:
When the race condition happen, it will return the error message(explanation) instead of redirecting user to other page without explaining
- fixes #4099
Checklist:

### DIFF
--- a/examples/tutorial/flaskr/auth.py
+++ b/examples/tutorial/flaskr/auth.py
@@ -60,21 +60,21 @@ def register():
             error = "Username is required."
         elif not password:
             error = "Password is required."
-        elif (
-            db.execute("SELECT id FROM user WHERE username = ?", (username,)).fetchone()
-            is not None
-        ):
-            error = f"User {username} is already registered."
 
         if error is None:
-            # the name is available, store it in the database and go to
-            # the login page
-            db.execute(
-                "INSERT INTO user (username, password) VALUES (?, ?)",
-                (username, generate_password_hash(password)),
-            )
-            db.commit()
-            return redirect(url_for("auth.login"))
+            try:
+                db.execute(
+                    "INSERT INTO user (username, password) VALUES (?, ?)",
+                    (username, generate_password_hash(password)),
+                )
+                db.commit()
+            except db.IntegrityError:
+                # The username was already taken, which caused the
+                # commit to fail. Show a validation error.
+                error = f"User {username} is already registered."
+            else:
+                # Success, go to the login page.
+                return redirect(url_for("auth.login"))
 
         flash(error)
 
--- a/docs/tutorial/views.rst
+++ b/docs/tutorial/views.rst
@@ -91,18 +91,18 @@ write templates to generate the HTML form.
                 error = 'Username is required.'
             elif not password:
                 error = 'Password is required.'
-            elif db.execute(
-                'SELECT id FROM user WHERE username = ?', (username,)
-            ).fetchone() is not None:
-                error = f"User {username} is already registered."
 
             if error is None:
-                db.execute(
-                    'INSERT INTO user (username, password) VALUES (?, ?)',
-                    (username, generate_password_hash(password))
-                )
-                db.commit()
-                return redirect(url_for('auth.login'))
+                try:
+                    db.execute(
+                        "INSERT INTO user (username, password) VALUES (?, ?)",
+                        (username, generate_password_hash(password)),
+                    )
+                    db.commit()
+                except db.IntegrityError:
+                    error = f"User {username} is already registered."
+                else:
+                    return redirect(url_for("auth.login"))
 
             flash(error)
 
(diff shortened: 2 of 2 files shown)


CHANGES.rst
```diff
@@ -18,6 +18,7 @@ Unreleased
     decorators. :issue:`4104`
 -   Fixed the issue where typing requires template global
     decorators to accept functions with no arguments. :issue:`4098`
+-   Update docs of register view, to return explanation if insert to db failed. :issue:`4099`
```
ThiefMaster, line 21: I don't think updates to the example need a change log entry. It's not useful information for people who want to see what changed in an update.

ThiefMaster:
You need to do a rebase to change the upstream branch of your changes; merge commits don't work for this, and just create a mess from which it may be tricky to recover

CHANGES.rst
awijaya22: It is removed now. Thanks

examples/tutorial/flaskr/auth.py
```diff
@@ -49,6 +49,8 @@ def register():
 
     Validates that the username is not already taken. Hashes the
     password for security.
+     .. versionchanged:: 2.0.2
+        Return explanation (error message) to user when insert data to db failed.
```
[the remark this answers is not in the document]
awijaya22: It is removed. Thanks

awijaya22:
Sorry I am new here, do you mean to rebase my branch to the current 2.0.x branches, and recommit my changes?

ThiefMaster:
`git rebase --onto origin/2.0.x main` and `git push -f` is all you need - at least with a clean branch. i don't know how well this works now with all those merge commits...

awijaya22:
Can't rebase cause by conflict at CHANGE.rst. 
I am sorry for causing this mess.

ThiefMaster:
```
git reset --hard origin/2.0.x
git cherry-pick 6ef483d66
git push --force
```

try this. that way anything but your relevant commit will be kept.

i would also recommend doing a `git commit --amend` to set a proper commit message. issue references are meaningless without context so they should not be in the first line of a commit message.

You could go for something like "Avoid race condition in example app" instead and then, in the 3rd line (2nd is always empty) of the commit message put `closes #4099`

awijaya22:
Done the above command (including the amend commit)

davidism:
The tutorial needs to be updated to reflect these changes, the wording will need to be changed to describe what's going on now.

awijaya22:
which part of the tutorial are you referring to ?

davidism:
The part that describes the code you're changing.

awijaya22:
Do I need to do rebase, before changing the tutorial comment, or can I just commit directly?

davidism:
That comment wasn't what I was referring to, I meant the actual text of the tutorial. Be sure to use the same writing style.

davidism:
I rewrote this a bit more. I also updated the email you used for the commit to match the anonymous one associated with your GitHub account, you may want to check your git config to make sure that's the default.

PR #4152, opened 2021-06-16
`static_folder` can be a `pathlib.Path` object

### LINKED ISSUE #4150: static_folder shows str type, not pathlib.Path
Using Python 3.9, mypy will complain about static_folder being `Optional[str]` and not accepting a `pathlib.Path`.

```
note: Revealed type is "pathlib.Path*" # This is what I'm feeding it
Revealed type is "Union[builtins.str, None]" # this is reveal_path() for app.static_folder
error: Incompatible types in assignment (expression has type "Path", variable has type "Optional[str]")
Found 1 error in 1 file (checked 1 source file)
```

Not sure why Python 3.8 doesn't do this.

I think maybe here: https://github.com/pallets/flask/blob/main/src/flask/scaffold.py#L80

I think this needs to be `_static_folder: t.Union[t.Optional[str], Path] = None`

Or maybe... `_static_folder: t.Optional[t.Union[str, Path]] = None`

And at the top: `from pathlib import Path`

Thank you!

Yourun-proger:
- fixes #4150 

Checklist:
- [x] Add .. versionchanged:: entries in any relevant code docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@ Unreleased
 -   Support View and MethodView instances with async handlers. :issue:`4112`
 -   Enhance typing of ``app.errorhandler`` decorator. :issue:`4095`
 -   Fix registering a blueprint twice with differing names. :issue:`4124`
+-   Fix the type of ``static_folder`` to accept ``pathlib.Path``.
+    :issue:`4150`
 
 
 Version 2.0.1
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -389,7 +389,7 @@ class Flask(Scaffold):
         self,
         import_name: str,
         static_url_path: t.Optional[str] = None,
-        static_folder: t.Optional[str] = "static",
+        static_folder: t.Optional[t.Union[str, os.PathLike]] = "static",
         static_host: t.Optional[str] = None,
         host_matching: bool = False,
         subdomain_matching: bool = False,
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -1,3 +1,4 @@
+import os
 import typing as t
 from collections import defaultdict
 from functools import update_wrapper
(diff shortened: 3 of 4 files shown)


pgjones:
This doesn't fix the issue - the typing needs to be extended to the static_folder arguments and then support for Path needs to be added where the static_folder is used.

davidism:
Also need to use `os.PathLike` when typing.

Yourun-proger:
@davidism, I want to clarify. That is, instead of `pathlib.Path`, you need to use `os.PathLike`  both in the static_folder and in the _ static_folder **too**?

Yourun-proger:
Hello. I understood your adjustments and fix!

i3Cheese:
`value = os.fspath(value).rstrip(r"\/")` returns `str`, so `static_folder` always return str, but can be set to `os.PathLike`

Yourun-proger:
Hey! Do my changes still need any adjustments?

i3Cheese:
Think 81 should be original  `_static_folder: t.Optional[str] = None`

Yourun-proger:
Okay. I'll squash those 3 commits, then fix line 81. Thanks!

Yourun-proger:
Okay. Now I seem to have done everything)
Or do you need something else?

Yourun-proger:
Hey, @davidism , is everything right now?

greyli:
You don't need to ask for reviewing or merging, please be patient. By the way, using "I fix it!" as the commit summary is not a good idea, and the current commit description is also meaningless.

Yourun-proger:
Sorry! I am silent... I changed the commit message.

Yourun-proger:
I found [this](https://github.com/pallets/flask/blob/7a73171edc3bcffef96ef6367977ab3ae9af9350/src/flask/app.py#L392) line of code and I suspect that it needs to be fixed too. But I'm not sure

Yourun-proger:
Okay. All is good!

Yourun-proger:
In Blueprint class i see `static_folder` too and i'll fix it tomorrow

Yourun-proger:
Okay. I changed this. But i don’t really know where exactly to add `.. versionchanged :: 2.0.2` in Flask and Blueprint classes. Can anyone please help? Or say about some other necessary changes. Thank you!

PR #4161, opened 2021-06-19
Improve the code example in testing docs

yuxiaoy1:
The `request` should be `flask.request`.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -333,7 +333,7 @@ happen.  With Flask 0.4 this is possible by using the
 
     with app.test_client() as c:
         rv = c.get('/?tequila=42')
-        assert request.args['tequila'] == '42'
+        assert flask.request.args['tequila'] == '42'
 
 If you were to use just the :meth:`~flask.Flask.test_client` without
 the ``with`` block, the ``assert`` would fail with an error because `request`


jab:
I'm not sure this change is really necessary; it should be fairly obvious that this code assumes `from flask import request`, which does appear elsewhere in this same documentation. Can you say whether this actually caused you or someone you know confusion?

On the other hand, in other places in this same documentation, `flask.request` is used, as in this change. If anything, perhaps this PR should change all other `request` (and `jsonify` etc.) references in these docs to `flask.request` (and `flask.jsonify` etc.) to be consistent.

In any case, I believe we can't accept this until it targets the `2.0.x` branch rather than the `main` branch.

yuxiaoy1:
I think we should use `flask.xxx` for consistency as it's also used in the flask test.
I will cover all the changes needed and branch off to 2.0.x.

davidism:
I prefer not using `flask.`.

yuxiaoy1:
The code example in same chapter [Other Testing Tricks](https://flask.palletsprojects.com/en/2.0.x/testing/#other-testing-tricks) and [Accessing and Modifying Sessions](https://flask.palletsprojects.com/en/2.0.x/testing/#accessing-and-modifying-sessions) use `flask.xxx`, IMO, we should keep the consistency.

davidism:
I guess a single PR that changes all uses of `flask.{name}` to the plain version would be fine.

yuxiaoy1:
Sure, will submit a new PR soon.

PR #4165, opened 2021-06-20
Change flask.xxx to plain version in testing docs

yuxiaoy1:


davidism:
You pointed out that this was inconsistent in multiple pages. Can you make this consistent throughout the docs?

yuxiaoy1:
The chapters I said are all within the testing doc, sorry for the misunderstanding.
The inconsistency only appears in the testing doc, thus I think you could merge the PR now.

PR #4175, opened 2021-06-26
Fix typo in views docs

greyli:
The "function" in this section should be "method". Just caught this when reviewing the Chinese translation of this chapter (https://github.com/greyli/flask-docs-zh/pull/70).

### DIFF
--- a/docs/views.rst
+++ b/docs/views.rst
@@ -113,8 +113,8 @@ Method Based Dispatching
 
 For RESTful APIs it's especially helpful to execute a different function
 for each HTTP method.  With the :class:`flask.views.MethodView` you can
-easily do that.  Each HTTP method maps to a function with the same name
-(just in lowercase)::
+easily do that.  Each HTTP method maps to a method of the class with the
+same name (just in lowercase)::
 
     from flask.views import MethodView
 


docs/views.rst
```diff
@@ -111,9 +111,9 @@ information::
 Method Based Dispatching
 ------------------------
 
-For RESTful APIs it's especially helpful to execute a different function
+For RESTful APIs it's especially helpful to execute a different method
```
jab, line 114: The change below is good, but I would remove the change on this line. I think it's clearer as-is. (And methods are a type of function, so it's not incorrect as-is.)

jab (changes requested):

docs/views.rst
greyli: Sounds good, will revert it. I was also a little hesitant to change this line.

greyli: Updated.

docs/views.rst
```diff
@@ -113,7 +113,7 @@ Method Based Dispatching
 
 For RESTful APIs it's especially helpful to execute a different function
 for each HTTP method.  With the :class:`flask.views.MethodView` you can
-easily do that.  Each HTTP method maps to a function with the same name
+easily do that.  Each HTTP method maps to a class method with the same name
```
jab, line 116: Actually, would this be clearer as “method of a class”, to reduce potential confusion with `classmethod` (since these are regular methods, not classmethods)?

jab: or even “corresponding method of a class”

greyli: Yes, it would be better.

The "corresponding" seems redundant since we already have "...with the same name". I would prefer to use "maps to **a method of a class**", or "maps to **a method of the class**", "the class" here refers to the example class below.

jab (approved):

PR #4196, opened 2021-07-09
Remove assumption that only us generate warnings.

postmasters:
In `test_max_cookie_size`, it is assumed that only Flask warnings are generated. This assumption could be broken if any of Flask dependencies or Python built-in modules produce, say, a `DeprecationWarning`. This commit asserts only what we can control.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -2002,14 +2002,13 @@ def test_max_cookie_size(app, client, recwarn):
         return r
 
     client.get("/")
-    assert len(recwarn) == 1
-    w = recwarn.pop()
-    assert "cookie is too large" in str(w.message)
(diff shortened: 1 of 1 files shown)


davidism:
I don't understand this change. If something else is generating warnings, then the test should fail and we should address the warnings.

postmasters:
How are you going to address warnings from built-in libraries? The purpose of this test is to ensure that the warning of cookie oversize is generated. It shouldn't assert anything else. This PR does not ignore warnings. It still asserts that such warning is generated.

davidism:
If a built-in library is warning about a behavior, then we need to not do that behavior. Presumably it's either deprecated or has a different way to go about it. Are you trying to run the tests and this one is failing because of another warning? If so, report that.

PR #4197, opened 2021-07-10
Add github disc

Yourun-proger:
Okay, i make change and it isn't in **docs**.
discussions on github are cool and i added these!

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -9,3 +9,6 @@ contact_links:
   - name: Questions and discussions
     url: https://discord.gg/pallets
     about: Discuss questions about your code on our Discord chat.
+  - name: Github Discussions
+    url: https://github.com/pallets/flask/discussions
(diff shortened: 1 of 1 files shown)


jab:
There is a typo in this which prevents us from accepting it in its current state, but I’m leaning against accepting it even if the typo is fixed. This isn’t that impactful an improvement, and you’ve been asked a couple times to try contributing a fix for an existing, known issue, rather than continuing to try to change something else, and you seem to be disregarding that request. Could you please focus on getting your first fix for an existing issue accepted, before submitting anymore of these?

Yourun-proger:
Thank you @jab! 
It took me a long time to formulate an answer for you.
You are right, I, apparently, was too inspired and convinced of the need for my changes, it was too stupid and naive to impose and pester you. Forgive all those whom I needlessly ruined their nerves and disturbed. Now I will act smarter in open-source projects and I will not be annoying so that people do not think how to get rid of me. Sorry and forgive me again.

PR #4200, opened 2021-07-14
Add coverage intro

DayuanTan:
only add a few sentences to make it more friendly to those who don't know "coverage" before.


davidism:

DayuanTan:
Well, it would be great if these few lines can be added. It stopped me and took me few minutes to figure out why "coverage run -m pytest" doesn't work for me. Adding them makes beginners like me less confused. I appreciate it.   @davidism

PR #4202, opened 2021-07-16
Fix tutorial `test_auth.test_logout`

### LINKED ISSUE #4201: Flask tutorial `test_auth.test_logout` doesn't fully log in before it logs out
In the Flask tutorial, `test_auth.test_logout` passes even if one comments out the call to `auth.logout()` on line 68:

https://github.com/pallets/flask/blob/7a73171edc3bcffef96ef6367977ab3ae9af9350/examples/tutorial/tests/test_auth.py#L64-L69

It should probably match the structure of `test_auth.test_login`, and call `client.get("/")` to populate the session:

https://github.com/pallets/flask/blob/7a73171edc3bcffef96ef6367977ab3ae9af9350/examples/tutorial/tests/test_auth.py#L49-L51

So that the test looks like this:
```
def test_logout(client, auth):
    auth.login()

    with client:
        client.get("/") # Or similar so that `"user_id" in session` is true here
        auth.logout()
        assert "user_id" not in session
```

mkillianey:
Fixes #4201

Checklist:

- [N/A] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [N/A] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [N/A] Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/docs/tutorial/tests.rst
+++ b/docs/tutorial/tests.rst
@@ -349,6 +349,7 @@ not contain ``user_id`` after logging out.
         auth.login()
 
         with client:
+            client.get('/')
             auth.logout()
             assert 'user_id' not in session
 
--- a/examples/tutorial/tests/test_auth.py
+++ b/examples/tutorial/tests/test_auth.py
@@ -62,5 +62,6 @@ def test_logout(client, auth):
     auth.login()
 
     with client:
+        client.get("/")
         auth.logout()
         assert 'user_id' not in session


davidism:
Please rebase to 2.0.x.

mkillianey:
> Please rebase to 2.0.x.

Done.  Should I also change the target branch to 2.0.x?

FWIW, it's a clean merge back to the 1.0.x and 1.1.x branches, if you want to port them there...see commit d14c54b5d95fb7d246ba3e7d0484e0ad7f015d12 in this PR's history.  (I actually discovered this issue doing the 1.1.x tutorial, since I just inherited a pre-existing codebase using 1.1.x.)

davidism:
Thanks, however after looking at this again it seems like the change isn't needed.

PR #4223, opened 2021-08-05
Add support for a `FLASK_RUN_EXTRA_FILES` config variable.

### LINKED ISSUE #4222: Add support for a `FLASK_RUN_EXTRA_FILES` config variable.
[This section of the flask documentation](https://flask.palletsprojects.com/en/2.0.x/cli/?highlight=flask_run_extra_files#watch-extra-files-with-the-reloader) discusses how you can watch additional files with the reloader. 

You can currently do this by using the `--extra_files` option of the `flask run` command. Or by setting a `FLASK_RUN_EXTRA_FILES` environment variable. 

Ideally you could bundle this in with your application config, since it's not something that's likely to change from environment to environment but is inherent to the application itself.

Currently, I work around this by hard-coding the setting of the environment variable in my development config class:

```python
class DevelopmentConfig(Config):
    DEBUG = True
    ENV = "development"
    extra_files = [os.path.join("foo","bar.txt"), os.path.join("assets","styles.scss")]
    if os.name != 'nt':
            # Posix
            os.environ["FLASK_RUN_EXTRA_FILES"] = ":".join(extra_files)
        else:
            # Windows
            os.environ["FLASK_RUN_EXTRA_FILES"] = ";".join(extra_files)
```

I'm proposing a config variable that supports a list of file paths to watch:
```python
class DevelopmentConfig(Config):
    DEBUG = True
    ENV = "development"
    FLASK_RUN_EXTRA_FILES = [os.path.join("foo","bar.txt"), os.path.join("assets","styles.scss")]
```

graydenshand:
This commit adds support for a `FLASK_RUN_EXTRA_FILES=[]` config variable to specify other files to watch. 

It amends the two methods of running the app (`flask run` & `app.run()`) to pass this config variable to the `extra_files` argument of the `run_server()` function. 

Open to other implementation suggestions.

- fixes #4222

## Testing
I'm not sure the right way to add this to the testing suite, and am grateful for any direction. That said, here's a minimal reproducible example to test. 

1) Create two files: `main.py` and `test.txt`
```python
# main.py
from flask import Flask

app = Flask(__name__)

#app.config['FLASK_RUN_EXTRA_FILES'] = ["test.txt"]

if __name__ == "__main__":
	app.run(debug=1)
```

```
# test.txt
test
```

2) Run the server with the config variable declaration commented out. Verify that the server does not reload when you change `test.txt`.

3) Stop the server.

4) Uncomment the config variable declaration in `main.py`.

5) Run the server and verify that the server now does reload when you change `test.txt`. 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Version 2.1.0
 Unreleased
 
 -   Update Click dependency to >= 8.0.
+-   Add support for a `FLASK_RUN_EXTRA_FILES` config variable.
 
 
 Version 2.0.2
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -232,6 +232,7 @@ separated with ``:``, or ``;`` on Windows.
 
 Alternatively, you can set the ``FLASK_RUN_EXTRA_FILES`` config
 variable, which takes a list of files::
+
   from flask import Flask
 
   app = Flask(__name__)
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -184,11 +184,15 @@ reloader.
 Watch Extra Files with the Reloader
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. versionadded:: 2.1
+  Support for the ``FLASK_RUN_EXTRA_FILES`` config variable was added.
+
 When using development mode, the reloader will trigger whenever your
 Python code or imported modules change. The reloader can watch
 additional files with the ``--extra-files`` option, or the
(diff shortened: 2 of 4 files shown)


davidism:
Rhetorical question: Why just this environment variable, and not every other environment variable that could be passed the CLI?

Application config is about the application. The development server  is not the application, and so is not configured with the application config. Instead, it works as it does now intentionally: dev server config is passed to the `app.run` method, or the `flask run` CLI args or env vars.

graydenshand:
Ok thanks anyways

PR #4230, opened 2021-08-09
Fix callback order for nested blueprints

### LINKED ISSUE #4229: Handlers are called in wrong order when using nested blueprints
When using nested blueprints, handlers registered via `url_value_preprocessor`, `before_request`, `context_processor`, or `url_defaults` are called in an inconsistent order: First on the app, then on the deepest blueprint, and then upwards through all parents (except the app). Here is a minimal example:

```python
from flask import Flask, Blueprint, g

app = Flask(__name__)
parent = Blueprint('parent', __name__)
child = Blueprint('child', __name__)

@app.before_request
def app_before():
    g.test = 'app'

@parent.before_request
def parent_before():
    g.test = 'parent'

@child.before_request
def child_before():
    g.test = 'child'

@child.route('/')
def test():
    return g.test

app.register_blueprint(parent)
```
I would expect this to return `child`, but it returns `parent`.

Environment:

- Python version: 3.9.6
- Flask version: 2.0.1

matipau:
- fixes #4229 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,9 @@ Unreleased
     :issue:`4096`
 -   The CLI loader handles ``**kwargs`` in a ``create_app`` function.
     :issue:`4170`
+-   Fix the order of ``before_request`` and other callbacks that trigger
+    before the view returns. They are called from the app down to the
+    closest nested blueprint. :issue:`4229`
 
 
 Version 2.0.1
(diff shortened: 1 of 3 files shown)


davidism:
@pgjones can you check this? Seems to go against the discussion of `request.blueprints` order that we had.

pgjones:
I can't remember the details of the discussion that led to the current order. I do think before-functions going from app down and after-functions going from app up makes more sense though.

davidism:
I added a changelog, rebased this to 2.0.x, and did a bit more refactoring.

PR #4233, opened 2021-08-13
DOCS: changed flask-uploads to flask-reuploaded

Abdur-rahmaanJ:
Recommend updated fork


greyli:
In my opinion, we may better just remove the mention of Flask-Uploads (#4234). I'm not questioning the future maintenance of Flask-Reuploaded, but in general, we are trying to stop mention Flask extensions in the docs to prevent updates like this. In the meanwhile, we are planning to create and maintain a Flask extension listing that will list all the active Flask extensions.

Abdur-rahmaanJ:
Great!

PR #4251, opened 2021-08-30
Add suggestion for resolving error caused by endpoint overlapping

### LINKED ISSUE #3144: Flask throwing AssertionError: View function mapping is overwriting an existing endpoint function: register
(I am not very exprienced) 
Whenever I try to run the code, I get this error: 

  File "C:\Users\User\source\repos\FlaskWebProject1\FlaskWebProject1\FlaskWebProject1\views.py", line 48, in <module>
    @app.route('/register', methods=['GET', 'POST'])
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\site-packages\flask\app.py", line 1250, in decorator
    self.add_url_rule(rule, endpoint, f, **options)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\site-packages\flask\app.py", line 66, in wrapper_func
    return f(self, *args, **kwargs)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\lib\site-packages\flask\app.py", line 1221, in add_url_rule
    'existing endpoint function: %s' % endpoint)
AssertionError: View function mapping is overwriting an existing endpoint function: register
Press any key to continue . . .

This is the function register:

@app.route('/register', methods=['GET', 'POST'])
def register():
    form = RegistrationForm(request.form)
    if request.method == 'POST' and form.validate():
        user = User(form.username.data, form.email.data,
                    form.password.data)
        db_session.add(user)
        flash('Thanks for registering')
        return redirect(url_for('/home'))
    else:
        return render_template(
            'register.html', 
            year=datetime.now().year,
            month=datetime.now().month,
            day=datetime.now().day,
            hour=datetime.now().hour,
            minute=datetime.now().minute,
            form=form
         )
Can anyone help me?

### LINKED ISSUE #3750: Random error function mapping over existing endpoint function.
I have a few dozen endpoints which all have different routes (e.g. `/api/login/`, `/api/register` etc). From time to time, after I make a few changes to my `app.py`, I get an `Assertion Error`:
`AssertionError: View function mapping is overwriting an existing endpoint function: login`

I checked several times and concluded that there are 0 functions mapping over one another (no functions with same names or routes listening for the same route). 

### Environment

* Python version: 3.8
* Flask version: 1.1.2

### LINKED ISSUE #3930: Can not wrap blueprint on others wrappers with same function name
By example:

```python
flask_app = Flask(...)
app = Blueprint(...)

def mywrapp(func):
    def wrapped():
        return func()
    return wrapped

@app.route('/one')
@mywrapp
def route1():
    pass

@app.route('/two')
@mywrapp
def route2():
    pass

flask_app.register_blueprint(app)
```

Flask says an error: `AssertionError: View function mapping is overwriting an existing endpoint function: name.wrapped`. When invert the position between route and mywrapp can not call the wrapp function.

Hepheir:
Seems like a lot of people are having problems for the same reason.
IMHO, I think it will make flask more accessible if this suggestion or some comments that describes on what causes the error or how to fix it.

- fixes #3930
- fixes #3750
- fixes #3144

Checklist:

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1093,7 +1093,9 @@ class Flask(Scaffold):
             if old_func is not None and old_func != view_func:
                 raise AssertionError(
                     "View function mapping is overwriting an existing"
-                    f" endpoint function: {endpoint}"
+                    f" endpoint function: {endpoint}.\n"
+                    "Use 'route(... , endpoint=\"other_name\")' to"
+                    " specify a different name."
                 )
             self.view_functions[endpoint] = view_func
 


davidism:
I updated this to be less verbose, but now that I look at it again I don't think it addresses the issue. The vast majority of the time, this comes up because the user either doesn't use `functools.wraps/update_wrapper` to preserve a decorated function, or is trying to create endpoints programmatically and using the same name for each. Telling them to write `endpoint="other"` in either case won't fix the issue with their code.

PR #4268, opened 2021-09-22
Fix grammatical error

Jalkhov:
Fix grammatical error in docs

### DIFF
--- a/src/flask/scaffold.py
+++ b/src/flask/scaffold.py
@@ -589,8 +589,8 @@ class Scaffold:
         stack of active contexts.  This becomes relevant if you are using
         such constructs in tests.
 
-        Teardown functions must avoid raising exceptions, since they . If they
-        execute code that might fail they
+        Teardown functions must avoid raising exceptions, because if
+        they execute code that might fail they
         will have to surround the execution of these code by try/except
         statements and log occurring errors.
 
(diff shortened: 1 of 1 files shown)


src/flask/scaffold.py
```diff
@@ -589,8 +589,8 @@ def teardown_request(self, f: TeardownCallable) -> TeardownCallable:
         stack of active contexts.  This becomes relevant if you are using
         such constructs in tests.
 
-        Teardown functions must avoid raising exceptions, since they . If they
-        execute code that might fail they
+        Teardown functions must avoid raising exceptions, because if
```
davidism, line 592: Sounds better with a stop: "...raising exceptions. If they..."

src/flask/scaffold.py
```diff
@@ -589,7 +589,7 @@ def teardown_request(self, f: TeardownCallable) -> TeardownCallable:
         stack of active contexts.  This becomes relevant if you are using
         such constructs in tests.
 
-        Teardown functions must avoid raising exceptions, because if
+        Teardown functions must avoid raising exceptions. If
         they execute code that might fail they
         will have to surround the execution of these code by try/except
```
davidism, line 594: "that code" instead of "these code"

src/flask/scaffold.py
```diff
@@ -589,7 +589,7 @@ def teardown_request(self, f: TeardownCallable) -> TeardownCallable:
...
         statements and log occurring errors.
```
davidism, line 595: "any errors" instead of "occurring errors"

src/flask/scaffold.py
Jalkhov: Ready

PR #4306, opened 2021-10-19
Fix Blueprint.errorhandler restricts type of result function which avoids chaining

### LINKED ISSUE #4297: Blueprint.errorhandler restricts type of result function which avoids chaining
In flask 2.0.1 it was possible to chain multiple uses of `Blueprint.errorhandler` as a decorator. In 2.0.2 the type of `Blueprint.errorhandler` is too strict to allow this. Let's assume you start out with a plain handler function of type `ErrorHandlerCallable[Exception]` or similar, applying `some_blueprint.errorhandler(SomeExceptionA)` to it, changes the type of the handler to `ErrorHandlerCallable[SomeExceptionA]`. Applying `some_blueprint.errorHandler(SomeExceptionB)` is now no longer possible, because it expects a `ErrorHandlerCallable[SomeExceptionB]` which the existing error handler cannot fullfill.

Consider this example program:

```Python
from flask import Blueprint

blueprint = Blueprint('example', __name__)

class ExceptionA(Exception):
    pass

class ExceptionB(Exception):
    pass

@blueprint.errorhandler(ExceptionB)
@blueprint.errorhandler(ExceptionA)
def error(e):
    return "example"
```

with flask 2.0.1 this typechecked, with flask 2.0.2 I get the following error:

```
$ mypy test.py
test.py:11: error: Argument 1 has incompatible type "ErrorHandlerCallable[ExceptionA]"; expected "ErrorHandlerCallable[ExceptionB]"
Found 1 error in 1 file (checked 1 source file)
```

I would expect this example to work in 2.0.2 just as well as it did in 2.0.1.

I think the semantically most correct way to fix this, would be to to give `errorhandler` a type smilar to `Callable[[SomeOtherGenericException], Callable[[ErrorHandlerCallable[GenericException]], ErrorHandlerCallable[GenericException]]` where `SomeOtherGenericException` would be a type variable that would always need to be a subtype of `GenericException`, but I am not sure if Python's type system can express relations between type variables like that.

Environment:

- Python version: 3.9.7
- Flask version: 2.0.2

Rohan-Salwan:
- fixed #4297 

Checklist:

- [x] I did tests that demonstrate the correct behavior of the change. Tests should fail without the change.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Released 2021-10-04
 -   Fix the order of ``before_request`` and other callbacks that trigger
     before the view returns. They are called from the app down to the
     closest nested blueprint. :issue:`4229`
+-   Fix ``Blueprint.errorhandler`` restricts type of result function which
+    avoids chaining. :issue:`4297`
 
 
 Version 2.0.1
--- a/src/flask/scaffold.py
+++ b/src/flask/scaffold.py
@@ -80,7 +80,7 @@ class Scaffold:
     name: str
     _static_folder: t.Optional[str] = None
     _static_url_path: t.Optional[str] = None
-
+    T = t.TypeVar("T", t.Callable[..., t.Any], t.Any)
     #: JSON encoder class used by :func:`flask.json.dumps`. If a
     #: blueprint sets this, it will be used instead of the app's value.
     json_encoder: t.Optional[t.Type[JSONEncoder]] = None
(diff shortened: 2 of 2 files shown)


davidism:
This appears to revert the changes from #4097, which were addressing #4095. It also essentially erases all type information, including the number of arguments to the function, which is not an improvement.

PR #4332, opened 2021-11-07
Correct the actual unescaped character

vinzid:
`&raquo;` should be unescaped to `»` after `striptags`.

Ref: https://flask.palletsprojects.com/en/2.0.x/api/#flask.Markup.striptags


davidism:
Since this is a documentation fix, please rebase it to 2.0.x, the latest maintenance branch.

davidism:
Also, `\xbb` is equivalent, try `print('\xbb')`.

vinzid:
@davidism Thank you, I've change the branch to base on 2.0.x.
Although `\xbb` and `»` are equivalent, `»` is the actual output, and I think it's more suitable to use it here as other examples, especially for newbie to understand.

PR #4334, opened 2021-11-11
Fix after_this_request fails with a meaningless error when used outside request context

### LINKED ISSUE #4333: `after_this_request` fails with a meaningless error when used outside request context
https://github.com/pallets/flask/blob/9039534eee6a87da98a1dee9e4338d1b73e861f8/src/flask/ctx.py#L133

While it makes no sense to use it outside a context, the error message could be nicer when you do so anyway (e.g. because you accidentally call a function using it from a celery task or some other place without a request context), like e.g. in `copy_current_request_context` [in the same file](https://github.com/pallets/flask/blob/9039534eee6a87da98a1dee9e4338d1b73e861f8/src/flask/ctx.py#L161-L167).

```
AttributeError: 'NoneType' object has no attribute '_after_request_functions'
  ...
  File "flask/ctx.py", line 133, in after_this_request
    _request_ctx_stack.top._after_request_functions.append(f)
```

Rohan-Salwan:
- fixed #4333 

Checklist:

- [x] I did tests that demonstrate the correct behavior of the change. Tests should fail without the change.

### DIFF
--- a/src/flask/ctx.py
+++ b/src/flask/ctx.py
@@ -130,7 +130,14 @@ def after_this_request(f: AfterRequestCallable) -> AfterRequestCallable:
 
     .. versionadded:: 0.9
     """
-    _request_ctx_stack.top._after_request_functions.append(f)
+    try:
+        _request_ctx_stack.top._after_request_functions.append(f)
+    except Exception:
+        print(
+            "This decorator can only be used at local scopes "
+            "when a request context is on the stack.  For instance within "
+            "view functions."
+        )
     return f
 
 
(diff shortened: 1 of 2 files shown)


src/flask/ctx.py
```diff
@@ -130,7 +130,14 @@ def add_header(response):
 
     .. versionadded:: 0.9
     """
-    _request_ctx_stack.top._after_request_functions.append(f)
+    try:
+        _request_ctx_stack.top._after_request_functions.append(f)
+    except Exception:
```
ThiefMaster, line 135: Do not catch all exceptions wildly (and neither just swallow them, it should remain an exception). I pointed out code in another function that does the right thing - simply do the same here.

ThiefMaster (changes requested):

PR #4342, opened 2021-11-16
deprecate `RequestContext.g`

davidism:
It was moved to `AppContext.g` a long time ago.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,8 @@ Unreleased
     -   ``add_etags`` is renamed to ``etag``.
     -   ``filename`` is renamed to ``path``.
 
+-   The ``RequestContext.g`` property is deprecated. Use ``g`` directly
+    or ``AppContext.g`` instead. :issue:`3898`
 -   ``copy_current_request_context`` can decorate async functions.
     :pr:`4303`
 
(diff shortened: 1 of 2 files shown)


laggardkernel:
@davidism The proposal started as a question from me at first. But after a 2nd thought. I guess we can just keep the property `RequestContext.g`.

If we follow the idea to access attributes only where it's stored. The `.app` should not kept on `RequestContext` neither. Obviously, we can't remove `.app` from `RequestContext` cause it's also used in other methods on `RequestContext`, like `.push()`. This makes me doubt my former proposal. And keeping `.g` makes `RequestContext` a handy place to access all the globals (app, g, request, session). I tried to start a discussion with you maintainers but haven't got any reply yet, only find a PR with this change got merged today.

davidism:
`g` is not bound to the request context, it doesn't make sense as a property there. Additionally, the property always points at the top of the app context stack, not the context that was associated with the specific request context. I'm fine with the change, otherwise I wouldn't have made it.

laggardkernel:
Whether keeping `g` on `RequestContext` or not is totally okay for me. I couldn't make the decision before which is the reason I replied in the original issue. 

> the property always points at the top of the app context stack, not the context that was associated with the specific request context.

Thanks for pointing it out. I didn't consider the problem from this aspect.

PR #4346, opened 2021-11-18
Update method to activate the environment

miftahafina:
Update method to activate the environment using ```source``` instead of ```.``` (dot)


davidism:
Both are correct.

PR #4348, opened 2021-11-19
Fix error message for `after_this_request` when used outside request context

### LINKED ISSUE #4333: `after_this_request` fails with a meaningless error when used outside request context

Yourun-proger:
- fixes #4333

Checklist:

### DIFF
--- a/src/flask/ctx.py
+++ b/src/flask/ctx.py
@@ -130,7 +130,15 @@ def after_this_request(f: AfterRequestCallable) -> AfterRequestCallable:
 
     .. versionadded:: 0.9
     """
-    _request_ctx_stack.top._after_request_functions.append(f)
+    top = _request_ctx_stack.top
+    if top is None:
+        raise RuntimeError(
+            "This decorator can only be used at local scopes "
+            "when a request context is on the stack. For instance within "
+            "view functions."
+        )
+    top._after_request_functions.append(f)
+
     return f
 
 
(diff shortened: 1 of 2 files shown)


src/flask/ctx.py
```diff
@@ -130,7 +130,15 @@ def add_header(response):
...
     .. versionadded:: 0.9
     """
+    top = _request_ctx_stack.top
+    if top is None:
+        raise RuntimeError(
+            "This decorator can only be used at local scopes "
+            "when a request context is on the stack. For instance within "
+            "view functions."
+        )
     _request_ctx_stack.top._after_request_functions.append(f)
```
Aditya-Gupta1, line 140: You can directly do: `top._after_request_functions.append(f)` since you already defined `top` in line 133. Just a minor suggestion.

Yourun-proger: Thank you for good catch!

Yourun-proger:
Fine. Everything works well. Thanks again!

PR #4350, opened 2021-11-20
Only use a custom JSONDecoder if needed

### LINKED ISSUE #4349: Slow json loading under pypy
For large chunks of JSON data, the following snippet (in e.g. any post request) is very slow under pypy:
```python
request.get_json()
```
in fact it's 4x slower than doing
```python
import json
json.loads(request.get_data())
```

## Expected behaviour

I would expect using the default `get_json` to be the happy path and be as fast as the in-built `json.loads`.

## Underlying cause

I tracked this down to the use of the custom `JSONDecoder` class here: https://github.com/pallets/flask/blob/main/src/flask/json/__init__.py#L53
Essentially pypy has a fast-path for the default `json.loads` - and if you use a hand-rolled decoder class you don't get to use that fast path.
Note that the `JSONDecoder` class does nothing except add some docstrings, but it's very hard for pypy to know that without some horrible introspection of the decoder class every time you call json.loads.

## Possible fix

Only set a custom `cls=` in `flask.json.load(s)` if something other than the default is set on `app.json_decoder`.

## Environment:

- Python version: pypy3.7
- Flask version: 2.0.2

olliemath:
The default JSONDecoder adds no extra functionality to `json.loads`, but using a custom class takes you off of an optimized fast path in pypy (it adds some overhead in cpython, but I think it's minimal). This change sticks with the default loading behaviour unless the user has specifically opted in to customising it.

- fixes #4349

Checklist:

Notes:
- I can update the above if the general outline of the fix is acceptable
- Testing performance is hard: but I can add a test to ensure the correct args are passed to `json.loads` in all cases.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@ Unreleased
 -   Fix type annotation for ``errorhandler`` decorator. :issue:`4295`
 -   Revert a change to the CLI that caused it to hide ``ImportError``
     tracebacks when importing the application. :issue:`4307`
+-   ``app.json_encoder`` and ``json_decoder`` are only passed to
+    ``dumps`` and ``loads`` if they have custom behavior. This improves
+    performance, mainly on PyPy. :issue:`4349`
 
 
 Version 2.0.2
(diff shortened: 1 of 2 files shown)


davidism:
I extended this to skip `json.JSONDecoder` as well, and apply to `dumps` with `json.JSONEncoder`, added a changelog, and rebased to 2.0.x.

PR #4368, opened 2021-12-02
Update requirements

greyli:


davidism:
I've started updating in 2.0.x instead of main, then merging forward, so that we are up to date on the support  too.

PR #4370, opened 2021-12-02
Don't error when MethodView is a Metaclass base (fixes #4369)

### LINKED ISSUE #4369: Unable to use MethodView as a Metaclass base
According to [Python docs](https://docs.python.org/3/reference/datamodel.html#determining-the-appropriate-metaclass)

> If the metaclass has a `__prepare__` attribute, it is called as `namespace = metaclass.__prepare__(name, bases, **kwds)` (where the additional keyword arguments, if any, come from the class definition)

> Once the class namespace has been populated by executing the class body, the class object is created by calling `metaclass(name, bases, namespace, **kwds)` (the additional keywords passed here are the same as those passed to `__prepare__`).

`MethodViewType` (and therefore `MethodView`) lacks this `**kwds`
https://github.com/pallets/flask/blob/9486b6cf57bd6a8a261f67091aca8ca78eeec1e3/src/flask/views.py#L106-L111

So the the three argument form of [`type`](https://docs.python.org/3/library/functions.html#type) won't work when supplied with `**kwargs` if `MethodView` is a base.

```python


class PatchedMethodViewType(type):
    def __init__(cls, name, bases, d, **kwargs):
        super().__init__(name, bases, d)


class PatchedMethodView(View, metaclass=PatchedMethodViewType):
    def dispatch_request(self, *args, **kwargs):
        pass


class MyMethodView(MethodView):
    @classmethod
    def __init_subclass__(cls, author=None):
        cls.author = author

    def get(self):
        return 1


class MyPatchedMethodView(PatchedMethodView):
    @classmethod
    def __init_subclass__(cls, author=None):
        cls.author = author

    def get(self):
        return 1

type('MetaMethodView', (MyMethodView,), {})
# __main__.MetaMethodView

type('MetaMethodView', (MyMethodView,), {}, author='bob')
# TypeError: __init__() got an unexpected keyword argument 'author'

type('MetaMethodView', (MyPatchedMethodView,), {}, author='bob')
# __main__.MetaMethodView
```

reubano:
`MethodViewType`'s init method lacks `**kwargs` so is unable to be used as a Metaclass base when supplying `**kwds` in the three argument form of [`type`](https://docs.python.org/3/library/functions.html#type).

- fixes #4369

Checklist:

### DIFF
--- a/src/flask/views.py
+++ b/src/flask/views.py
@@ -108,7 +108,7 @@ class MethodViewType(type):
     defines.
     """
 
-    def __init__(cls, name, bases, d):
+    def __init__(cls, name, bases, d, **kwargs):
         super().__init__(name, bases, d)
 
         if "methods" not in d:


davidism:
> is unable to be used as a Metaclass base

I don't understand this phrase. I don't think it matches an intended use case of `MethodView`. Can you show a specific example of what you're attempting to do that is not working?

davidism:
See discussion in issue.

PR #4380, opened 2021-12-11
Remove python 2 syntax (fixes #4379)

### LINKED ISSUE #4379: The documentation's tutorial uses python2 syntax in an example
I was reading the documentation, and noticed that one of the examples creates a class that inherits from object. Here: https://flask.palletsprojects.com/en/2.0.x/tutorial/tests/#id6

BrickGrass:


ThiefMaster:
You can remove the `()` as well; and please send docs fixes against the active maintenance branch

PR #4382, opened 2021-12-12
Fix the command for checking port on macOS in server docs

greyli:
The original `netstat` command is not working on macOS, I change it to `lsof` for a better output.

Continues #4343

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -80,27 +80,34 @@ messages:
 Either identify and stop the other program, or use
 ``flask run --port 5001`` to pick a different port.
 
-You can use ``netstat`` to identify what process id is using a port,
-then use other operating system tools stop that process. The following
-example shows that process id 6847 is using port 5000.
+You can use ``netstat`` or ``lsof`` to identify what process id is using
+a port, then use other operating system tools stop that process. The
+following example shows that process id 6847 is using port 5000.
 
 .. tabs::
 
-    .. group-tab:: Linux/Mac
+    .. tab:: ``netstat`` (Linux)
 
         .. code-block:: text
 
             $ netstat -nlp | grep 5000
             tcp 0 0 127.0.0.1:5000 0.0.0.0:* LISTEN 6847/python
 
-    .. group-tab:: Windows
+    .. tab:: ``lsof`` (macOS / Linux)
+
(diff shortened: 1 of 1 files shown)


davidism:
From my understanding, `lsof` is not available by default on MacOS. What is failing in the `netstat` command, can the options be changed?

greyli:
> What is failing in the netstat command, can the options be changed?

The `-p` option means protocol on macOS, and there isn't a `-t` option:

```
$ netstat -nlp | grep 5000
netstat: option requires an argument -- p
Usage:	netstat [-AaLlnW] [-f address_family | -p protocol]
	netstat [-gilns] [-f address_family]
	netstat -i | -I interface [-w wait] [-abdgRtS]
	netstat -s [-s] [-f address_family | -p protocol] [-w wait]
	netstat -i | -I interface -s [-f address_family | -p protocol]
	netstat -m [-m]
	netstat -r [-Aaln] [-f address_family]
	netstat -rs [-s]
```

It should be:

```
$ netstat -anlp tcp | grep 5000
tcp4       0      0  127.0.0.1.5000         *.*                    LISTEN
tcp6       0      0  *.5000                 *.*                    LISTEN
tcp4       0      0  *.5000                 *.*                    LISTEN
```

But it didn't show the process name or PID, so I think it's better to use `lsof`.

> From my understanding, lsof is not available by default on MacOS.

From [this answer](https://apple.stackexchange.com/questions/120991/which-mac-os-x-versions-support-lsof-command):

> `lsof` has been available since OS X 10.5 (2007). From the manual page :

    Lsof  revision 4.87 lists on its standard output file information about
    files opened by processes for the following UNIX dialects:

    Apple Darwin 9 and Mac OS X 10.[567]

greyli:
Oops, the typo fix from 2.0.x was overwritten. I will rebase it later.

davidism:
Which typo? I was about to merge this.

greyli:
got -> go in [this line](https://github.com/pallets/flask/pull/4382/files#diff-2e8b3bd0cd400f73069709441b7f7f5499adb1c247b191088ec7e89ee3a2cbacR111).

davidism:
Never mind, I found it, I'll fix it.

PR #4383, opened 2021-12-15
Added docstrings to several undocumented functions

fluencydoc:
Documented various functions that we found particularly lacking of docstrings.


davidism:
Thanks, however I don't see the utility in adding these. None of the functions are public API or even called directly, two of them are even function-level helpers. I'm happy to consider improvements to existing docs and public API.

ThiefMaster:
also the "written by trelent" part in each docstring is pure spam.

fluencydoc:
> also the "written by trelent" part in each docstring is pure spam.

We prefer to credit our work but understand if this is too visible. Were these changes accepted, we would be happy to remove this. Alas, thank you for the feedback!

PR #4391, opened 2021-12-28
Lint the code-block in README.rst

baseplate-admin:
Hi there,

Currently we have [```.. code-block:: text in line 46```](https://github.com/pallets/flask/blob/660994efc761efdfd49ca442b73f6712dc77b6cf/README.rst#L44).

But it would be better linted with ```.. code-block:: python```  ( In my opinion )

# Before Pic :

# After Pic:

### This pull request will essentially lint the README code-block a little better

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,7 +43,7 @@ A Simple Example
     def hello():
         return "Hello, World!"
 
-.. code-block:: text
+.. code-block:: python
 
     $ flask run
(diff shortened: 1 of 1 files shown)


davidism:
That's not Python though, that's terminal output.

baseplate-admin:
I have renamed python to bash which pygments seems to [have](https://pygments.org/docs/lexers/) a lexer on

Issue #4392, opened 2021-12-28
don't recommend using `X-XSS-Protection`


mousetail:
The [security page](https://flask.palletsprojects.com/en/2.0.x/security/) of the docs recommends using X-XSS-Protection to prevent against XSS attacks. However, this is no longer recommended since it protects only against extremely rudimentary XSS attacks while at the same time causing while new types of more difficult to prevent XSS attacks.

See [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#x-xss-protection-header) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection) for reasons why this is a problem.

This section should at least include a warning in case someone blindly copy pastes that code.

davidism:
Go ahead and make a PR to remove it if it's no longer relevant to any supported browsers.

PR #4397, opened 2021-12-31
Remove X-XSS-Protection suggestion

cameronthecoder:
According to [caniuse.com](https://caniuse.com/?search=X-XSS-Protection), this header is not supported by all major browsers except Safari and MDN includes this warning in their [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection):

> Warning: Even though this feature can protect users of older web browsers that don't yet support CSP, in some cases, XSS protection can create XSS vulnerabilities in otherwise safe websites. See the section below for more information. 

This pull request addresses #4392

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -173,18 +173,6 @@ invisibly to clicks on your page's elements. This is also known as
 
 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options
 
-X-XSS-Protection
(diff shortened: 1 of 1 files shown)


davidism:
When making PRs, be sure to make them from a new branch. Since this was a docs PR, I had to rebase onto 2.0.x and force push that to your main branch. After merging, to get your main back to tracking our main, you'll need to hard reset it.

PR #4407, opened 2022-01-05
Add no_auto_head option in app route decorator

### LINKED ISSUE #4395: Endpoint registered for the HEAD method not called
Hi,

I am trying to implement a HEAD method for a route already having a GET route. The goal is to just return the headers without doing heavy computations to build the body. I know Werkzeug already provides a HEAD method, but I want to skip the heavy computations of the GET method as the body is not needed to answer quickly with just the headers. Actually the Werkzeug behavior is causing me some troubles.

Here is an exemple application:
```Python
from flask import Flask

app = Flask(__name__)

@app.route("/", methods=["GET"])
def hello_world():
    return "<p>Hello, World!</p>"

@app.route("/", methods=["HEAD"])
def hello_head():
    return '', 200, {"some-header": "some-value"}
```

This code is not HTTP compliant (both the GET and the HEAD methods do not return the same headers) but that is not relevant to see my problem.

I then run the app:
```
~ via 🐍 v3.9.9 on ☁️  erdnaxeli@myhost
❯ export FLASK_APP=t

~ via 🐍 v3.9.9 on ☁️  erdnaxeli@myhost
❯ flask run
 * Serving Flask app 't' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off

```

Then here is the response I got with on a HEAD query:
```
❯ curl -vvv -X HEAD localhost:5000
Warning: Setting custom HTTP method to HEAD with -X/--request may not work the 
Warning: way you want. Consider using -I/--head instead.
*   Trying ::1:5000...
* connect to ::1 port 5000 failed: Connexion refusée
*   Trying 127.0.0.1:5000...
* Connected to localhost (127.0.0.1) port 5000 (#0)
> HEAD / HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.76.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text/html; charset=utf-8
< Content-Length: 20
< Server: Werkzeug/2.0.2 Python/3.9.9
< Date: Wed, 29 Dec 2021 14:37:55 GMT
< 
* transfer closed with 20 bytes remaining to read
* Closing connection 0
curl: (18) transfer closed with 20 bytes remaining to read
```

We can notice the "some-header" header is missing.

Environment:

```
❯ flask --version
Python 3.9.9
Flask 2.0.2
Werkzeug 2.0.2
```

Some real world details. I am trying to do this on a big flask applications with tens of routes (129 rules in the `MapAdapter.map._rules` object). If I explore the rules I can see my two rules: one for the GET method with GET / HEAD / OPTIONS (HEAD being added by Werkzeug, OPTIONS by Flask) and one for my HEAD method with HEAD / OPTIONS. By changing _things_ unrelated in my code (like adding an element to a list that had nothing to do with Flask or Werkzeug or even just adding `pdb.set_trace()` in some location) it seems I can have an influence on which rule will be find first by werkzeug. I didn't figured out why.

Ideally I would like a way to disable the automatically added HEAD method to my GET endpoint. This could be detected by Flask / Werkzeug or could be an option on the Flask decorator.

malwaredllc:
Fixes #4395 

Adds ```no_auto_head``` as an optional keyword argument to the app route decorator to optionally prevent Werkzeug from automatically adding a HEAD method to a route that only has a GET method specified. This flexibility will allow the user to implement their own logic to handle HEAD requests separately, thus addressing the issue described in #4395.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,10 @@ Unreleased
     or ``AppContext.g`` instead. :issue:`3898`
 -   ``copy_current_request_context`` can decorate async functions.
     :pr:`4303`
+-   ``no_auto_head`` keyword argument added to ```app.route``` decorator
+    to optionally prevent Werkzeug from automatically adding a HEAD method
+    to a route that only has a GET method specified, allowing the user to
+    implement their own logic to handle HEAD requests. :pr:`4407`
 
 
 Version 2.0.3
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -697,7 +697,7 @@ instead of the `view_func` parameter.
                 of methods this rule should be limited to (``GET``, ``POST``
                 etc.).  By default a rule just listens for ``GET`` (and
                 implicitly ``HEAD``, although this can be disabled with the
-                keyword argument ```no_auto_head=True```).  Starting with
+                keyword argument ``no_auto_head=True``).  Starting with
                 Flask 0.6, ``OPTIONS`` is implicitly added and handled by
                 the standard request handling.  They have to be specified
                 as keyword arguments.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -696,9 +696,11 @@ instead of the `view_func` parameter.
                 Werkzeug is handling of method options.  methods is a list
                 of methods this rule should be limited to (``GET``, ``POST``
                 etc.).  By default a rule just listens for ``GET`` (and
-                implicitly ``HEAD``).  Starting with Flask 0.6, ``OPTIONS`` is
-                implicitly added and handled by the standard request
-                handling.  They have to be specified as keyword arguments.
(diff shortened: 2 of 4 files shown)


malwaredllc:
@davidism would appreciate your thoughts/review on this (see linked issue #4395 for context). I followed the contribution checklist and everything is good to go.

davidism:
Thanks for looking at this. However, I've decided to close the referenced issue, see my comment there. Additionally, if I had wanted to accept this, it doesn't belong in Flask, it would be a change to Werkzeug's `Rule` class.

PR #4454, opened 2022-02-17
Fix typing on `Flask.session_interface`

### LINKED ISSUE #4453: Missing type annotation on `Flask.session_interface`
https://github.com/pallets/flask/blob/main/src/flask/app.py#L377-L381

This leads type checkers (`mypy` in our case) to complain if we try to assign something to `session_interface` that isn't `SecureCookieSessionInterface`, even if it implements `SessionInterface`.

To reproduce, create a custom session interface that extends `SessionInterface`. Set `Flask.session_interface` to an instance of this. Then run `mypy` on this file.

The type annotation should allow other session interface implementations to be used.

Environment:

- Python version: 3.8
- Flask version: 2.0.2

colincadams:
This creates an explicit type annotation for `Flask.session_interface` that allows any session interface implementation to be used.

- fixes #4453

Checklist:

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -50,6 +50,7 @@ from .scaffold import find_package
 from .scaffold import Scaffold
 from .scaffold import setupmethod
 from .sessions import SecureCookieSessionInterface
+from .sessions import SessionInterface
 from .signals import appcontext_tearing_down
 from .signals import got_request_exception
 from .signals import request_finished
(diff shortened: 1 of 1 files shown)


colincadams:
Given how small this change is, I wasn't sure what from the checklist I should actually follow through on. Let me know and I'm happy to take care of them!

PR #4465, opened 2022-02-23
fix: update reference to 'query' parameter

### LINKED ISSUE #4464: docs: 'Sending Requests with the Test Client' references non-existent 'query' parameter
In the [Sending Requests with the Test Client](https://flask.palletsprojects.com/en/2.0.x/testing/#sending-requests-with-the-test-client) section of the documentation it's said that:

> The `client` has methods that match the common HTTP request methods, such as `client.get()` and `client.post()`. They take many arguments for building the request; you can find the full documentation in [EnvironBuilder](https://werkzeug.palletsprojects.com/en/2.0.x/test/#werkzeug.test.EnvironBuilder). Typically you’ll use `path`, `query`, `headers`, and `data` or `json`.

And also:

> Pass a dict `query={"key": "value", ...}` to set arguments in the query string (after the ? in the URL). Pass a dict `headers={}` to set request headers.

But looking at the linked [EnvironBuilder class](https://werkzeug.palletsprojects.com/en/2.0.x/test/#werkzeug.test.EnvironBuilder) it's clear that there is no `query` parameter, but rather `query_string`, which does not take a `dict`. Passing `query` instead of `query_string` will result in the following error being thrown:

```text
TypeError: __init__() got an unexpected keyword argument 'query'
```

I only ran into this because I was following the documentation and didn't think to double-check the API reference.

Environment:

- Python version: 3.9.10
- Flask version: 2.0.1

usrme:
Also added mention of being able to pass string, not only just a dictionary.

Fixes #4464.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
         name: Reorder Python imports (src, tests)
         files: "^(?!examples/)"
         args: ["--application-directories", "src"]
+        additional_dependencies: ["setuptools>60.9"]
   - repo: https://github.com/psf/black
     rev: 22.1.0
     hooks:


greyli:
Thanks for working on this! Please always use a new branch for submitting PR (so that you can keep your main branch clean and synced with the main branch of upstream), and the docs fix should base on the latest .x branch (2.0.x).

To keep things easier, I would recommend recreating a PR, then do the following to create the branch:

```
git checkout -b fix-testing-docs origin/2.0.x
```

BTW, be sure to keep the line lengths consistent ([see this line](https://github.com/pallets/flask/pull/4465/files#diff-3d0f12620667116255f8af843160db5736519a740c49fbf2d727dea599f9a725R111)).

Check out the [contributing guide](https://github.com/pallets/flask/blob/main/CONTRIBUTING.rst) for more details.

davidism:
I'll fix it tomorrow morning, sorry for the confusion.

davidism:
Redone in https://github.com/pallets/flask/pull/4467

PR #4479, opened 2022-03-08
Allow loading of environment variables into the config

pgjones:
This new method will pick out any environment variables with a certain
prefix and place them into the config named without the prefix. This
makes it easy to use environment variables to configure the app as is
now more popular than when Flask started.

The prefix should ensure that the environment isn't polluted and the
config isn't polluted by environment variables.

Checklist:

### DIFF
--- a/src/flask/config.py
+++ b/src/flask/config.py
@@ -78,7 +78,7 @@ class Config(dict):
     """
 
     def __init__(self, root_path: str, defaults: t.Optional[dict] = None) -> None:
-        dict.__init__(self, defaults or {})
+        super().__init__(defaults or {})
         self.root_path = root_path
 
     def from_envvar(self, variable_name: str, silent: bool = False) -> bool:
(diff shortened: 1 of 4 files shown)


davidism:
I think only accepting strings is too big a limitation, given that it is possible to convert types in all other config loading methods. Attempting to convert at least int, float, and bool would cover a lot of cases.

It would be nice to also accept list and dict, and maybe use `__` double underscore for nested access.

Maybe this should take a `convert_value` function similar to how `from_file` takes a format loading function, so that users can override it further?

src/flask/config.py
```diff
@@ -97,6 +97,31 @@ def from_envvar(self, variable_name: str, silent: bool = False) -> bool:
             )
         return self.from_pyfile(rv, silent=silent)
 
+    def from_prefixed_env(self, prefix: str = "FLASK_") -> bool:
```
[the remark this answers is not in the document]
pgjones: I'm not sure here, although I agree dynaconf seems to have set the president of the `_` being implied. I'd rather keep it explicit, but I'm happy either way.

pgjones:
I agree with the conversion, I've followed dynaconf's convention (although simplified) and used json.loads (dynaconf uses toml). I've also made it configurable via a `loads` argument.

src/flask/config.py
```diff
@@ -97,6 +105,44 @@ def from_envvar(self, variable_name: str, silent: bool = False) -> bool:
...
+        :return: Always returns ``True``.
+
+        .. versionadded:: 2.1.0
+
+        """
+        mapping = {}
+        for raw_key, value in os.environ.items():
+            if raw_key.startswith(prefix):
+                key = raw_key[len(prefix) :]  # Use removeprefix with Python 3.9
+                mapping[key] = loads(value)
```
davidism, line 142: I think the `try/except Excetion` should be here so that users don't have to write wrappers for any library they decide to use.

davidism:
I'm working on finalizing this, including the review I posted above, no need to push anything new.

davidism:
I've added support for nested dict access using the `__` separator.

If an intermediate key doesn't exist, it will be initialized to an empty dict. This matches how top-level keys work (any config is added, it doesn't already have to be in the config dict), and allows specifying configuration for extensions without requiring calling `init_app` first, and without requiring `init_app` to add all possible config keys.

I did not add support for merging dicts or other nested access like lists or objects, or other advanced features. If even more control of config is needed, users should reach for Dynaconf or another dedicated config library.

ThiefMaster:
Do we really need/want this level of magic?

davidism:
I use it _constantly_ at work, it is really handy for being able to configure deployments in Docker and other platform services. Nested access is nice for grouping config together rather than making it a flat namespace with long individual keys. Flask-SQLAlchemy, for example, has a dict as a config value for engine options.

tests/test_config.py
```diff
@@ -38,6 +38,68 @@ def test_config_from_file():
...
+    app = flask.Flask(__name__)
+    app.config.from_prefixed_env()
+
+    assert app.config["STRING"] == "value"
+    assert app.config["BOOL"] is True
+    assert app.config["INT"] == 1
+    assert app.config["FLOAT"] == 1.2
+    assert app.config["LIST"] == [1, 2]
+    assert app.config["DICT"] == {"k": "v"}
+    assert "OTHER" not in app.config
```
Yourun-proger, line 59: This is probably what was meant here:
```py
assert "NOT_FLASK_OTHER" not in app.config
```
Just some guess

davidism: `NOT_FLASK` is a prefix, `OTHER` is the config key (or not, in this case).

Yourun-proger: uh. I don't know what to say but let it be

Issue #382, opened 2012-01-17
Set TRAP_BAD_REQUEST_ERRORS to True as the default when debugging is enabled


catlee:
When debugging an application, it's surprising to get a simple "400 Bad Request" when accessing a form parameter that doesn't exist instead of a full traceback. I always forget that I also need to set TRAP_BAD_REQUEST_ERRORS.

Is there a reason why this isn't set by default when debugging is enabled?

Tjorriemorrie:
@mitsuhiko  Please think positively about it; it just wasted hours of my life :(

jeffwidman:
Tenatively assigned to 1.0 milestone as that's the logical time to make a decision around changing this API.

PR #4492, opened 2022-03-23
update some debug message behavior

davidism:
* Don't intercept routing redirects if the code hasn't been changed from 308 (or 307). The message about not preserving the request method and body is not relevant for the modern status code. Also edited the message a bit, although it shouldn't be seen anymore.
* When `files` raises a `DebugFilesKeyError`, it looks like the original error rather than a chained exception.
* `on_json_loading_failed` tries to return the `super()` implementation first, then reraises any error in debug mode.

I originally started working on this due to https://github.com/pallets/werkzeug/pull/2355, because I thought it might be better to not suppress JSON errors outside debug mode. 400 errors indicate a problem with client data, so it seems ok to show what that problem was. I've stopped short of doing that for this release, but want to revisit suppressing information in 400 errors later.

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -1457,17 +1457,26 @@ class Flask(Scaffold):
         )
 
     def raise_routing_exception(self, request: Request) -> "te.NoReturn":
-        """Exceptions that are recording during routing are reraised with
-        this method.  During debug we are not reraising redirect requests
-        for non ``GET``, ``HEAD``, or ``OPTIONS`` requests and we're raising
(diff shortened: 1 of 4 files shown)


davidism:
For future reference, previous tickets related to hiding 400 messages:

* https://github.com/pallets/flask/issues/382
* https://github.com/pallets/werkzeug/pull/1592
* https://github.com/pallets/werkzeug/pull/1395
* https://github.com/pallets/click/issues/2146
* https://github.com/pallets/flask/pull/2348

PR #4503, opened 2022-03-29
Fix werkzeug updating and causing an issue with EnvironBuilder

farooq-bboxx:
EnvironBuilder no longer accepts `as_tuple` in 2.1.0 but since we have `>= 2.0` in `install_requires`, it updates without us noticing it and builds can fail suddenly.

Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 setup(
     name="Flask",
     install_requires=[
-        "Werkzeug >= 2.0",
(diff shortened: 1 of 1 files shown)


davidism:
Use a tool like [pip-tools](https://pypi.org/project/pip-tools/) to pin your dependency tree and control when you get updates. Be sure to run your tests with deprecation warnings treated as errors so that you get notified of these types of changes early.

PR #4506, opened 2022-03-29
Update Testing Docs

s2t2:
NOTE: This is a documentation-only change.

Summary:

Sharing pytest fixtures using modular scope helps prevent duplicate initialization. 

See: https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -56,7 +56,7 @@ set up and tear down resources.
     import pytest
     from my_project import create_app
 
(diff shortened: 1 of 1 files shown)


davidism:
This isn't correct, each test is intentionally isolated with a fresh app instance. That's one of the benefits of the app factory pattern.

PR #4513, opened 2022-04-01
Defined constants to reduce duplication

Agaba-Ed:
"localtoc.html" and "ethicalads.html" were repeated 3 times in docs/conf.py. Defined constants to reduce code duplication.


ThiefMaster:
Honestly, I don't see how this really improves much. It's a config file in the end...

davidism:
This type of duplication in this context is not worth refactoring.

PR #4534, opened 2022-04-12
Adapted input parameters type restriction for json.loads function - fixes #4519 

### LINKED ISSUE #4519: Argument 1 to "loads" has incompatible type "bytes"; expected "str"
From the [documentations](https://docs.python.org/3/library/json.html#json.loads), json.loads accepts bytes as an argument, so i'm not sure what's the reasoning behind flask.json annotating it to `str` only. https://github.com/pallets/flask/blob/main/src/flask/json/__init__.py#L156. As a result, mypy is raising an error.

Is this correct?

```python
import requests
from flask.json import loads

# Making a get request
response = requests.get('https://api.github.com')

# With response.content
loads(response.content)
Argument 1 to "loads" has incompatible type "bytes"; expected "str"

# With response.text
loads(response.text)
No issue
```

Environment:

- Python version: 3.8x
- Flask version: 2.1x

rzimmerdev:
Adapted input parameters type restriction for
json.loads function. According to Python 3.6
changes, built-in :func:json.loads now accepts
bytes and bytes array as input.

- fixes #4519 

Checklist:

- [X] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [X] Add or update relevant docs, in the docs folder and in code.
- [X] Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/src/flask/json/__init__.py
+++ b/src/flask/json/__init__.py
@@ -153,7 +153,11 @@ def dump(
     _json.dump(obj, fp, **kwargs)
 
 
-def loads(s: str, app: t.Optional["Flask"] = None, **kwargs: t.Any) -> t.Any:
+def loads(
+    s: t.Union[str, bytes],
+    app: t.Optional["Flask"] = None,
+    **kwargs: t.Any,
+) -> t.Any:
     """Deserialize an object from a string of JSON.
 
     Takes the same arguments as the built-in :func:`json.loads`, with
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 2.1.2
 
 Unreleased
 
+-   Fix type annotation for ``json.loads``, it accepts str or bytes.
+    :issue:`4519`
+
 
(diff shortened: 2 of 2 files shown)


src/flask/json/__init__.py
```diff
@@ -164,6 +168,9 @@ def loads(s: str, app: t.Optional["Flask"] = None, **kwargs: t.Any) -> t.Any:
         or defaults.
     :param kwargs: Extra arguments passed to :func:`json.loads`.
 
+    .. versionchanged:: 2.1
```
davidism, line 171: This doesn't need a docstring changelog.

rzimmerdev:
Okay, thanks for the comment, I'll fix it.

davidism:
It looks like this is still not rebased correctly.

karmingc:
@rzimmerdev any updates on this? @davidism I can make a PR, but not sure what's the standard if somebody already has stale PR...

PR #4560, opened 2022-04-28
Fix old behavior in `register_error_handler()`

### LINKED ISSUE #4559: Fix old behavior in `register_error_handler()`
## Background
[`Line 700 src/scaffold.py`](https://github.com/pallets/flask/blob/f976d5bb88216e921a96998f767df31d7039e4ef/src/flask/scaffold.py#L700) has a comment `# old broken behavior`.  
```python
if isinstance(code_or_exception, HTTPException):  # old broken behavior
    raise ValueError(
        "Tried to register a handler for an exception instance"
        f" {code_or_exception!r}. Handlers can only be"
        " registered for exception classes or HTTP error codes."
    )
```
The purpose of this code is check if the handler is an exception instance.   
If so, raise an error to warn the user: the handler should be HTTPException classes or HTTP error codes.   
### Trigger test case: 
In [`test/test_user_error_handler.py`](https://github.com/pallets/flask/blob/f976d5bb88216e921a96998f767df31d7039e4ef/tests/test_user_error_handler.py), add:
```python
with pytest.raises(ValueError):
    app.register_error_handler(HTTPException(), None)
```
### Why is this old and broken?  
From version 0.7, `self.register_error_handler` can accept all `Exception` not just `HTTPException` so it is not accurate now. Also, the main check part for handlers is in `self._get_exc_class_and_code`.

## Purposed Solution
Remove this part of code.  
Enhance the check for handlers in `self.register_error_handler`.  

## Detail Discussion
- Since the main check is in `self._get_exc_class_and_code`, new check part is written in `self._get_exc_class_and_code` to keep consistency.
- `isinstance(exc_class, Exception)` is kept for warning error-prone cases.
- Idea from [`inspect.isclass`](https://docs.python.org/3/library/inspect.html#inspect.isclass) to solve `Line 735 scr/scaffold.py` `TypeError: issubclass() arg 1 must be a class` when arg1 is a instance.  
This idea uses `New-style Classes` feature in Python3. Since Flask needs Python>=3.7, it should not cause any compatibility issues.
- Move `except KeyError` after confirm `int` type to improve type safety.
- Print `exc_class` when AssertionError to improve traceback.

A PR will be linked.

qingpeng9802:
- Fixes #4559

### DIFF
--- a/src/flask/scaffold.py
+++ b/src/flask/scaffold.py
@@ -697,22 +697,7 @@ class Scaffold:
 
         .. versionadded:: 0.7
         """
-        if isinstance(code_or_exception, HTTPException):  # old broken behavior
-            raise ValueError(
-                "Tried to register a handler for an exception instance"
-                f" {code_or_exception!r}. Handlers can only be"
-                " registered for exception classes or HTTP error codes."
-            )
-
-        try:
-            exc_class, code = self._get_exc_class_and_code(code_or_exception)
-        except KeyError:
-            raise KeyError(
-                f"'{code_or_exception}' is not a recognized HTTP error"
-                " code. Use a subclass of HTTPException with that code"
-                " instead."
-            ) from None
-
+        exc_class, code = self._get_exc_class_and_code(code_or_exception)
         self.error_handler_spec[None][code][exc_class] = f
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


src/flask/scaffold.py
```diff
@@ -728,13 +714,32 @@ def _get_exc_class_and_code(
...
 
+        if isinstance(exc_class, Exception):
+            raise ValueError(
+                "Tried to register a handler for an Exception instance"
+                f" {exc_class!r}. Handlers can only be"
+                " registered for Exception classes or HTTP error codes."
+            )
+
+        # make sure `exc_class` is a class before checking subclass
+        assert isinstance(
```
davidism, line 736: I don't understand why this assert is needed.

tests/test_user_error_handler.py
```diff
@@ -14,15 +14,38 @@ class CustomException(Exception):
     class UnacceptableCustomException(BaseException):
         pass
 
+    ACCEPT_INS_REPR = "Acceptable_Instance_Repr()"
```
davidism, line 17: These constants are unnecessary.

tests/test_user_error_handler.py
```diff
@@ -14,15 +14,38 @@ class CustomException(Exception):
...
         assert isinstance(e, CustomException)
         return "custom"
 
-    with pytest.raises(
-        AssertionError, match="Custom exceptions must be subclasses of Exception."
-    ):
+    with pytest.raises(AssertionError) as e:
         app.register_error_handler(UnacceptableCustomException, None)
+    assert (
+        UnacceptableCustomException.__name__ + " must be a subclass of Exception"
```
davidism, line 37: Use `f""` strings instead of concatenation.

src/flask/scaffold.py
qingpeng9802: This assertion is used to avoid the potential TypeError in `issubclass(exc_class, Exception)` since I found error msg `TypeError: issubclass() arg 1 must be a class` might be confusing to users. If it's deleted, we should be also fine.

PR #4576, opened 2022-05-02
Add 'aborter' and 'aborter_class' attrs to 'Flask' object

### LINKED ISSUE #4567: Move `abort` to the `Flask` app object
Add an `abort` method to the `Flask` app object. Similar to functions like `flask.json.dumps`, `flask.abort` should look for a `current_app` and call its `abort` method. This will allow applications to override the abort behavior.

dzcode:
* Add an `aborter_class` attribute on the `Flask` object, similar to the `config_class`
* Upon `__init__`, create an `aborter` instance of the `aborter_class`
* Rather than exposing the `werkzeug.exceptions.abort` function in `flask` directly, add an `abort` function in `flask.helpers` that calls the `aborter` instance on the `current_app`. This is equivalent to what the `werkzeug.exceptions.abort` function is doing.

- fixes #4567

Checklist:

### DIFF
--- a/src/flask/helpers.py
+++ b/src/flask/helpers.py
@@ -10,6 +10,7 @@ from functools import update_wrapper
 from threading import RLock
 
 import werkzeug.utils
+from werkzeug.exceptions import abort as _wz_abort
 from werkzeug.routing import BuildError
 from werkzeug.urls import url_quote
 from werkzeug.utils import redirect as _wz_redirect
@@ -24,6 +25,7 @@ from .signals import message_flashed
 if t.TYPE_CHECKING:  # pragma: no cover
     from werkzeug.wrappers import Response as BaseResponse
     from .wrappers import Response
+    import typing_extensions as te
 
 
 def get_env() -> str:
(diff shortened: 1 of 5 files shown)


src/flask/helpers.py
```diff
@@ -788,3 +789,23 @@ def _split_blueprint_path(name: str) -> t.List[str]:
...
+    it will wrap it in a proxy WSGI exception and raise that::
+
+       abort(404)  # 404 Not Found
+       abort(Response('Hello World'))
+
+    .. versionchanged:: 2.2
+       Call :attr:`flask.Flask.aborter` on the :data:`~flask.current_app`
+       rather than exposing :class:`werkzeug.exceptions.abort` directly.
+    """
+    current_app.aborter(status, *args, **kwargs)  # type: ignore[misc]
```
dzcode, line 811: The typing error here is: `src/flask/helpers.py:794: error: Implicit return in function which does not return  [misc]`.

The `aborter` is an instance of the `Aborter` class, which is [properly annotated to `NoReturn`](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/exceptions.py#L851-L853) on it's `__call__` function. This seems like potentially a type-checking bug, unless I'm missing something

PR #4577, opened 2022-05-02
update setupmethod behavior

### LINKED ISSUE #4571: Unify `setupmethod` behavior between `Flask` and `Blueprint`
Many of the methods in the `Flask` app object are decorated with `setupmethod`, which prevents registering things with the application after it starts running. A similar behavior is in `Blueprint`, we want to prevent modifying it after registering it at least once. However, the blueprint checking is much less consistent, it's divided up into two parts, is only a warning sometimes, and has a different message. Figure out how we can make `Flask` and `Blueprint` more consistent with this error.

hallacy:
Addresses #4571 by 
* Removing `warn_on_modifications` from blueprint.py so that warnings aren't ignorable
* Adds a TODO and modifies the warning to inform users that the warning from blueprints.py will be an error in 2.3
* Adds @setupmethod calls to `route` and `endpoint` to check places where reregistering would be bad.

- fixes #4571 

Checklist:

(I owe a bunch of these but since is my first change, I'd like to sanity check that I'm on the right path before digging in more if thats ok)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ Unreleased
 
 -   Refactor ``register_error_handler`` to consolidate error checking.
     Rewrite some error messages to be more consistent. :issue:`4559`
+-   Use Blueprint decorators and functions intended for setup after
+    registering the blueprint will show a warning. In the next version,
+    this will become an error just like the application setup methods.
+    :issue:`4571`
 
 
 Version 2.1.2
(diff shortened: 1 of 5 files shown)


davidism:
Thanks for working on this. I added some more changes on top of your changes to the blueprint behavior.

* Added `@setupmethod` above all relevant methods. This ensures the error is raised closer to the decorator, when previously it might be raised by an internal call to another setup method.
* Refactored `setupmethod` to call `Scaffold._check_setup_finished`, which now does the check _and_ raises the error. This allows `Flask` and `Blueprint` to show different messages.
* Refactored tests that performed setup after requests (including some that changed config, although this isn't an error right now).

PR #4579, opened 2022-05-02
docs: adds typing to an example app

### LINKED ISSUE #4574: test typing
Add a `typing_examples.py` file that has examples of different uses of Flask with type annotations. For example, show examples of views that return each valid return type for a view. We can run mypy against this to check that *using* Flask passes typing, not just the internal type checking.

lecovi:
Examples of typing for:
- String view
- JSON view
- Error Handler
- Before Request
- Class Based View

fixes #4574

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -87,7 +87,7 @@ per-file-ignores =
     src/flask/__init__.py: F401
 
 [mypy]
-files = src/flask
+files = src/flask, tests/typing
 python_version = 3.7
 show_error_codes = True
 allow_redefinition = True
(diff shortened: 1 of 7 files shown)


davidism:
Does mypy check this file when running the typing test env?

lecovi:
This solves #4574

lecovi:
@davidism Yes, It does! 👍🏽 
Screen Shot 2022-05-02 at 16 17 33

davidism:
I moved these from the examples to tests directory, and added the directory to the mypy config so that they are tested automatically. I also did some refactoring to organize additional tests in the future.

Based on your `@app.route` tests, I realized that the decorated functions weren't actually being type checked. I've adjusted the type definitions, so now a view function that is annotated with an invalid return value will actually fail to type check.

I reported two issues to mypy while working on this:

* https://github.com/python/mypy/issues/12936
* https://github.com/python/mypy/issues/12945

PR #4580, opened 2022-05-02
Move url_for to the Flask app object

### LINKED ISSUE #4568: Move `url_for` to the `Flask` app object
Add a `url_for` method to the `Flask` app object. Similar to functions like `flask.json.dumps`, `flask.url_for` should look for a `current_app` and call its `url_for` method. This will allow applications to override the url building behavior.

ionesu:
fixes #4568

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -23,6 +23,7 @@ from werkzeug.routing import MapAdapter
 from werkzeug.routing import RequestRedirect
 from werkzeug.routing import RoutingException
 from werkzeug.routing import Rule
+from werkzeug.urls import url_quote
 from werkzeug.utils import redirect as _wz_redirect
 from werkzeug.wrappers import Response as BaseResponse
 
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for starting this, I'm going to continue it and get it ready for merge.

This was an interesting one, unlike `abort` and `redirect` the entire implementation had to be moved to the new method. That's given me the opportunity to review and clean up the implementation:

* The error message when no app context is available can be more concise and helpful about other config.
* The special build args, like `_external`, can be explicitly defined in the signature, as keyword-only, instead of popping them from the values.
* Calling `helpers.url_for` means an app context is active, which has already created a `url_adapter`. But calling `app.url_for` directly doesn't require a context, so in that case an adapter should be created.
* `adpater.build` gained a `url_scheme` argument in Werkzeug 2.0, so the hack to replace then restore it is no longer needed.

PR #4588, opened 2022-05-04
Tutorial Jinja fix

### LINKED ISSUE #4587: Tutorial Jinja template logic error
The [Jinja2 template](https://github.com/pallets/flask/blob/main/docs/tutorial/blog.rst#index) compares the user id of the currently logged in user with the `author_id` of each blog post in order to determine if the logged in user has edit permissions on that blog post. However, one of `g.user['id']` and `post['author_id']` is a string and the other is an integer. This causes the equality comparison to fail in every case, which is both unexpected behavior and confusing to users learning via the tutorial.

Environment:

- Python version: 3.9.12
- Flask version: 2.1.2

zachwick:
Fixes a comparison within one of the tutorial's Jinja2 templates.

The Jinja2 template previously compared the user id of the currently logged in user with the `author_id` of each blog post in order to determine if the logged in user had edit permissions on that blog post. However, one of `g.user['id']` and `post['author_id']` was a string and the other was an integer. This caused the equality comparison to fail in every case.

Now, a Jinja filter is used to cast each of the operands to the same type prior to the equality comparison being evaluated. 

This ensures that the "Edit" link/button now shows up when appropriate.

- fixes #4587 

This PR only touches the docs, so removing the checklist as indicated by the comment.

### DIFF
--- a/docs/tutorial/blog.rst
+++ b/docs/tutorial/blog.rst
@@ -106,7 +106,7 @@ available in the result.
               <h1>{{ post['title'] }}</h1>
               <div class="about">by {{ post['username'] }} on {{ post['created'].strftime('%Y-%m-%d') }}</div>
             </div>
-            {% if g.user['id'] == post['author_id'] %}
+            {% if g.user['id']|int() == post['author_id']|int() %}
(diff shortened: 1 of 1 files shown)


davidism:
I can't reproduce this. Both values are integers.

PR #4603, opened 2022-05-20
Fix type annotation for `after_request` callable type

### LINKED ISSUE #4600: callback types should use `bound` type vars instead of base types
* Flask 2.1.2
* Python 3.9.12
* mypy 0.950 (tested also 0.910)

For this type definition on `after_request`:

https://github.com/pallets/flask/blob/bd56d19b167822a9a23e2e9e2a07ccccc36baa8d/src/flask/typing.py#L39

It generates this mypy error:

```
error: Argument 1 to "after_request" of "Scaffold" has incompatible type "Callable[[Arg(Response, 'response')], Response]"; expected "Callable[[Response], Response]"
```

For the sample code of:

```python3
from flask import Flask, Response

def my_func(response: Response) -> Response:
    print("my after request func")
    return response

app = Flask(__name__)
app.after_request(my_func)  # <-- type error
```

This code works just fine, and did not generate typing errors for Flask 2.0.2

f3ndot:
Adds `ResponseType` typevar bound to Werkzeug's `Response` for `AfterRequestCallable`.

Flask passes in its own `Response` subclass of Werkzeug's when running the after request callback. As a result, simply having the `AfterRequestCallable` expect werkzeug's Response is will fail typing. Note: `ResponseValue` for whatever reason is still content with the original Werkzeug type being used.

- fixes #4600

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 
 -   Inline some optional imports that are only used for certain CLI
     commands. :pr:`4606`
+-   Relax type annotation for ``after_request`` functions. :issue:`4600`
 
 
 Version 2.1.2
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -26,6 +26,7 @@ from werkzeug.wrappers import Response as BaseResponse
 
 from . import cli
 from . import json
+from . import typing as ft
 from .config import Config
 from .config import ConfigAttribute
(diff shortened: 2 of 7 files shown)


f3ndot:
Tests don't seem to have typing, unsure if I should add a basic typing case for this under `tests/` which is what I did locally to ensure everything was ok

PR #4610, opened 2022-05-26
find instance_path for namespace package

### LINKED ISSUE #4602: Flask.auto_find_instance_path() can return wrong path for namespace packages installed in development mode
https://github.com/pallets/flask/blob/bd56d19b167822a9a23e2e9e2a07ccccc36baa8d/src/flask/scaffold.py#L798

If there are several packages under the same namespace, all installed in development mode, like:

```
~/namespace-package1/
    namespace/
        package1/
            __init__.py
            app.py
    instance/

~/namespace-package2/
    namespace/
        package2/
            __init__.py
            app.py
    instance/
```
and the code in `namespace.package2` uses `app.instance_path`, then its expected value is `~/namespace-package2/instance` (["Uninstalled package" decision path](https://flask.palletsprojects.com/en/2.1.x/config/#instance-folders)).

Instead of that the following happens:
* `find_package()` [cuts import info](https://github.com/pallets/flask/blob/bd56d19b167822a9a23e2e9e2a07ccccc36baa8d/src/flask/scaffold.py#L846) to the very top package name, `namespace`,
* then `_find_package_path()` finds module specification for the whole namespace package, which contains several submodule search locations, like `ModuleSpec(name='namespace', loader=<_frozen_importlib_external._NamespaceLoader object at ...>, submodule_search_locations=_NamespacePath(['~/namespace-package1/namespace', '~/namespace-package2/namespace']))`
* and then the quoted line returns first, i.e. _arbitrary_, package from that namespace, e.g. `~/namespace-package1`, which produces wrong instance path.

Suggestion: pass also `import_name` into `_find_package_path` and use it for resolving ambiguity at this point, like:

```
def _find_package_path(root_mod_name, import_name):
...
            if spec.origin in {"namespace", None}:
                package_spec = importlib.util.find_spec(import_name)
                package_path = os.path.commonpath(package_spec.submodule_search_locations)
                return os.path.dirname(next(
                    location for location in spec.submodule_search_locations
                    if package_path.startswith(location)
                ))
```

eprigorodov:
Fix `Flask.auto_find_instance_path()` so that it handles correctly namespace packages residing in several different locations, e.g. from several projects installed in development mode.

Return not the first submodule search location but the one most close to Flask app import name.

- fixes #4602 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Unreleased
 -   Inline some optional imports that are only used for certain CLI
     commands. :pr:`4606`
 -   Relax type annotation for ``after_request`` functions. :issue:`4600`
+-   ``instance_path`` for namespace packages uses the path closest to
+    the imported submodule. :issue:`4600`
 
 
 Version 2.1.2
--- a/src/flask/scaffold.py
+++ b/src/flask/scaffold.py
@@ -780,30 +780,46 @@ def _matching_loader_thinks_module_is_package(loader, mod_name):
     )
 
 
-def _find_package_path(root_mod_name):
+def _find_package_path(import_name):
     """Find the path that contains the package or module."""
-    try:
-        spec = importlib.util.find_spec(root_mod_name)
+    root_mod_name, _, _ = import_name.partition(".")
 
-        if spec is None:
+    try:
+        root_spec = importlib.util.find_spec(root_mod_name)
+
+        if root_spec is None:
             raise ValueError("not found")
     # ImportError: the machinery told us it does not exist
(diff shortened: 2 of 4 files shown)


davidism:
Well, this revealed that every `instance_path` test fails under tox. The tox tmp directory is under the virtualenv path, which means all paths look installed. I'll try to fix the tests, otherwise the code looks good and I'll just mark it xfail with the rest.

eprigorodov:
thank you for the help, @davidism, broken tox tests would have been the last thing I'd suspect

PR #4620, opened 2022-06-04
Remove extra backtick in docs

Numerlor:
Removes an extra backtick from the inline code


davidism:
Please make docs changes against the latest feature release.

PR #4629, opened 2022-06-09
allow view to return generator

pgjones:
### Directly accept generators as returned from view functions

My expectation was that this was already possible to stream responses
and I can't think of a reason why this shouldn't be allowed. Quart
already does this without any issues I know of.

### Add generate_template and generate_template_string functions

These helper functions make it easier to render a template piece by
piece with the correct context. Much like render_template and
render_template_string make it easy to render the entire template.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,8 @@ Unreleased
     value. :pr:`4645`
 -   View functions can return generators directly instead of wrapping
     them in a ``Response``. :pr:`4629`
+-   Add ``stream_template`` and ``stream_template_string`` functions to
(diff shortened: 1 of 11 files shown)


davidism:
* Instead of using `yield from` directly, use a closure so that we can wrap it with `stream_with_context` automatically if a request is active. This still returns a generator, but the function itself is not a generator.
* Use `isinstance(value, collections.abc.Iterator)` instead of `inspect.isgenerator(value)`. This allows returning `iter(["a", "b", "c"])`, as well as generators and generator expressions.
* Update the types and type tests to allow `Iterator`, which also allows `Generator`.

PR #4637, opened 2022-06-12
rewrite deployment docs

davidism:
Completely rewrite the deployment docs. Documents the following:

* Gunicorn, with sync, gevent, or eventlet workers
* Waitress
* mod_wsgi-express
* uwsgi, without special nginx/apache uwsgi support, with sync or gevent workers
* gevent
* eventlet
* WSGI server pages all use a standard format. Show how to install in a virtualenv and run with an app or factory. Shows how to listen on external addresses, discusses not running as root and using a reverse proxy instead (except for mod_wsgi). Servers that support async workers show how and why to use them.
* Using `ProxyFix` when behind a reverse proxy or hosting platform
* nginx
* Apache httpd
* HTTP server pages give a brief overview of acquiring a domain name or editing the hosts file.
* Keeps links to the most common hosting platforms, links to `ProxyFix` docs as well.

I will copy this to Werkzeug's docs as well.


davidism:
Removed the recommendation to install `mod_wsgi-standalone`, as it turns out it's not recommended for normal use. a9878c018bdecc377788204a3bc93497ec1f050a @GrahamDumpleton

davidism:
Docs can be read here: https://flask.palletsprojects.com/en/2.1.x/deploying/

PR #4645, opened 2022-06-17
Use timezone aware datetime object on session

kkirsche:
Naïve datetime objects are discouraged in https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

> Warning Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling datetime.now(timezone.utc).

- fixes N/A

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,8 @@ Unreleased
 -   It is no longer required to decorate custom CLI commands on
     ``app.cli`` or ``blueprint.cli`` with ``@with_appcontext``, an app
     context will already be active at that point. :issue:`2410`
+-   ``SessionInterface.get_expiration_time`` uses a timezone-aware
+    value. :pr:`4645`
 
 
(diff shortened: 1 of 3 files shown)


kkirsche:
Sorry for not including all the contributing items. Some of the entries if you are open to this change I will do once I update my personal access token (can't work with this repo on the CLI right now due to not having the workflow permission authorized from the CLI)

davidism:
I updated everything else to be timezone aware, this makes sense too. This case is already treated as UTC in `set_cookie`, but might as well be consistent.

kkirsche:
1. Changes entry added
2. I don't believe that there is a relevant documentation update, but please let me know if that's incorrect
3. Existing tests related to verifying the session's expiration cover this, I believe


Please let me know if there is anything else you need and I'll do my best to provide those updates

PR #4648, opened 2022-06-18
Modified bluprints.py and debughelpers.py to adhere more to PEP8 stan…

cattells:
…dards

This did not seem to fall into any criteria for creating an issue ticket

Checklist:

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -12,6 +12,7 @@ from .scaffold import setupmethod
 if t.TYPE_CHECKING:  # pragma: no cover
     from .app import Flask
 
+
 DeferredSetupFunction = t.Callable[["BlueprintSetupState"], t.Callable]
 
 
@@ -68,6 +69,7 @@ class BlueprintSetupState:
         self.url_defaults = dict(self.blueprint.url_values_defaults)
         self.url_defaults.update(self.options.get("url_defaults", ()))
 
+
     def add_url_rule(
         self,
         rule: str,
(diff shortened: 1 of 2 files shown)


cattells:
There are some inconsistencies regarding blank lines/whitespace. For example, according to the PEP-257 Docstring Conventions, there should not be a blank line before or after any docstrings unless the docstring is for a class (not a function/method). This is only a recommendation per PEP-257, so it does not have to be changed, however there are many functions/methods without a blank line after their respective docstrings, so inconsistencies do exist, and the file may appear to be more congruent if the spacing was uniform throughout.

Some docstrings are a bit wordy. I think that some of the details provided in the docstrings (such as version changes) may be relocated to some documentation file or other form of version tracking, like on Github. This is a suggestion however I did not change this because it did not seem appropriate for me to do.

Some methods are missing docstrings entirely. I marked these with # TODO, as I did not change this because it did not seem appropriate for me to do.

@setupmethod decorator is never described. If one is not familiar with this particular import, then we do not know its purpose. It is referenced many times, so it would be helpful to at least include some comments.

Debughelpers.py
These problems occur throughout. The next file that I found with complications was debughelpers.py. In this file, we also encounter the same inconsistencies regarding blank lines following docstrings. Once again, not all methods have docstrings. Additionally, there are inconsistencies with whitespace throughout. I think that whitespace would greatly increase readability.

davidism:
Thanks for your interest. However, these changes are not needed, and we already use tools to enforce code style.

PR #4649, opened 2022-06-22
wrap ContextVar more directly instead of Werkzeug LocalStack

davidism:
Werkzeug's `Local` and `LocalStack` switched to using `ContextVar`, but it's a fairly complex implementation, and it's not clear that creating a `ContextVar` in `Local.__init__` is correct.

Instead, a `ContextVar` is defined globally, and passed to a new `CtxStack` implementation that has the same `push`, `pop`, `top` operations of the existing stack. This removes a lot of indirection from the Werkzeug implementation.

### DIFF
--- a/src/flask/globals.py
+++ b/src/flask/globals.py
@@ -1,12 +1,13 @@
 import typing as t
-from functools import partial
+from contextvars import ContextVar
 
 from werkzeug.local import LocalProxy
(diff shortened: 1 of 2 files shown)


davidism:
This was just a proof of concept, I'm going to try to refactor `werkzeug.local` based on this, then I'll make a separate PR using that.

PR #4654, opened 2022-06-26
Update quickstart to reflect get() and post() decorator syntax

hankhank10:
Added a short paragraph and example to the http-methods section to explain that http methods can be referenced directly by using the `get()` and `post()` decorators as well as the `route(methods=['GET', 'POST])` decorator.

I believe this will be helpful for those new to Flask as it is simpler syntax and doesn't require referencing the `request` object which hasn't yet been introduced by this point in the quickstart.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -390,6 +390,24 @@ of the :meth:`~flask.Flask.route` decorator to handle different HTTP methods.
         else:
             return show_the_login_form()
 
+The example above keeps all methods for the route within one function,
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Separate functions are not a good idea for typical (wt) form use cases...

hankhank10:
I thought it would be helpful to follow the same format as the existing example for clarity, but happy to rework it to a better use case if you think better?

Equally feel free to reject if you think not helpful in any case.

PR #4657, opened 2022-06-28
(GitHub workflows CI) Bump tests CI versions

ArjunSharda:
This bumps the versions for the tests CI workflow up to a higher version as two of the versions currently used in the file are outdated.


davidism:
Thanks, however we use a different process to manage these updates.

Issue #4659, opened 2022-06-29
Flask return a list directly


vba34520:
```python
from flask import Flask

app = Flask(__name__)


@app.route('/')
def index():
    return [1, 2, 3]


if __name__ == '__main__':
    app.run()
```

It will raise **TypeError: The view function did not return a valid response. The return type must be a string, dict, tuple, Response instance, or WSGI callable, but it was a list.**

Then I have to do like this.

```python
from flask import Flask, jsonify

app = Flask(__name__)


@app.route('/')
def index():
    return jsonify([1, 2, 3])


if __name__ == '__main__':
    app.run()
```

And I found that a dict can be jsonify automatically in [app.py#L1924](https://github.com/pallets/flask/blob/main/src/flask/app.py#L1924)

So why does flask can't jsonify a list like a dict ?

davidism:
Because a sequence of values already has another meaning.

vba34520:
How to auto jsonify a list before make_response()?

DrizztDU:
> How to auto jsonify a list before make_response()?

`@app.route('/')` is a decorator which adds an endpoint to the `app` object. So you have two ways to do it.

1. implement a function decorator.
2. create a class `MyFlask` that inherites from class `Flask`, and override class method `route`.

pgjones:
I would actually be better to override the `make_response` method on the `Flask` class as the route just registers the function and rule.

As David has said though, returning `["Hello", 201]` means return `"Hello"` and the status code `201`. So you'll need to somehow distinguish between this and your JSON list response.

davidism:
So @greyli ended up implementing this in #4672. I'm a bit worried that you can return so many different things that looks somewhat similar but do different things now, but I guess everything is distinct enough still and seems to align with expectations.

- tuple is not a list, it will not be serialized to json because it provides headers and status instead
- iter[list[str | bytes]] is a streaming response, not json
- list is converted to json

PR #4672, opened 2022-07-03
Allow to return list as JSON

greyli:
This adds support to serializing list return values to JSON, like what we did for the dictionary:

```python
@app.route('/')
def index():
    return ['this', 'will', 'become', 'JSON']
```

There are some discusses related to this feature in #3111, and the concern is the list return value looks too much like the tuple. Maybe we can rethink about this feature? I propose this because the following reasons:

- List as JSON response is a common use case for Web APIs. We support returning dict as JSON, but not list. This makes people feel inconsistent (#4533, #4659).
- With #2737, we only treat tuple as a valid shortcut for return values, so there will be no program conflicts to allow returning list (we also throw an error for list return value currently).
- In terms of semantics, there is no concept of "tuple" in JSON. Although list and tuple are both sequences in Python, there are differences in syntax and meaning. We can add docs to address the different meaning of list and tuple in return values.
- In terms of visuals, most people will not mix `[1, 2, 3]` and `(1, 2, 3)`, and people usually return tuple items without the bracket in the view function.

If this change is acceptable, I will update the docs and changelog.

Sorry for break the "zero status". :P

Checklist:

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -681,22 +681,25 @@ The return value from a view function is automatically converted into
 a response object for you. If the return value is a string it's
 converted into a response object with the string as response body, a
 ``200 OK`` status code and a :mimetype:`text/html` mimetype. If the
-return value is a dict, :func:`jsonify` is called to produce a response.
-The logic that Flask applies to converting return values into response
-objects is as follows:
+return value is a dict or list, :func:`jsonify` is called to produce a
+response. The logic that Flask applies to converting return values into
+response objects is as follows:
 
 1.  If a response object of the correct type is returned it's directly
     returned from the view.
 2.  If it's a string, a response object is created with that data and
     the default parameters.
-3.  If it's a dict, a response object is created using ``jsonify``.
-4.  If a tuple is returned the items in the tuple can provide extra
+3.  If it's an iterator or generator returning strings or bytes, it is
+    treated as a streaming response.
+4.  If it's a dict or list, a response object is created using
+    :func:`~flask.json.jsonify`.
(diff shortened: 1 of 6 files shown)


greyli:
The failing "Pallets Development Versions" check is due to a missing import in Werkzueg, I will make a PR to fix it later.

davidism:
Looks like that fix should be in Werkzeug

davidism:
I think my original hesitation was that returning a dict is explicitly different from returning a database model. Returning a list of database results is probably something a lot more users will try. But I guess it's not that different to say they have to do `obj.to_json()` and also support `[obj.to_json() for obj in objs]`.

greyli:
I see. If they don't return a list of objects directly and call jsonify for the list, they will get the same error. Maybe we could update the docs to mention that the ORM/ODM object is not jsonify-able, they need to convert them to a dict or use a serialization library.

davidism:
Another problem now that we support returning iterators/generators is that you're going to get very different results doing `return ["a", "b", "c"]` or `return iter(["a", "b", "c"])`.

docs/quickstart.rst
```diff
@@ -741,20 +744,20 @@ view, it will be converted to a JSON response.
...
-
-.. code-block:: python
-
-    from flask import jsonify
-
     @app.route("/users")
     def users_api():
         users = get_all_users()
-        return jsonify([user.to_json() for user in users])
+        return [user.to_json() for user in users]
```
greyli, line 750: Haha, I just made some very similar changes to this section locally.

PR #4676, opened 2022-07-06
Typing

pgjones:
See also https://github.com/pallets/quart/commit/7f9e9e25d1d2d6823673b286ad981d6089a26e3b

Checklist:


davidism:
Can you add some tests to `test/typing`? See the routes for an example, it doesn't have to represent a real application, just lots of stuff that should each pass these type checks.

PR #4687, opened 2022-07-11
Fix CHANGES.rst typo

Jalkhov:
I think the first `due` was too much, other than that, I would change it:

> ``tojson`` used in templates...

By

> The ``tojson`` used in templates...

but I didn't put it in the commit with the intention of not making unnecessary modifications.


davidism:
Please combine all further fixes to this file into a single PR after you're finished reviewing it completely. No need to justifiy each one individually.

PR #4692, opened 2022-07-13
add JSON provider interface

davidism:
This adds the ability to fully customize the JSON implementation used by a Flask application. Using a different JSON implementation can greatly speed up API applications that need to work with JSON in most requests.

`app.json` is an instance of `Flask.json_provider_class`. `flask.json.provider.JSONProvider` is the base class that defines `dumps`, `dump`, `loads`, `load`, and `response` methods, of which only `dumps` and `loads` need to be implemented.  For example, here's a provider for [orjson](https://github.com/ijl/orjson):

```python
from flask.json.provider import JSONProvider
import orjson

class OrJSONProvider(JSONProvider):
    def dumps(self, obj, *, option=None, **kwargs):
        if option is None:
            option = orjson.OPT_APPEND_NEWLINE | orjson.OPT_NAIVE_UTC
        
        return orjson.dumps(obj, option=option)

    def loads(self, s, **kwargs):
        return orjson.loads(s)

# assign to an app instance
app.json = OrJSONProvider(app)

# or assign in a subclass
class MyFlask(Flask):
    json_provider_class = OrJSONProvider

app = MyFlask(__name__)
```

The methods in `flask.json` call the methods on `app.json` if an app context is active, or fall back to the `json` library. `jsonify` calls `app.json.response`. The `|tojson` filter uses `app.json.dumps`. `Request.json` uses `app.json.loads` and `Response.json` uses `app.json.dumps`; the test client uses these as well.

Customizing `json_encoder` or `json_decoder` on an app or blueprint, and the `JSONEncoder` and `JSONDecoder` classes, are deprecated. This was not an effective way to use other libraries. Customizing per blueprint was requested by an API extension that is no longer maintained and didn't appear to use the feature. It's not clear how it would work with the new provider interface and added overhead to every request. Instead, API frameworks should be using a dedicated object serialization library, then taking advantage of a fast JSON serializer at the application level.

The `DefaultJSONProvider` is the existing implementation using the built-in `json` library. The `app.config` keys `JSON_AS_ASCII`, `JSON_SORT_KEYS`, `JSONIFY_MIMETYPE`, and `JSONIFY_PRETTYPRINT_REGULAR` are deprecated and have moved to attributes on the default provider. Other providers are not required to support these options.

### DIFF
--- a/src/flask/json/provider.py
+++ b/src/flask/json/provider.py
@@ -1,0 +1,310 @@
+from __future__ import annotations
+
+import dataclasses
+import decimal
+import json
+import typing as t
+import uuid
+import weakref
+from datetime import date
+
+from werkzeug.http import http_date
+
+from ..globals import request
+
+if t.TYPE_CHECKING:  # pragma: no cover
+    from ..app import Flask
+    from ..wrappers import Response
+
+
+class JSONProvider:
+    """A standard set of JSON operations for an application. Subclasses
+    of this can be used to customize JSON behavior or use different
+    JSON libraries.
+
+    To implement a provider for a specific library, subclass this base
+    class and implement at least :meth:`dumps` and :meth:`loads`. All
+    other methods have default implementations.
+
+    To use a different provider, either subclass ``Flask`` and set
+    :attr:`~flask.Flask.json_provider_class` to a provider class, or set
+    :attr:`app.json <flask.Flask.json>` to an instance of the class.
+
+    :param app: An application instance. This will be stored as a
+        :class:`weakref.proxy` on the :attr:`_app` attribute.
+
+    .. versionadded:: 2.2
+    """
+
+    def __init__(self, app: Flask) -> None:
+        self._app = weakref.proxy(app)
+
+    def dumps(self, obj: t.Any, **kwargs: t.Any) -> str:
+        """Serialize data as JSON.
+
+        :param obj: The data to serialize.
+        :param kwargs: May be passed to the underlying JSON library.
+        """
+        raise NotImplementedError
+
+    def dump(self, obj: t.Any, fp: t.IO[str], **kwargs: t.Any) -> None:
+        """Serialize data as JSON and write to a file.
+
+        :param obj: The data to serialize.
+        :param fp: A file opened for writing text. Should use the UTF-8
+            encoding to be valid JSON.
+        :param kwargs: May be passed to the underlying JSON library.
+        """
+        fp.write(self.dumps(obj, **kwargs))
+
+    def loads(self, s: str | bytes, **kwargs: t.Any) -> t.Any:
+        """Deserialize data as JSON.
+
+        :param s: Text or UTF-8 bytes.
+        :param kwargs: May be passed to the underlying JSON library.
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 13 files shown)


src/flask/json/provider.py
```diff
@@ -0,0 +1,310 @@
...
+
+        Either positional or keyword arguments can be given, not both.
+        If no arguments are given, ``None`` is serialized.
+
+        :param args: A single value to serialize, or multiple values to
+            treat as a list to serialize.
+        :param kwargs: Treat as a dict to serialize.
+        """
+        obj = self._prepare_response_obj(args, kwargs)
+        return self._app.response_class(self.dumps(obj), mimetype="application/json")
```
pgjones, line 106: Worth a cls default_mimetype here to save extensions having to override this method and as matching with responses?

davidism: Discussed this more here: https://github.com/pallets/flask/pull/1728#issuecomment-1181806155

I'm not sure this should actually be a configuration option. The original issue seemed to be about a specific type of response, not all JSON responses.

I did originally have this as a `JSONProvider.mimetype` attribute, but I ended up moving all existing behavior to `DefaultProvider` and making the base very simple.

Yourun-proger: @davidism, maybe add your orjson example to the documentation?

davidism: I don't want to include it at this time.

Yourun-proger: ok, just a small suggestion

src/flask/json/provider.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+
+    default: t.Callable[[t.Any], t.Any] = staticmethod(
+        _default
+    )  # type: ignore[assignment]
+    """Apply this function to any object that :meth:`json.dumps` does
+    not know how to serialize. It should return a valid JSON type or
+    raise a ``TypeError``.
+    """
+
```
pgjones, line 146: What do you think about adding a dict_to_object hook here as well (for the loads side). Allows something like this,
```python
class MoneyJSONProvider(DefaultJSONProvider):

        @staticmethod
        def default(object_):
            if isinstance(object_, date):
                return http_date(object_)
            if isinstance(object_, (Decimal, UUID)):
                return str(object_)
            if is_dataclass(object_):
                return asdict(object_)
            if hasattr(object_, "__html__"):
                return str(object_.__html__())
            if isinstance(object_, Money):
                return {'amount': object_.amount, 'currency': object_.currency}

            raise TypeError(f"Object of type {type(object_).__name__} is not JSON serializable")

        @staticmethod
        def dict_to_object(dict_):
            if 'amount' in dict_ and 'currency' in dict_:
                return Money(Decimal(dict_['amount']), dict_['currency'])
            else:
                return dict_ 
```

davidism: I left it out for a few reasons. `object_hook` is not as consistently supported by different libraries as `default` is, and I didn't want to put a perceived requirement for it on all other providers. You usually want to perform validation when deserializing, and that gets very messy trying to cram it all in `object_hook` along with proper error collection. Instead, any project should use a serialization library, leaving the provider to only handle the JSON and basic types.

PR #4693, opened 2022-07-13
Add error message

Yourun-proger:
When trying to use the `render_template` and `render_template_string` functions outside of the context of an application, a silly None error occurs that does not tell the user that this function only works when there is a flask application. I just fixed it.
I decided not to create a problem report about this, because this seems to be a fairly simple and small matter. Plus it's just like #4333 , which I've already fixed. Sorry, if this is still necessary, I will immediately create this message so that everything goes "by the rules". 

Checklist:


davidism:
This is no longer needed, it's completely refactored in 2.2.

PR #4695, opened 2022-07-14
Allow TypedDict as a ResponseValue

zr40:
When using a `TypedDict` as the return type for a route, this is rejected by mypy with the following error:

```
class ExampleDict(TypedDict):
    status: str

@blueprint.route("/example")
def example() -> ExampleDict:
    return {"status": "ok"}

# error: Value of type variable "ft.RouteDecorator" of function cannot be "Callable[[str], ExampleDict]"  [type-var]
```

The problem first appeared in 2.1.3 due to #4579, which changed the route decorator's callable to `ViewCallable`, which requires the return type to be `ResponseValue`. `ResponseValue` is a union that already includes `dict[str, Any]`, however it does not include `TypedDict`, which would behave the same as `dict[str, Any]` at runtime.

This PR adds `TypedDict` to the `ResponseValue` union.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,8 @@ Unreleased
 
 -   Allow returning a list from a view function, to convert it to a
     JSON response like a dict is. :issue:`4672`
+-   When type checking, allow ``TypedDict`` to be returned from view
+    functions. :pr:`4695`
 
 
 Version 2.1.3
--- a/src/flask/typing.py
+++ b/src/flask/typing.py
@@ -11,7 +11,8 @@ ResponseValue = t.Union[
     str,
     bytes,
     t.List[t.Any],
-    t.Dict[str, t.Any],
+    # Only dict is actually accepted, but Mapping allows for TypedDict.
+    t.Mapping[str, t.Any],
     t.Iterator[str],
     t.Iterator[bytes],
 ]
--- a/tests/typing/typing_route.py
+++ b/tests/typing/typing_route.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import typing as t
 from http import HTTPStatus
 
+import typing_extensions as te
+
 from flask import Flask
 from flask import jsonify
 from flask import stream_template
@@ -36,6 +38,15 @@ def hello_json_dict() -> t.Dict[str, t.Any]:
 @app.route("/json/dict")
 def hello_json_list() -> t.List[t.Any]:
     return [{"message": "Hello"}, {"message": "World"}]
+
+
+class StatusJSON(te.TypedDict):
+    status: str
+
+
+@app.route("/typed-dict")
+def typed_dict() -> StatusJSON:
+    return {"status": "ok"}
 
 
 @app.route("/generator")


zr40:
The python 3.7 tests fail because that version doesn't have `typing.TypedDict`. Type checking would still work correctly on all supported versions if instead `typing_extensions.TypedDict` is used, however I can't get that to work in CI. Any suggestions? :)

davidism:
We use `typing_extensions` in other places. What does "can't get it to work" mean?

davidism:
`TypedDict`, either from typing_extensions or with MyPy set to a higher Python version, does not appear to be a valid type on its own. So there is no way to merge this right now as it introduces errors (or ambiguity if ignored). It seems likely that https://github.com/python/mypy/issues/12934 or another related issue is what needs to be fixed.

davidism:
No, we specifically accept dict, not anything that satisfies the mapping protocol.

See my subsequent comments. This is an issue with MyPy or TypedDict, there is nothing that can be done in Flask. Either it's an issue because `TypedDict` isn't valid where `dict` is, or it's an issue because `TypedDict` on its own is not a valid type in unions.

davidism:
I'm not sure what CI you're looking at. The CI for this PR has failures everywhere. When I fix the TypedDict import, it causes more failures.

```
src/flask/typing.py:16: error: Variable "typing_extensions.TypedDict" is not valid as a type  [valid-type]
src/flask/typing.py:16: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

zr40:
I've added the version check in 261c00288aaaaee0b6340698d120377bc78f0e62, but since the PR is closed CI doesn't run. However I believe all tests should pass, since they did before for 3.8+ and the version check fixes the 3.7 case.

davidism:
Report the issue to MyPy (or lend your feedback to the issue I already linked). It's an issue with MyPy or typing in general, not with Flask.

zr40:
It's not a issue from MyPy's point of view, as I've detailed above in https://github.com/pallets/flask/pull/4695#issuecomment-1184473249.

Being able to pass a TypedDict to an interface that accepts dict, means that this interface can modify the dict without regard to the type invariants afforded by TypedDict, so any remaining references to that TypedDict cannot rely on its type invariants still being held. Therefore, the MyPy error is correct and this is not an issue with the MyPy project.

Also, the MyPy project will tell you that Mapping is the appropriate type to accept, since Flask does not modify the dict and TypedDicts do satisfy the Mapping protocol.

As things stand now, I must introduce a version bound for Flask that rejects any version newer than 2.1.2. Our deployment validation workflow is broken by the release of 2.1.3 containing this regression.

zr40:
Do you suppose that type checking is the right place to enforce this, at the cost of solving real-world problems? You can already return any non-dict object if you simply don't type check and it will still break visibly.

Right now, Flask's typing.py rejects valid real-world usage. Adding `Mapping` solves this real-world problem and strictly makes the situation better than it was in 2.1.2.

davidism:
> Do you suppose that type checking is the right place to enforce this

Yes, given that the entire point of type checking is to describe the correct types and catch things before runtime.

> Right now, Flask's typing.py rejects valid real-world usage.

But with `Mapping`, it would accept invalid uses, so that doesn't seem "strictly better" just "different for that specific use case".

zr40:
> Yes, given that the entire point of type checking is to describe the correct types and catch things before runtime. If they're not able to do that, that's a design issue that needs to be fixed with typing or type checkers.

In that case, I would suggest to revert #4579.

> But with Mapping, it would accept invalid uses, so that doesn't seem "strictly better" just "different for that specific use case".

But in 2.1.2, literally every type was accepted, including every invalid type. Yes, being able to reject all invalid types before runtime is ideal, but not at the cost of also rejecting valid types.

With `Mapping`, all invalid non-Mapping types are still rejected, and non-dict Mapping types cause a very clearly written runtime error: `The view function did not return a valid response. The return type must be a string, dict, list, tuple with headers or status, Response instance, or WSGI callable, but it was a object.` which cannot be misconstrued.

davidism:
I'm not willing to revert that commit or use `Mapping`. I really don't appreciate being dragged into a huge discussion on the behavior of type checking and whether your specific use case trumps other use cases. I don't like being told that it's Flask's fault something doesn't work instead of typing's fault, it's tiring and not particularly convincing. Take it up with typing first if that's the path of discussion you want to go on.

davidism:
I'm not dismissing your efforts or your use case, I'm saying I don't like the way you continue to frame the discussion. It feels argumentative, not productive, and you don't appear to be willing to consider the fact that it should be fixed in typing.

davidism:
I've pushed a change to use `typing_extensions.TypedDict`. I've also temporarily bumped MyPy's Python version to 3.8. Notice that the error "TypedDict is not valid as a type" is shown whether you use it from `typing` or `typing_extensions`.

The fact that it fails with MyPy in 3.8 mode suggests that the `sys.version` check is just masking the type completely, hiding the new issue. So that doesn't seem like a valid solution.

I'm not willing to use `Mapping` at this time. If this cannot be solved without using `Mapping`, that needs to be reported to typing, as it's forcing a less precise type when that shouldn't be required.

zr40:
Well, I definitely agree there's a contravariance problem when sending a TypedDict to a dict receiver, but I'm not sure what the typing or MyPy projects could theoretically do to resolve this.

Suppose they could define a `WeakTypedDict` that's still a dict at runtime and that can validly be sent to a dict receiver. But then type checkers would lose the ability to reason correctly about the contents of that dict, which isn't a desirable outcome.

Still, I shall open a issue there and link it here.

davidism:
For Flask's purpose though, it doesn't matter what the contents are. You want to check the return value of your own funtion, where it does matter. But `route` and `make_response` only care that it's a dict, and specifically a dict and not another mapping.

zr40:
https://github.com/python/mypy/issues/13122

davidism:
Thanks. I suppose mapping is ok since there's still an error at runtime, although hopefully typing can address it eventually. I'll clean this up in a bit.

PR #4698, opened 2022-07-16
Support calling with_appcontext decorator with parenthesis

filips123:
I've added type hints for the `with_appcontext` decorator. This fixes mypy error "Untyped decorator makes function untyped" for otherwise-typed function that uses this decorator. Although explicit `with_appcontext` is probably not needed in most cases as of #4647, I think it would still make sense if it's typed.

I also included a small change to allow calling `with_appcontext` with parenthesis, because the same [is possible](https://github.com/pallets/quart/blob/main/src/quart/cli.py#L271-L273) in Quart.

If possible, it would be nice if added type hints could be backported to 2.1, so I don't have to wait for 2.2 to fix mypy issues.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -95,6 +95,8 @@ Unreleased
     functions. :pr:`4695`
 -   It is now possible to also call ``@with_appcontext`` decorator
     with parenthesis, to have additional parity with Quart.
+-   The ``@with_appcontext`` decorator decorator has type hints.
+
 
 Version 2.1.3
 -------------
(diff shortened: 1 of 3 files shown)


pgjones:
I'm not sure if calling the decorator with parenthesis adds anything useful, I think we should go the other way and remove the ability to do so from Quart.

filips123:
That's true, I just noticed Quart supports decorator with parenthesis and decided to also add it here. I can remove that commit if you want.

davidism:
This decorator is essentially unnecessary in Flask 2.2. I don't plan to change it more at this time.

filips123:
Fair enough. Do you have any approximation when 2.2 will be ready?

PR #4712, opened 2022-07-28
fix `MAX_CONTENT_LENGTH` calc to match description

yedpodtrzitko:
- description below says the value is in megabytes, so it should be `* 1024` instead of `* 1000`?

Checklist:

### DIFF
--- a/docs/patterns/fileuploads.rst
+++ b/docs/patterns/fileuploads.rst
@@ -141,7 +141,7 @@ config key::
     from flask import Flask, Request
 
     app = Flask(__name__)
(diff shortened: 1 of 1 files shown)


davidism:
This change isn't needed. The 1000 vs 1024 issue is not worth it.

yedpodtrzitko:
The next line in docs says "The code above will limit the maximum allowed payload to 16 megabytes.", which is not this number. Hence this PR to fix the discrepancy.

davidism:
You're thinking of mebi (1024) vs mega (1000), it's correct as-is. But personally I don't care about the distinction.

PR #4718, opened 2022-08-01
remove `--eager-loader/--lazy-loader` options

### LINKED ISSUE #4715: remove ability to toggle lazy/eager loading
When the CLI was first introduced, it would always lazily load the application, and then resolve that lazy import the first time the app was needed. For the `run` command, in debug mode, that would be _after_ the server started, so that errors would show up in the debugger and the reloader would work across errors. This meant that errors wouldn't show up immediately when running the command, which was confusing. The `--eager-loading/--lazy-loading` option controlled whether that was disabled/enabled regardless of debug mode.

Later, this behavior was changed so that the app is always eagerly loaded the first time, and only lazily loaded on reloads. This makes errors show up consistently when running any command, including `run`, while still allowing the reloader to work across errors.

There shouldn't be a reason now to control loading. Errors will always be shown immediately in the terminal when a command is run. Lazy loading should always be used within the reloader to handle errors.

davidism:
closes #4715

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -92,6 +92,12 @@ Unreleased
     JSON response like a dict is. :issue:`4672`
 -   When type checking, allow ``TypedDict`` to be returned from view
     functions. :pr:`4695`
+-   Remove the ``--eager-loading/--lazy-loading`` options from the
+    ``flask run`` command. The app is always eager loaded the first
+    time, then lazily loaded in the reloader. The reloader always prints
+    errors immediately but continues serving. Remove the internal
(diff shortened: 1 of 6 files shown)


davidism:
I realized that the previous lazy implementation using the `DispatchingApp` middleware was unnecessarily complicated. Instead, the `run` command can always `try/except` loading the app, and then either re-raise the error immediately or create an `app` callable that will raise it later. There doesn't appear to be a reason to use threads and locks to do this.

When using the reloader, tracebacks can be printed immediately while also deferring raising them. This makes the error immediately visible in the terminal without needing to access the app again.

Issue #4740, opened 2022-08-04
Wrong default value for FLASK_ENV


studer-l:
Previously, if `FLASK_ENV` was not set, it would default to `"production"`, now it defaults to `"development"`.

Reproduction:

```python
from flask import Flask

app = Flask(__name__)

@app.route("/")
def hello_world():
    print(f"{app.config.get('ENV')=}")
    return "<p>Hello, World!</p>"
```

Run using gunicorn, `gunicorn flask_demo.app:app` with `FLASK_ENV` unset,

Flask 2.2.1, this handler prints:
`app.config.get('ENV')='development'`

downgrade to 2.1.0:
`app.config.get('ENV')='production'`

I assume this due to  #4720, `src/flask/app.py:693`

Environment:

- Python version: 3.10.4
- Flask version: 2.2.1

davidism:
Note that this only affects the value of `app.env`, not `FLASK_ENV`. Both are deprecated. `app.env` has no effect on Flask itself anymore. Only `FLASK_ENV=development` will switch to debug mode.

PR #4741, opened 2022-08-04
Fixes wrong default value for FLASK_ENV

studer-l:
Fixes issue #4740 by setting the previous default value.


davidism:
Closing in favor of #4742, which is made against the 2.2.x maintenance branch, includes a changelog, and references `app.env` not `FLASK_ENV`.

PR #4752, opened 2022-08-07
#378 explictly set charset for jsonify()

clach04:
https://webhint.io/docs/user-guide/hints/hint-content-type/ reference for best practice

Without this Microsoft Edge devtools post a warning

https://github.com/pallets/flask/issues/378 appears to be the first report (but was also possibly caused by bad payload).

- fixes #378 

Checklist:

Opening as a place holder. I don't have time to setup a dev environment and create a new test for this.


davidism:
If you need this, you can set `app.json.mimetype`.

PR #4759, opened 2022-08-11
Removed checks of useless keys by the property of sorted elements

izveigor:
The function "from_prefixed_env" finds all words with the concrete prefix (by default "FLASK_"). The description of the function has the important condition: all keys are loaded in **sorted** order. Sorted order guarantees that all keys with needed prefix are consecutive (trivial statement). So, the function can use this property to reduce the number of operations.

Checklist:

### DIFF
--- a/src/flask/config.py
+++ b/src/flask/config.py
@@ -127,10 +127,16 @@ class Config(dict):
         prefix = f"{prefix}_"
         len_prefix = len(prefix)
 
+        variables = list()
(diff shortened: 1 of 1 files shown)


pgjones:
I don't think this makes sense. As I understand it, only the first prefixed environment variable will be loaded.

It also isn't sure what you are hoping to achieve - maybe add a test to demonstrate the need?

PR #4765, opened 2022-08-15
fix model parameter is missing

sattamjh:
missing argument in example code


davidism:
Remember to make docs changes on a separate branch, and against the 2.2.x fix branch. Also, there were other code issues in these docs. Addressed in d94634b1

PR #4767, opened 2022-08-15
Changes

lebedevhhh:
made several minor changes. Most of them are triple quotation string to make it easier to read. Per exemple

(after)
f"""

{some var here}\n
{some other var here}

"""

(before)
f"" {some var here\n ""
f""{some other var here}""


lebedevhhh:
Why all those failed checks ?

PR #4768, opened 2022-08-15
fix one typo and two pycodestyle suggestions

fchapoton:
Just fixing one small typo in "request" in the documentation.

and also two very small suggestions of pycodestyle in the code, about no space before colons


davidism:
Thanks for looking at this. Is there some sort of event going on right now encouraging new users to make small changes? This is like the fifth one we've got recently.

I've fixed this in 2.2.x 2d3f7257, which is what docs changes should be made against. You'll notice the style changes were automatically reverted by our own style tools. All of this is mentioned in our contributing docs.

PR #4776, opened 2022-08-21
Re-add `*args` to the `view` function in `View` classes

greyli:
#4624 added a `View.init_every_request` attribute, but also removed the `*args` for the `view` function and `MethodView.dispatch_request` method.

It could cause unexpected behavior in applications or libraries that rely on the original implementation. For example, APIFlask rewrote the `Flask.dispatch_request` method to pass view args as positional arguments to the view function to ensure [the natural order of arguments](https://apiflask.com/comparison/#a-natural-order-of-view-arguments) (this change broke the class-based view support https://github.com/apiflask/apiflask/issues/341).

It seems an undocumented change, and I'm not sure the motivation behind this change. Is it possible to revert it?

### DIFF
--- a/src/flask/views.py
+++ b/src/flask/views.py
@@ -100,17 +100,17 @@ class View:
         """
         if cls.init_every_request:
 
-            def view(**kwargs: t.Any) -> ft.ResponseReturnValue:
+            def view(*args: t.Any, **kwargs: t.Any) -> ft.ResponseReturnValue:
                 self = view.view_class(  # type: ignore[attr-defined]
                     *class_args, **class_kwargs
                 )
-                return current_app.ensure_sync(self.dispatch_request)(**kwargs)
+                return current_app.ensure_sync(self.dispatch_request)(*args, **kwargs)
 
         else:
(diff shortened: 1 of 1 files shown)


davidism:
Views, regardless of if they are class based, only receive keyword args.

greyli:
OK. Does it mean I have to rewrite the whole `flask.view` module in APIFlask if I want to add `*args` to the `view` function in the `View` class and the `dispatch_request` of `MethodView`?

davidism:
No view ever received anything in `*args`. If it did, you were already rewriting something. I do not want to enable positional arguments in Flask itself, the intention seems to always have been keyword arguments.

davidism:
I see what you're doing in APIFlask. But I think it would make just as much sense to change APIFlask instead to pass everything as keyword arguments. You can have the default arg name for query args be `query`, and the default name for the body be `data`. You can have a `arg_name=` argument to decorators to change what they are passed as. That way, users _really_ don't have to worry about order.

greyli:
Good idea, thanks! I will do some experiments later. In the long run, this will be a better solution.

PR #4778, opened 2022-08-21
Improve the description of flask run command

### LINKED ISSUE #4777: CLI flask run describes that uses --debug to enable debugger and reloader, but flask run does not have the -- debug option
flask run --help description:   The reloader and debugger are enabled by default with the '--debug' option.
But when I use flask run --debug, it tells me that there is no --debug option

flask run --debug

image

According to the description of flask run --help, the --debug option should be supported.

Environment:

- Python version: 3.10.6
- Flask version: 2.2.2

greyli:
Point out that the `--debug` option belongs to `flask` instead of `flask run`.

- fixes #4777

Maybe we could add a separate `--debug` to `flask run` since it's common to use `--debug` with `run` (see #4779).

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -890,8 +890,8 @@ def run_command(
     This server is for development purposes only. It does not provide
     the stability, security, or performance of production WSGI servers.
 
-    The reloader and debugger are enabled by default with the '--debug'
-    option.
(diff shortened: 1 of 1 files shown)


davidism:
Going with #4779.

PR #4779, opened 2022-08-21
Add `--debug` option to flask run

### LINKED ISSUE #4777: CLI flask run describes that uses --debug to enable debugger and reloader, but flask run does not have the -- debug option

greyli:
The alternative of #4778, fixes #4777

Using `flask run` with `--debug` is common usage, and adding this option to `flask run` makes the command more intuitive:

```
flask run --debug
```

If this change is accepted, then I will update the docs and add tests.

Checklist:


davidism:
It's a bit confusing to have both `--debug` and `--debugger` flags, but I'm guessing most people don't know about the second flag anyway. This mostly matches the signature for `app.run`, which calls them `debug` and `use_debugger`.

greyli:
Rebased to 2.2.x and updated docs. I'm not sure if we need to add tests here, I was thinking something like this:

```py
@pytest.mark.parametrize(
    "option,value",
    (
        (["--debug"], True),
        (["--no-debug"], False),
    )
)
def test_run_debug(option, value):
    ctx = run_command.make_context("run", option)
    assert ctx.params["debug"] == value
```

But it seems unnecessary.

greyli:
> It's a bit confusing to have both `--debug` and `--debugger` flags, but I'm guessing most people don't know about the second flag anyway. This mostly matches the signature for `app.run`, which calls them `debug` and `use_debugger`.

When this option was added, I think most people will use `--debug` over `--reload` and `--debugger`.

BTW, `--reloader` matches `--debugger` and `use_reloader` better. Do you think we need to rename `--reload` to `--reloader`?

davidism:
I agree we should change `reload` to `reloader` as well, but I want to wait until Click can show a deprecation message for individual options.

PR #4799, opened 2022-09-02
Close p tag in templateinheritance docs

danielcompton:
I noticed this when going through the docs.


davidism:
The HTML is valid as is.

PR #4809, opened 2022-09-07
add a `reset_blueprint` method

### LINKED ISSUE #4786: Nested blueprint registration causes error when using test-scoped app fixtures
For apps that use nested blueprints, the way that Flask currently tracks how child blueprints have been registered causes both an error and a warning (that will soon be an error) when creating multiple app objects. This is a problem when testing.

For the below code, `test_2` will fail because `child` has already been registered on `parent`, resulting in
~~~
ValueError: The name 'child' is already registered for this blueprint 'parent.child'. Use 'name=' to provide a unique name.
~~~

Additionally, the second call to `parent.register_blueprint()` emits a warning:
~~~
UserWarning: The setup method 'register_blueprint' can no longer be called on the blueprint 'parent'. It has already been registered at least once, any changes will not be applied consistently.
  Make sure all imports, decorators, functions, etc. needed to set up the blueprint are done before registering it.
  This warning will become an exception in Flask 2.3.
    parent.register_blueprint(child, url_prefix="/child")
~~~

~~~py
import pytest

from flask import Blueprint, Flask

parent = Blueprint("parent", __name__)
child = Blueprint("child", __name__)


def create_app():
    app = Flask(__name__)

    parent.register_blueprint(child, url_prefix="/child")
    app.register_blueprint(parent, url_prefix="/parent")

    return app


@pytest.fixture
def app():
    app = create_app()
    yield app


def test_1(app):
    pass


def test_2(app):
    pass
~~~

User workarounds:

- Session-scope the app fixture — this is what I'm currently doing but creates issues with test isolation in `FlaskLoginClient` from Flask-Login (but I worked around that too)
- Create the blueprint objects inside `create_app()` — this isn't ideal because it makes registering routes on the blueprints more complex
- Alter the parent blueprint's private attrs in the app fixture — adding these lines after the `yield` statement also fixes the problem for a given parent blueprint but is not a very flexible solution overall:
~~~py
parent._blueprints = []
parent._got_registered_once = False
~~~

Environment:

- Python version: 3.8.10
- Flask version: 2.2.2

rasimandiran:
Added `reset_blueprint` method to the `Blueprint` class. This method re-init the `_blueprints` list as an empty list and sets the `_got_registered_once` flag to False

- fixes #4786

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Unreleased
 Version 2.2.3
 -------------
 
-Unreleased
+-   Added reset_blueprint method to prevent error and warning when using test-scoped app fixtures :issue:`#4786`
 
 
 Version 2.2.2
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -472,6 +472,8 @@ class Blueprint(Scaffold):
     def reset_blueprint(self):
         """Resets this blueprint. Clears registered child blueprints and
         set _got_registered_once flag to False
+
+        .. versionchanged:: 2.2.3
         """
         self._blueprints = []
         self._got_registered_once = False
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -469,6 +469,13 @@ class Blueprint(Scaffold):
             bp_options["name_prefix"] = name
             blueprint.register(app, bp_options)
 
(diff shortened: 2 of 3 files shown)


davidism:
I don't think this is the correct solution, only an application of the workaround mentioned in the issue. This needs more discussion first.

PR #4856, opened 2022-10-31
Fix subdomain inheritance for nested blueprints.

### LINKED ISSUE #4834: Nested blueprints are not respected when mounted on subdomains
Hello, and thanks for all your work 🙏🏻 

Nested blueprints [as described in the docs](https://flask.palletsprojects.com/en/2.2.x/blueprints/#nesting-blueprints) work perfectly fine when using `url_prefix`. However, when mounting the parent blueprint using a subdomain, the child routes are not accessible.

```python
from flask import Flask
from flask import Blueprint

app = Flask(__name__)
app.config["SERVER_NAME"] = "localhost:5000"
parent = Blueprint("parent", __name__)
child = Blueprint("child", __name__)

@app.route('/')
def index():
    return "index"

@parent.route('/')
def parent_index():
    return "parent"

@child.route('/child/')
def child_index():
    return "child"

app.register_blueprint(parent, subdomain="api")


if __name__ == '__main__':
    app.run(debug=True)
```

The index route works as expected:

```
❯ http http://localhost:5000/
HTTP/1.1 200 OK
Connection: close
Content-Length: 5
Content-Type: text/html; charset=utf-8
Date: Tue, 04 Oct 2022 10:44:10 GMT
Server: Werkzeug/2.2.2 Python/3.10.4

index
```

So does the parent route in the subdomain:

```
❯ http http://api.localhost:5000/
HTTP/1.1 200 OK
Connection: close
Content-Length: 6
Content-Type: text/html; charset=utf-8
Date: Tue, 04 Oct 2022 10:44:06 GMT
Server: Werkzeug/2.2.2 Python/3.10.4

parent
```

But the child responds with a 404:

```
❯ http http://api.localhost:5000/child/
HTTP/1.1 404 NOT FOUND
Connection: close
Content-Length: 207
Content-Type: text/html; charset=utf-8
Date: Tue, 04 Oct 2022 10:45:42 GMT
Server: Werkzeug/2.2.2 Python/3.10.4

<!doctype html>
<html lang=en>
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.</p>
```

If we change the `subdomain="api"` for `url_prefix="/api"` when registering the blueprint however everything works as expected:

```
❯ http http://localhost:5000/api/
HTTP/1.1 200 OK
Connection: close
Content-Length: 6
Content-Type: text/html; charset=utf-8
Date: Tue, 04 Oct 2022 10:46:53 GMT
Server: Werkzeug/2.2.2 Python/3.10.4

parent

❯ http http://localhost:5000/api/child/
HTTP/1.1 200 OK
Connection: close
Content-Length: 5
Content-Type: text/html; charset=utf-8
Date: Tue, 04 Oct 2022 10:46:59 GMT
Server: Werkzeug/2.2.2 Python/3.10.4

child
```

This was surprising to me as I expected the same nesting to apply regardless of whether the parent is mounted using a subdomain or a URL prefix. Am I missing something?

Environment:

- Python version: 3.10
- Flask version: 2.2.2

confuzeus:
When child blueprints are mounted onto parents, they don't inherit the subdomain set by the parent.

This PR ensures that routes in child blueprint will be available at the same subdomain set by the parent.

However, children can also specify their own subdomain if necessary.

- fixes #4834

Checklist:

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -453,6 +453,15 @@ class Blueprint(Scaffold):
         for blueprint, bp_options in self._blueprints:
             bp_options = bp_options.copy()
             bp_url_prefix = bp_options.get("url_prefix")
+            bp_subdomain = bp_options.get("subdomain")
+
+            if bp_subdomain is None:
+                bp_subdomain = blueprint.subdomain
+
+            if state.subdomain is not None and bp_subdomain is None:
+                bp_options["subdomain"] = state.subdomain
+            elif bp_subdomain is not None:
+                bp_options["subdomain"] = bp_subdomain
 
             if bp_url_prefix is None:
                 bp_url_prefix = blueprint.url_prefix
--- a/tests/test_blueprints.py
+++ b/tests/test_blueprints.py
@@ -950,6 +950,53 @@ def test_nesting_url_prefixes(
     assert response.status_code == 200
 
 
+def test_nesting_subdomains(app, client) -> None:
+    subdomain = "api"
+    parent = flask.Blueprint("parent", __name__)
+    child = flask.Blueprint("child", __name__)
+
+    @child.route("/child/")
+    def index():
+        return "child"
+
+    parent.register_blueprint(child)
+    app.register_blueprint(parent, subdomain=subdomain)
+
+    client.allow_subdomain_redirects = True
+
+    domain_name = "domain.tld"
+    app.config["SERVER_NAME"] = domain_name
+    response = client.get("/child/", base_url="http://api." + domain_name)
+
+    assert response.status_code == 200
+
+
+def test_child_overrides_parent_subdomain(app, client) -> None:
+    child_subdomain = "api"
+    parent_subdomain = "parent"
+    parent = flask.Blueprint("parent", __name__)
+    child = flask.Blueprint("child", __name__, subdomain=child_subdomain)
+
+    @child.route("/")
+    def index():
+        return "child"
+
+    parent.register_blueprint(child)
+    app.register_blueprint(parent, subdomain=parent_subdomain)
+
+    client.allow_subdomain_redirects = True
+
+    domain_name = "domain.tld"
+    app.config["SERVER_NAME"] = domain_name
+    response = client.get("/", base_url=f"http://{child_subdomain}.{domain_name}")
+
+    assert response.status_code == 200
+
+    response = client.get("/", base_url=f"http://{parent_subdomain}.{domain_name}")
+
+    assert response.status_code == 404
+
+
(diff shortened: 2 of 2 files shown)


src/flask/blueprints.py
```diff
@@ -453,6 +453,15 @@ def extend(bp_dict, parent_dict):
         for blueprint, bp_options in self._blueprints:
             bp_options = bp_options.copy()
             bp_url_prefix = bp_options.get("url_prefix")
+            bp_subdomain = bp_options.get("subdomain")
+
+            if bp_subdomain is None:
+                bp_subdomain = blueprint.subdomain
+
+            if state.subdomain is not None and bp_subdomain is None:
```
pgjones, line 461: I think we should also consider what to do if `state.subdomain is not None and bp_subdomain is not None`. I think we should combine as `state.bp` (much like with the prefix). Would you expect it to work this way as well?

confuzeus: Yes good point.

I thought that if both parent and child have subdomain set, it would be weird to combine. It makes sense for paths but subdomains in the form foo.bar.baz are rare.

For example if parent is bar and child is foo, we would get foo.bar.

But I can get this by supplying foo.bar in the child blueprint subdomain. So I went with the child overrides parent subdomain way that way the user can use a combined subdomain as needed.

Not sure this is a good idea though because some use cases might need them to be combined by default.

Another possible solution would be to introduce an extra parameter called something like "combine_subdomains". This will also allow different child blueprint to have different subdomain strategies. For example, foo can choose to be combined but baz can choose not to.

What do you think?

pgjones: Given it might confuse it is probably safest to raise a ValueError/RuntimeError then if both are set. (This would be raised during initialization and so isn't likely to cause any issues for existing apps)

confuzeus: It could be an issue if the user wants to combine the subdomain anyway.

For example, parent blueprint's subdomain is *bar* then user wants child blueprint X to be available at "foo.bar" so he sets the subdomain. Then another blueprint Y needs to be at "baz.bar" so he sets the subdomain like that. But parent needs to be available at *bar* subdomain no matter what.

Raising ValueError would break this use case.

pgjones: I've had quite a bit of time to think about this, and to talk to others about their expectations. I think we should combine the subdomains (much like we do for the paths). I think this is the clearest expectation for a nested blueprint i.e. that the routes defined within it nest both by path and subdomain. Could you update this PR to do so and then I'll merge.

davidism:
continued in #4935

PR #4857, opened 2022-10-31
Add disclaimer for proper error message handling in the documentation

Jhnbrn90:
### Introduction
I think we should propagate that the used examples for the error messages in the section on Logging in and Registering, are purely used for debugging purposes and to clearly see that the code works as expected. I think it would be good to note that these error messages should be generic as detailed by the OWASP project and considered best practice from a security point of view. 

### What has been done
As the Flask tutorial explicitly returns messages that might reveal useful information to potential attackers, a section was added to raise awareness that in a real world application the reader might want to use more generic error messages as advised by the OWASP project (a link to the documentation of OWASP is provided).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
         args: ["--application-directories", "src"]
         additional_dependencies: ["setuptools>60.9"]
   - repo: https://github.com/psf/black
-    rev: 22.6.0
+    rev: 22.8.0
     hooks:
       - id: black
   - repo: https://github.com/PyCQA/flake8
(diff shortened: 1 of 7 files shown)


davidism:
I've never been particularly convinced by the argument that "Username already exists" is a security issue, the OWASP page doesn't go into any detail explaining why. It depends on the application and whether usernames are a publicly known value or not. I don't want to add that to an already complex tutorial.

Jhnbrn90:
Allright, I understand. Thanks 👍 

Sorry for the messy commits in the PR by the way, I was following the tutorial and tried to use the 'Edit on GitHub' link within the ReadTheDocs menu, but started from the wrong branch etc. I should've git cloned the project instead 😅  .

PR #4869, opened 2022-11-16
Explain AssertionError on the new @setupmethod 'shell_context_processor' behavior

### LINKED ISSUE #4868: docs should explain the `@setupmethod` behavior
The [appcontext documentation](https://flask.palletsprojects.com/en/2.2.x/appcontext/) should explain the new @setupmethod 'shell_context_process' behavior, linked to https://github.com/pallets/flask/pull/4577.

Flask raises an Assertion Error.

> AssertionError: The setup method 'shell_context_processor' can no 
    longer be called on the application. It has already handled its 
    first request, any changes will not be applied consistently. Make
    sure all imports, decorators, functions, etc. needed to set up the 
    application are done before running it.

The error should be documented, and a correct implementation should be provided.

HLFH:
This appcontext updated doc explain the new @setupmethod 'shell_context_processor' behavior and gives guidance to developers so they can start refactoring their code and define the Flask application object under a basic application factory `create_app` to avoid the AssertionError.

- fixes #4868 

I think the updated doc page is still wrapped at 72 characters.
I don't think I need to an entry in CHANGES.rst or include .. versionchanged:: inline changelogs in relevant docstrings?

### DIFF
--- a/docs/appcontext.rst
+++ b/docs/appcontext.rst
@@ -66,11 +66,7 @@ Flask automatically *pushes* an app context when handling a request. Flask canno
     first request, any changes will not be applied consistently. Make
     sure all imports, decorators, functions, etc. needed to set up the
     application are done before running it.
-
-    This typically means that you attempted to use functionality that
-    needed to interface with the current application object in some way.
-    To solve this, set up an application context with app.app_context().
-
+    
 
 Manually Push a Context
 -----------------------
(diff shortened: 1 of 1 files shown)


davidism:
I started to leave a review, but really this whole explanation is incorrect. You've misunderstood when and why modifying an app after setup is an error.

HLFH:
@davidism https://github.com/pallets/flask/issues/4868#issuecomment-1317522589

Thanks

PR #4871, opened 2022-11-17
Code Optimized

praveen2010:
- fixes #<issue number>

Checklist:

### DIFF
--- a/src/flask/blueprints.py
+++ b/src/flask/blueprints.py
@@ -272,11 +272,7 @@ class Blueprint(Scaffold):
         self.url_prefix = url_prefix
         self.subdomain = subdomain
         self.deferred_functions: t.List[DeferredSetupFunction] = []
-
-        if url_defaults is None:
-            url_defaults = {}
-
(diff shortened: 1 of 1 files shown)


src/flask/blueprints.py
```diff
@@ -272,11 +272,7 @@ def __init__(
         self.url_prefix = url_prefix
         self.subdomain = subdomain
         self.deferred_functions: t.List[DeferredSetupFunction] = []
-
-        if url_defaults is None:
-            url_defaults = {}
-
-        self.url_values_defaults = url_defaults
+        self.url_values_defaults = {} if url_defaults is None else url_defaults
```
Abdur-rahmaanJ, line 275: This code fix does not seem to change anything.

praveen2010: Yes, But seems in single line. As python provide one line if else

ThiefMaster: Both options are perfectly valid, please do not send PRs with such minor changes. They are more noise than useful.

PR #4917, opened 2022-12-22
Set `zip(strict)` flag explicitly for Python 3.10

Dasha2205:
- fixes #<issue number>

Checklist:


greyli:
Thanks for working on this.

The `zip(strict)` flag was added in Python 3.10. Since we need to support Python < 3.10, I think we should ignore style check on this flag instead of adding this flag to all `zip` calls (#4919).

Please add a proper title and description for the pull request next time. And the bugfixes should be based on the latest dev branch (e.g. 2.2.x) instead of main. Check out the [contributing guide](https://github.com/pallets/flask/blob/main/CONTRIBUTING.rst) for more details.

PR #4919, opened 2022-12-22
Ignore style check for zip strict flag

greyli:
An alternative of #4917.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,6 +80,8 @@ ignore =
     E722
     # bin op line break, invalid
     W503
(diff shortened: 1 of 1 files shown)


davidism:
Let's put this globally in `setup.cfg`, with a comment like "requires Python 3.10". For this sort of compatibility stuff, I search "Python 3.x" etc. when dropping support, so I can undo this when needed.

greyli:
Updated.

davidism:
Handled in #4922 while I was fixing the docs build too. Missed that you updated this, sorry.

PR #4921, opened 2022-12-22
Allow pathlib.Path for template_folder

### LINKED ISSUE #4892: Allow template_folder to be a pathlib.Path object
When constructing the Flask object, it would be nice if template_folder could be either a string (currently) or a pathlib.Path object. This would match static_folder, which is currently typed as `Union[str, os.PathLike]`. That functionality was added in  #3579; I was unable to find any reason why the same wasn't done for template_folder, so I assume it was an unintentional omission.

https://flask.palletsprojects.com/en/2.2.x/api/#application-object

Not having this functionality is not a dealbreaker, as it's pretty simple to do `template_folder=str(path)`, but it would be nice and match static_folder. In SecureDrop our code is going to look like:
```python3
    app = Flask(
        __name__,
        template_folder=str(config.JOURNALIST_TEMPLATES_DIR.absolute()),
        static_folder=config.STATIC_DIR,
    )
```

...so implementing this would be nice.

ivanovmg:
This PR allows ``template_folder`` argument to be a ``pathlib.Path`` object.

- fixes #4892

Checklist:

I added a new test, which ensures that ``template_folder`` can be ``pathlib.Path`` object.
However, when I added this tests, it was already successful.
I noticed that there are no similar tests of this kind in regards to ``template_folder``.

I suppose that the issue referenced is related to typing only, because according to the successful test, one can use this out of the box, and there is no need to convert it into a str object.
``` python
app = Flask(
    __name__,
    template_folder=pathlib.Path("templates"),
)
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@ Version 2.2.3
 
 Unreleased
 
+-   Fix the type of ``template_folder`` to accept ``pathlib.Path``. :issue:`4892`
+
 
 Version 2.2.2
 -------------
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -558,7 +558,7 @@ class Flask(Scaffold):
         static_host: t.Optional[str] = None,
         host_matching: bool = False,
         subdomain_matching: bool = False,
-        template_folder: t.Optional[str] = "templates",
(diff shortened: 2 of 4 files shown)


legoktm:
Thank you!

PR #4927, opened 2022-12-26
Update debugger flag

s2t2:
I am running Flask 2.2.2. The `--debug` option that's currently mentioned in the docs leads to an error, while changing it to `--debugger` works! So updating the docs.

> NOTE: this is a documentation only change


davidism:
This is not correct. The issue is that we added the `run --debug` flag and changed the docs to reflect that, but 2.2.3 isn't released yet. Use `flask --app example --debug run` for now.

PR #4934, opened 2023-01-01
Rename `id` variable and mentions to `id_`

matheusfelipeog:
[`id` is the name of a built-in Python function](https://docs.python.org/3/library/functions.html#id). Use `id_` instead, as indicated by [pep8](https://peps.python.org/pep-0008/#descriptive-naming-styles).


ThiefMaster:
I disagree. Especially the `id` builtin is very rarely used in regular code that deals with web requests, and using `id_` just makes the code harder to read.

matheusfelipeog:
@ThiefMaster `id` not being used much in this context doesn't seem to be a reason not to rename it according to convention.

You're right about `post_id` being better. Although this is an example, I think it's good to use different names than the built-in functions.

Reconsider that I can change them to `post_id` if allowed

Issue #4954, opened 2023-01-26
Why do docs recommend opening/closing sqlite3 connection for each request?


marcospgp:
In [this page](https://flask.palletsprojects.com/en/2.2.x/patterns/sqlite3/) the Flask docs present opening and closing a connection per request as the "official" way of working with sqlite3.

This does not make sense to me - surely opening and closing connections has some overhead, which can be avoided if a connection is shared across an entire application. Perhaps this overhead is negligible here, but I still couldn't get over it.

The problem is that a global connection created with `sqlite3.connect(DATABASE)` will raise an `Objects created in a thread can only be used in that same thread` error, due to the way (I assume) Flask processes web requests in worker threads.

As it turns out and [as explained here](https://ricardoanderegg.com/posts/python-sqlite-thread-safety/) this can be safely resolved by creating the connection with `sqlite3.connect(DATABASE, check_same_thread=False)` instead, as long as `sqlite3.threadsafety == 3` (the available variant of sqlite3 is in serialized mode).

Given this information, shouldn't the documentation suggest a global shared connection as the standard way of working with sqlite3?

ThiefMaster:
If you use sqlite, performance is unlikely to be your main goal anyway. It's just something simple that works well for an example app.

In a real application you are likely using sqlalchemy or something else to maintain a connection pool and thus do not have the problem. Connecting to the sqlite database on startup / on first request just adds complexity though, for no real gain.

Also, by sharing a connection there's a good chance that you cannot have separate transactions for separate parallel requests, can you?

This is something on that topic I fond when googling "sqlite one connection per request": https://sqlite.org/forum/info/ff2fe46b1ff102bfbc78c112ea4eb25105354571b6c7551956d3cd72fe3e4624

marcospgp:
> It's just something simple that works well for an example app.

But then maybe that should at least be made clear in the docs? Currently it is more or less implied that the presented way is the best approach.

I think this is particularly bad because one may interpret this to mean that it is always the best approach to open a connection per request, for any database!

> Also, by sharing a connection there's a good chance that you cannot have separate transactions for separate parallel requests, can you?

It does make sense that if Flask is multithreaded (perhaps you can clear this up for me), transactions that consist in more than one call to `connection.execute()` could find themselves mixed with some other thread's calls.
Although that seems like a smaller issue to me, and one that could be pointed out as a use case for having a request-specific connection.

[...]

PR #4955, opened 2023-01-26
Add more detail to connection approach in sqlite3.rst

marcospgp:
Addresses https://github.com/pallets/flask/issues/4954 by adding a little more detail to documentation.

This should make it clear that one connection per request is a simple approach to using SQLite 3 but not necessarily the only one, or the best in terms of efficiency (opening connections may have [significant overhead](https://sqlite.org/forum/forumpost/2d2a88dde6249962), especially in the context of web requests where ms can be taken into account).

### DIFF
--- a/docs/patterns/sqlite3.rst
+++ b/docs/patterns/sqlite3.rst
@@ -1,11 +1,16 @@
 Using SQLite 3 with Flask
 =========================
 
-In Flask you can easily implement the opening of database connections on
-demand and closing them when the context dies (usually at the end of the
-request).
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, however I don't think this is necessary for the purpose of these docs.

marcospgp:
As is, the docs imply the one correct approach to using SQLite with Flask is to create a database connection for each HTTP request, which is not the case. SQLite is thread safe by default, see here: https://github.com/python/cpython/pull/101351

I would urge you to reconsider in the least clarifying that the approach in the documentation is simply a suggestion and that alternatives exist.

PR #4957, opened 2023-01-31
improving test_cli by adding test_enable_dotenv_from_env tests

andrehora:
Improved `test_cli` with cases to enable dotenv from env (with the envvar `FLASK_SKIP_DOTENV`), which is currently untested.

We can enable dotenv from env with three values: "0", "false", or "no".
See: https://github.com/pallets/flask/blob/main/src/flask/helpers.py#L81

Thus, this PR creates three tests:

- test_enable_dotenv_from_env_zero
- test_enable_dotenv_from_env_false
- test_enable_dotenv_from_env_no

Checklist:

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -562,6 +562,30 @@ def test_disable_dotenv_from_env(monkeypatch, runner):
     assert "FOO" not in os.environ
 
 
+@need_dotenv
+def test_enable_dotenv_from_env_zero(monkeypatch, runner):
+    monkeypatch.chdir(test_path)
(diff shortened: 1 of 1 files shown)


davidism:
* `FLASK_SKIP_DOTENV` is false by default, this test doesn't make sense.
* The actual function to test is `flask.helpers.get_load_dotenv`, there's no reason to invoke the CLI.
* Tests like this should be parametrized rather than copied.

Based on the first point, I don't plan to merge this.

andrehora:
Thanks for the feedback! Agreed!

PR #5034, opened 2023-03-26
Allow Response subclasses to set default_mimetype as None

adamtheturtle:
This matches the type - see https://github.com/pallets/werkzeug/blob/92c6380248c7272ee668e1f8bbd80447027ccce2/src/werkzeug/sansio/response.py#L94.

### DIFF
--- a/src/flask/wrappers.py
+++ b/src/flask/wrappers.py
@@ -151,7 +151,7 @@ class Response(ResponseBase):
         Added :attr:`max_cookie_size`.
     """
 
-    default_mimetype = "text/html"
+    default_mimetype: t.Optional[str] = "text/html"
 
     json_module = json
 


davidism:
The linked annotation should be fixed instead.

adamtheturtle:
In case it changes your mind:

My use case is that I want to give responses without a `Content-Type` header.
In particular, I am making a mock of a service which does that, and I want to copy that service.

adamtheturtle:
@davidism There's even [this](https://github.com/pallets/werkzeug/blob/92c6380248c7272ee668e1f8bbd80447027ccce2/src/werkzeug/test.py#L1282-L1284) in the Werkzeug codebase:

```
    .. versionchanged:: 2.2
        Set the ``default_mimetype`` to None to prevent a mimetype being
        assumed if missing.
```

[and](https://github.com/pallets/werkzeug/blob/92c6380248c7272ee668e1f8bbd80447027ccce2/src/werkzeug/test.py#L1293-L1294):

```
    default_mimetype = None
    # Don't assume a mimetype, instead use whatever the response provides
```

- That is exactly what I want to achieve here.

PR #5036, opened 2023-03-27
Use latest patch of Werkzeug

Samira-El:
This is pulling a security fix for a vulnerability in Werkzeug 2.2.2
A patch has been release https://github.com/pallets/werkzeug/releases/tag/2.2.3

- fixes N/A (see security report)

Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 ]
 requires-python = ">=3.7"
 dependencies = [
-    "Werkzeug>=2.2.2",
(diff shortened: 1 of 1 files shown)


davidism:
The fact that Flask works with a previous version of Werkzeug isn't itself a security issue.

When writing an application, you _must_ use a tool like [pip-tools](https://pypi.org/project/pip-tools/) to pin your application's full dependency tree. This gives you reproducible deployments, allowing you to control when you get updates.

PR #5073, opened 2023-04-20
use new logo

davidism:
We've got new logos! I'll announce that officially at some point, but that kept getting pushed back. So I'll just start adding them as I get to things.


digglife:
At one moment I thought I was visiting a fake site when I saw the new logo. 😂 Is it a horn? Or a bottle for alcohol?

beucismis:
I like more the old logo.

PR #5124, opened 2023-05-14
Json encoder support Set object

idoshr:
Json encoder support Set object
and automatically encode to list

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+Development
+-------------
+
+-   Set object will automaticly encode to list on jsonify
+
+
 Version 2.3.2
 -------------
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Not sure if this is a good idea. For example, it's not clear whether you' want to sort the set elements before converting them to a list (usually I do that because you go from an unordered to an ordered data type, and having consistency there can make debugging easier).

davidism:
You're welcome to add support for this in a custom provider, but I don't think this should be in the default one.

idoshr:
Ok
I just think that by default for simple user it will be better if that will be automatically converted 
But that your call
🙂🙂🙂

PR #5125, opened 2023-05-14
Add basic Dockerfile for development

Mr-Destructive:
- fixes #<issue number>

Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:3.10
+
+WORKDIR /app
+
+COPY . .
+
+RUN python3 -m venv venv
+
+RUN python3 -m pip install .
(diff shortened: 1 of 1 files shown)


Mr-Destructive:
Wouldn't it be great to add a Dockerfile for testing the changes locally without any installation on the local system?
Have added a basic Dockerfile that builds the current flask package and runs the test. This can be extended by a bash interactive shell for specific commands from the container.

Please let me know of any changes or if should this be really added.
Thanks you :)

davidism:
This isn't how docker is supposed to be used. You build something, then use it to run something. You don't run everything during the build step. We also just added a dev container, you can get an entire dev environment in your browser.

PR #5127, opened 2023-05-15
Split the App and Blueprint into Sansio and IO parts

pgjones:
This follows a similar structure in Werkzeug and allows for async based IO projects, specifically Quart, to base themselves on Flask.

Note that the globals, and signals are specific to Flask and hence specific to Flask's IO. This means they cannot be moved to the sansio part of the codebase. (We may be able to change this in the future, but I think we should let this change settle first).

Checklist:


src/flask/scaffold.py
```diff
@@ -1,923 +1,10 @@
...
-        # Unix (prefix/lib/pythonX.Y/site-packages)
-        if os.path.basename(parent).lower() == "lib":
-            return os.path.dirname(parent), package_path
-
-        # something else (prefix/site-packages)
-        return site_parent, package_path
-
-    # not installed
-    return None, package_path
+class Scaffold(SansioScaffold, _StaticMixin):
```
pgjones, line 9: Remove, likely not public API.

src/flask/_static_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+class ScaffoldMixin(t.Protocol):
+    has_static_folder: bool
+    root_path: str
+    static_folder: str | None
+
+    def get_send_file_max_age(self, filename: str | None) -> int | None:
+        ...
+
+
+class _StaticMixin:
```
pgjones, line 21: Remove, implement in Blueprint and Flask (dupe is ok).
Note "This isn't Sansio, hence dupe in docstring"

pgjones (commented):
Split commits into file moves and refactoring commits.

davidism:
Should be able to rebase to fix the pre-commit failure.

davidism:
Remind me what the issue with using `current_app` and `request` is? Was it typing?

src/flask/app.py
```diff
@@ -365,124 +226,18 @@ def __init__(
         root_path: str | None = None,
     ):
         super().__init__(
-            import_name=import_name,
-            static_folder=static_folder,
-            static_url_path=static_url_path,
-            template_folder=template_folder,
-            root_path=root_path,
+            import_name,
```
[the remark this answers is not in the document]
pgjones: Fixed

pgjones:
> Remind me what the issue with sharing current_app and request is? Was it typing?

Partially, as the wrong type does make things difficult for users. Which on a related typing note, I think the next step for this (ideally post merge) is to make the sansio parts Generic in the response return type, to allow for Quart to use async iterators (which is too hard I think in Flask to achieve) and for Flask's typing to again be narrowed.

The main issue is I'd like to keep https://github.com/pgjones/quart-flask-patch working, which will require flask's current_app and request to be different instances to quart's.

PR #5147, opened 2023-05-30
Mention about sites-enabled directory

Hritik14:
This is only a docs suggestion, so I have not checked the list below. A better wording to communicate the same fact should be helpful.

Checklist:


davidism:
Different distributions put config files in different places, and there's no requirement to split into different site files to begin with, so I'll leave this as is.

PR #5177, opened 2023-06-27
use flit_core as build backend

davidism:
flit_core is simpler and faster than setuptools. It doesn't require a `MANIFEST.in` file, which simplifies the tutorial and examples as well.


davidism:
:facepalm: actually removed `MANIFEST.in` in f39c7160

PR #5195, opened 2023-07-08
force return json instead of raise error for request.json

mouday:
force return json instead of raise error for `request.json`


ThiefMaster:
No. You can do that in your own application if you want this behavior.

PR #5204, opened 2023-07-17
Main

### LINKED ISSUE #5203: Give possibility to change the MAX_FORM_PARTS of werkzeug
## Adding  MAX_FORM_PARTS to the config of flask
This feature allows for the customization of the maximum number of form parts when reading a request. The max_form_parts parameter has been introduced in Werkzeug version 2.2.3 - https://github.com/pallets/werkzeug/blob/4048509fb24196df67f944a24aaa4ce1afcba92f/src/werkzeug/wrappers/request.py#L90.

Real use case: Apache Superset uses Flask-AppBuilder for editing user roles in the application. Currently, when there are roles with 1000+ users, the editing process fails with a 413 error due to the max_form_parts limit, which cannot be changed in the current version of Flask.

Always-prog:
- fixes #5203

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 -   Require Werkzeug >= 2.3.6.
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
 -   Refactor how an app's root and instance paths are determined. :issue:`5160`
+-   Added MAX_FORM_PARTS to the app config.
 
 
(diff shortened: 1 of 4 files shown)


Always-prog:
Hello!
Do I need to add test cases for that config key?

ThiefMaster:
As mentioned in your issue, this PR is not necessary because you can already override it pretty easily.

PR #5217, opened 2023-08-12
Update dispatch by path example

### LINKED ISSUE #5214: Application Dispatching Docs Reference Removed `werkzeug` Function
The ["Dispatch by Path" section of app dispatching docs](https://github.com/pallets/flask/blob/main/docs/patterns/appdispatch.rst#dispatch-by-path) reference now-removed functions, `werkzeug.wsgi.peek_path_info` and `werkzeug.wsgi.pop_path_info`.

These functions were [deprecated in Werkzeug 2.2.0](https://github.com/pallets/werkzeug/pull/2461) and [removed in Werkzeug 2.3.0](https://github.com/pallets/werkzeug/pull/2592).

Using the documented example as-is produces `ImportError`s.

Since Flask 2.3 requires Werkzeug 2.3.6 or later, the documentation contain an example that works with Werkzeug's updated API. 

Environment:

- Python version: 3.9
- Flask version: 2.3.2

pavithra-m13:
- fixes #5214

### DIFF
--- a/docs/patterns/appdispatch.rst
+++ b/docs/patterns/appdispatch.rst
@@ -18,33 +18,19 @@ Working with this Document
 --------------------------
 
 Each of the techniques and examples below results in an ``application``
-object that can be run with any WSGI server. For production, see
-:doc:`/deploying/index`. For development, Werkzeug provides a server
-through :func:`werkzeug.serving.run_simple`::
+object that can be run with any WSGI server. For development, use the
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, this was a good way to fix the docs. However, it looks like this whole section can be removed since it just recreates manually what `DispatcherMiddleware` already does.

PR #5229, opened 2023-08-20
Pass the request ctx rather than use the globals in the app

pgjones:
The globals have a performance penalty which can be justified for the convinience in user code. In the app however the ctx can easily be passed through the method calls thereby reducing the performance penalty.

This may affect extensions if they have subclassed the app and overridden these methods.

Checklist:


pgjones:
I've done something similar for Quart after finding it had a significant impact on micro benchmarks.

davidism:
My initial reaction is that while this makes sense, I'm uncomfortable with making the change. I'll have to think about this for a little bit.

davidism:
Could you give more information about the benchmarks? How micro are we talking; how fast does the complexity of the app overshadow any speed gains?

Is the speed loss in the use of the proxy, or is the gain in the use of the local variable? Perhaps using `_cv_req.get()` directly and avoiding the proxy could have a similar improvement?

pgjones:
This is an interesting one, with Quart the equivalent to this makes around a 1.5 times improvement, whereas for Flask it doesn't seem to have any affect (maybe a minor positive one). I don't think I can therefore remove it from Quart, but I'd like the APIs to match. 

(I tried `_cv_request` in Quart and it made little difference.)

Also I don't think this makes it more complex - without knowledge of Flask's globals it would be expected to be passed via the call chain.

pgjones:
This will be deferred to allow further discussion.

davidism:
continued in #5818

PR #5231, opened 2023-08-21
bump werkzeug 2.3.7

davidism:
We were already bumping to 2.3.6, and 2.3.7 includes more important fixes than that. Also use `importlib.metadata.version` to prepare for `__version__` being deprecated.


davidism:
Merged into main by accident, cherry picked to 2.3.x in 9f9550247a2798b9dcc28081c15e266442f4e28a

PR #5264, opened 2023-09-24
Allow self as an argument to url_for

### LINKED ISSUE #5258: TypeError: Flask.url_for() got multiple values for argument 'self'
[Official documentation](https://flask.palletsprojects.com/en/2.3.x/api/#flask.url_for) doesn't mention any preconditions for values argument, but `self=1` raises an exception.

Can be replicated with the following test case:

```python
from unittest import TestCase

from flask import Flask, url_for


class UrlForTests(TestCase):
    def setUp(self) -> None:
        self.app = Flask(__name__)

    def test_with_self_argument(self) -> None:
        with (
            self.app.app_context(),
            self.assertRaisesRegex(
                TypeError, "got multiple values for argument 'self'"
            ),
        ):
            url_for("doesnotmatter", self=1)
```

Traceback example:

```
  File "/home/dmitry/git/github.com/minetest/contentdb/tests.py", line 6, in <module>
    url_for("doesnotmatter", self=1)
  File "/home/dmitry/.cache/pypoetry/virtualenvs/contentdb-MK2UWkId-py3.10/lib/python3.10/site-packages/flask/helpers.py", line 256, in url_for
    return current_app.url_for(
TypeError: Flask.url_for() got multiple values for argument 'self'
```

The expected behavior is the returned URL has query string like '?self=1'.

Environment:

- Python version: 3.10.6
- Flask version: 2.3.3

pgjones:
This makes the Flask.url_for self argument positional only (Flask supports Python 3.8+) thereby restoring the ability to pass self as a value argument to url_for.

Closes #5258

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Unreleased
     ``importlib.metadata.version("flask")``, instead. :issue:`5230`
 -   Restructure the code such that the Flask (app) and Blueprint
     classes have Sans-IO bases. :pr:`5127`
+-   Allow self as an argument to url_for. :pr:`5264`
 
 
 Version 2.3.3
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -952,6 +952,7 @@ class Flask(App):
 
     def url_for(
         self,
+        /,
         endpoint: str,
         *,
         _anchor: str | None = None,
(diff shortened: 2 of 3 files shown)


its-joseph-230606:
Good

PR #5274, opened 2023-10-02
Lock werkzeug version to <3 to avoid compatability issue

luong-komorebi:
- I did not file any issue for this because I consider it minor

While it is recommended to using tools like [pip-tools ](https://pypi.org/project/pip-tools/) to manage dependencies, both direct and transitive ones, sometimes for smaller projects, one will just go with a requirements.txt that list only `flask` and meet some issues when werkzeug was just upgraded 3 days ago on Sept 30. 

Therefore, I propose locking the major version of Werkzeug on flask 2.3.x. This change would safeguard users from errors that happen because Werkzeug was upgraded to version 3.x (one of them I have met myself is `ImportError: cannot import name 'url_quote' from 'werkzeug.urls'`)

Checklist: ( _I am checking those but dont  think they are necessary_ )

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "Werkzeug>=2.3.7",
+    "Werkzeug>=2.3.7, <3.0",
     "Jinja2>=3.1.2",
     "itsdangerous>=2.1.2",
     "click>=8.1.3",


davidism:
When writing an application, you _must_ use a tool like [pip-tools](https://pypi.org/project/pip-tools/) to pin your application's full dependency tree. This gives you reproducible deployments, allowing you to control when you get updates. Be sure to run your tests with deprecation warnings treated as errors so that you get notified of those types of changes early.

Please review any of the following for more information:

* https://hynek.me/articles/semver-will-not-save-you/
* https://www.youtube.com/watch?v=WSVFw-3ssXM&t
* https://snarky.ca/why-i-dont-like-semver/
* https://caremad.io/posts/2016/02/versioning-software/
* https://bernat.tech/posts/version-numbers/
* https://iscinumpy.dev/post/bound-version-constraints/

PR #5300, opened 2023-10-18
Update factory.rst

jscmidt:
Wrong argument order, see [StackOverflow](https://stackoverflow.com/questions/75951245/flask-no-such-option-debug)


davidism:
The order is correct, update Flask.

PR #5315, opened 2023-10-31
Update factory.rst

### LINKED ISSUE #5314: Typo in docs/tutorial/factory
I was working through the tutorial and ran into a problem getting the example code to work.

It appears that you need to put a trailing backslash on this route for it to render.

---

```python
@app.route("/hello") # no trailing backslash
def hello():
    return "Hello, World!"
```
This code returns:
not_found_error

---

```python
@app.route("/hello/") # with trailing backslash
def hello():
    return "Hello, World!"
```
returns:
hello_world

Environment:

- Python version: 3.11.2
- Flask version: 3.0.0

markalanboyd:
I was working through the tutorial and ran into a problem getting the example code to work.

It appears that you need to put a trailing backslash on this route for it to render.

---

```python
@app.route("/hello") # no trailing backslash
def hello():
    return "Hello, World!"
```
This code returns:
not_found_error

---

```python
@app.route("/hello/") # with trailing backslash
def hello():
    return "Hello, World!"
```
returns:
hello_world

Environment:

- Python version: 3.11.2
- Flask version: 3.0.0

### DIFF
--- a/docs/tutorial/factory.rst
+++ b/docs/tutorial/factory.rst
@@ -62,7 +62,7 @@ directory should be treated as a package.
             pass
 
         # a simple page that says hello
-        @app.route('/hello')
+        @app.route('/hello/')
         def hello():
             return 'Hello, World!'
 


markalanboyd:
> Tutorial is correct as is. Compare the link in the tutorial to what's in your browser, there's a difference.

I should have added that the trailing backslash gets automatically added by the browser - even if I try to delete it and press enter, it adds it back again.

That said if you’ve checked this works as is on your end, it’s probably  something I’m not seeing on mine.

DUOLabs333:
I've just tested this --- the tutorial works as expected (this is with Chrome 118). Maybe there's something with cacheing?

markalanboyd:
I figured out that it had something to do with Firefox since I couldn’t repro it in Chrome or Safari - I’m sorry for wasting your time I hadn’t thought to check for that! I would have written sooner but it didn’t seem like my comments were getting through here. Thank you 👍

davidism:
It's ok, browsers are weird. Thanks for following up.

ThiefMaster:
my guess would be that while testing things you had the route with `/hello/` and accessed `/hello` which caused a (cacheable) 301 redirect to `/hello/`. so the next access to `/hello` automatically redirected (w/o even hitting the server first) to `/hello/`

markalanboyd:
> my guess would be that while testing things you had the route with `/hello/` and accessed `/hello` which caused a (cacheable) 301 redirect to `/hello/`. so the next access to `/hello` automatically redirected (w/o even hitting the server first) to `/hello/`

Thanks for pointing that out, I will look for it in the future. It was weird because I was trying to do a force reload (cmd+shift+r) but no dice.

Oh well there goes my chance for being a 1 character contributor to a great repo! I’ll keep my eyes peeled for a misplaced comma as I work through this thanks everyone.

ThiefMaster:
browsers completely suck in terms of "developer experience" when it comes to 301 caching. there is absolutely NO way to see which redirects are cached, and the ONLY way to get rid of such a cached redirect is clearing your cache altogether

PR #5331, opened 2023-11-09
use ruff linter and formatter

davidism:


ThiefMaster:
Do we need to set the minimum supported Python version somewhere? Or does ruff automatically look at the package metadata for its pyupgrade integration?

davidism:
It looks it up from `requires-python` in `pyproject.toml`.

PR #5338, opened 2023-11-15
Changed concatenation of strings to f-strings

eukub:
Checklist:


davidism:
Thanks for looking at this. However, in general we don't accept these sorts of style PRs, as they're a bit noisy. In this case, it doesn' change the performance or correctness of anything, and isn't thorough (there are other concatenations that were not addressed).

PR #5384, opened 2024-01-18
Update mongoengine.rst

ahmetelgun:
https://github.com/MongoEngine/flask-mongoengine has not been maintained for 2 years. It doesn't work with Flask>=2.3. I forked this project and added Flask>=2.3 support.

[new project](https://github.com/ahmetelgun/flask-mongoengine2/)

### DIFF
--- a/docs/patterns/mongoengine.rst
+++ b/docs/patterns/mongoengine.rst
@@ -5,12 +5,12 @@ Using a document database like MongoDB is a common alternative to
 relational SQL databases. This pattern shows how to use
 `MongoEngine`_, a document mapper library, to integrate with MongoDB.
 
-A running MongoDB server and `Flask-MongoEngine`_ are required. ::
+A running MongoDB server and `Flask-MongoEngine2`_ are required. ::
 
-    pip install flask-mongoengine
+    pip install flask-mongoengine2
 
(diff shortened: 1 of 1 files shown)


davidism:
I really appreciate that you've taken up maintaining a popular extension that had been neglected lately. However, I don't know you or what your long term plans are for supporting your fork. I don't want to point users to a fork only to have it go unmaintained as well.

A solution to this sort of issue that we're working on is the Pallets Community Ecosystem. Projects are added to the https://github.com/pallets-eco organization, so that we can add other maintainers later on if people express interest, and can provide a central community for supporting extension maintainers. The idea is similar to https://jazzband.co/ if you're familiar with it.

Would you be willing to move your project into the pallets-eco org? I'd also like to hear a bit about your plans for maintaining the library regardless.

davidism:
Looking at the original's issue tracker, it looks like someone else has also forked the project recently. Since this has all happened in the last few weeks, I'd like to wait a bit more to see what happens to those forks and the community before recommending an alternative.

It would probably be better if the docs showed how to use mongoengine (or another mongo connector) without the extension, the same way Flask's page on SQLAlchemy mentions Flask-SQLAlchemy but does not use it: https://flask.palletsprojects.com/en/3.0.x/patterns/sqlalchemy/

PR #5423, opened 2024-02-27
Use per-release URLs in GH env UI when publishing to the PyPI

webknjaz:
This essentially, makes the UI nicer in a few places with a clickable link to the released version being presented in the web interface of GitHub.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -55,9 +55,11 @@ jobs:
           GH_TOKEN: ${{ github.token }}
   publish-pypi:
     needs: [provenance]
-    # Wait for approval before attempting to upload to PyPI. This allows reviewing the
-    # files in the draft release.
-    environment: publish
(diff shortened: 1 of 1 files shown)


davidism:
Thanks! I had been looking into how to use that feature but hasn't figured it out yet.

webknjaz:
No problem :) I wondered into that file by accident so I figured I'd send an example.
FWIW my PyPUG guide shows a simplified URL (unversioned) example but I normally render-in the version too, when it's known: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

webknjaz:
One place where there link will appear is https://github.com/pallets/flask/deployments/publish and the other is the graph on the workflow run summary pages like https://github.com/pallets/flask/actions/runs/7769383805 (in the box representing the job where the environment is applied).

PR #5444, opened 2024-03-22
avoiding potential type's mismatch

igeni:
- avoiding potential type's mismatch during concatenation

Checklist:


davidism:
The only valid type here is string, if there's something else that's a separate issue. Be sure to provide a minimal reproducible example when reporting an issue, or a test when submitting a PR.

PR #5466, opened 2024-04-22
Update views.rst

vddo:
Changed html methods from tuple to list.


ThiefMaster:
Why do you want to change this?

vddo:
So it would be like it is shown in Quickstart > HTTP Methods.

PR #5485, opened 2024-05-19
Added python type annotation and return type.

MirMohammadd:
Implements missing python type annotation.


davidism:
Huh, why wasn't this caught by mypy strict?

PR #5487, opened 2024-05-21
Edited variable names in quickstart guide

jaketanis:
Changed variable names in the rendering templates code snippets. This was to remove confusion between the differences of the html variables and the python variables.

Also discussed here: #5484


jaketanis:
This pr will require changing a reference to the 'name' html variable in the second paragraph after the template code snippet. I will commit that change as well.

PR #5490, opened 2024-05-29
Update install.rst to add full path to pyproject.toml

za4g:


davidism:
Thanks, but the paths are all (not only this one) relative so it doesn't matter where a user puts the project folder.

PR #5496, opened 2024-06-05
Provide a configuration option to control automatic option responses

pgjones:
By default Flask will provide responses to OPTIONS requests that are automatically generated. These responses list the valid methods in the response headers. Whilst this is useful, it can be frowned on by auditors hence an ability to disable it wholesale is useful.


davidism:
Does this make sense as a config option? Lately I've been tending to use init params instead of config unless it actually makes sense to change something per deployment. (And you could still change it per deployment, you'd just have to do it manually in the app factory.) It seems like you'd either want automatic options or not.

How would this interact with `Response` CORS attributes/headers, the Flask-CORS extension, or OpenAPI extensions?

pgjones:
> Does this make sense as a config option?

I would probably have different values in development and production given the information is useful in development. I do this via environment variables and the config.from_prefixed_env, so a config variable would be the easiest. I'd also argue it is more consistent with the other Flask options.

> How would this interact with Response CORS attributes/headers, the Flask-CORS extension, or OpenAPI extensions?

I think these would manage the OPTIONS responses directly anyway and this could make it easier to turn off the Flask default, e.g. [Flask-CORS](https://github.com/corydolphin/flask-cors/blob/main/flask_cors/decorator.py).

davidism (approved):

Issue #5504, opened 2024-06-20
allow setting encoding in open_resource()


liffiton:
This is a duplicate of #1740 &mdash; that may have been closed for lack of a clear rationale, however, and I'd like to suggest it again with the following reasoning.

The documentation currently gives this example for using `open_resource()`:

```python
with app.open_resource("schema.sql") as f:
    conn.executescript(f.read())
```

On Windows, however, this can fail to open a file encoded in UTF-8, which most are these days, and safer code looks like this:

```python
with app.open_resource("schema.sql", mode="rb") as f:
   conn.executescript(f.read().decode("utf-8"))  # type: ignore [attr-defined]
```

(The type comment is needed to prevent mypy from complaining about `f.read()` possibly being a string with no `.decode()` method, as it can't tell that the file was opened in 'rb' mode.)

It would be cleaner and more flexible to be able to write:

```python
with app.open_resource("schema.sql", encoding="utf-8") as f:
   conn.executescript(f.read())
```

PR #5525, opened 2024-07-10
Add unit tests to show, that encoding parameter in open_resource method is required

Lopkop:
I am creating this PR so you can see the necessity of fixing issue #5504 
Tests are marked as skipped, so it will not break.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -349,7 +349,7 @@ class TestHelpers:
             os.remove("tests/static/test.txt")
 
         with open("tests/static/test.txt", "w", encoding=encoding) as f:
-            f.write("" + "Hello World!")
+            f.write("Hello World!")
 
         with app.open_resource("static/test.txt", mode="r", encoding=encoding) as f:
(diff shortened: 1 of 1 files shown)


tests/test_helpers.py
```diff
@@ -341,6 +341,19 @@ def test_open_resource(self, mode):
...
 
+    @pytest.mark.skip(reason="this feature is not yet implemented")
+    @pytest.mark.parametrize("encoding", ("utf-8", "utf-16-le"))
+    def test_open_resource_with_encoding(self, encoding):
+        app = flask.Flask(__name__)
+        if os.path.isfile("tests/static/test.txt"):
+            os.remove("tests/static/test.txt")
+
+        with open("tests/static/test.txt", "w", encoding=encoding) as f:
+            f.write("" + "Hello World!")
```
ThiefMaster, line 352: Is that an empty string or something non-printable/zero-width? If it's the latter, please use an escape sequence instead so it's clear it's not something useless.

PR #5526, opened 2024-07-10
Add encoding parameter to open resource

### LINKED ISSUE #5504: allow setting encoding in open_resource()

Lopkop:
- Add encoding parameter to open_resource method and pass it to built-in open function
- Remove skip mark from unittests, that checks this functionality

This PR fixes #5504

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,9 @@ Version 3.1.0
 
 -   Provide a configuration option to control automatic option
     responses. :pr:`5496`
-
+-   ``Flask.open_resource``/``open_instance_resource`` and
+    ``Blueprint.open_resource`` take an ``encoding`` parameter to use when
+    opening in text mode. It defaults to ``utf-8``. :issue:`5504`
 
 Version 3.0.3
 -------------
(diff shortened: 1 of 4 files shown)


davidism:
Thanks for getting this started. I've made the change to `open_instance_resource` and `Blueprint.open_resource` as well; added version markers, added a change log, and refactored the docs and tests a bit.

PR #5532, opened 2024-07-19
Provide more clear flask -e, --env-file flag desription

differencialx:
Currently the description for `flask` cli command `-e, --env-file` flag is misleading. I understand that this functionality is covered and perfectly explained in docs, but in case you are not familiar with documentation you will just use `flask --help` command to read which flags are allowed.

This PR just adds few sentences to make `-e, --env-file` flag description more clear and exhaustive.
**Before:**
image
**After**
image

Initial PR with for `-e, --env-file` flag: https://github.com/pallets/flask/pull/4646

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Released 2024-04-07
     :issue:`5448`
 -   Don't initialize the ``cli`` attribute in the sansio scaffold, but rather in
     the ``Flask`` concrete class. :pr:`5270`
-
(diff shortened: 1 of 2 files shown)


davidism:
I forgot that `load_dotenv` does not override existing env vars, so loading multiple files can't override "less specific/secret" files. I feel like there has to be a way to fix this, so from most to least preferred is: os.environ > custom > .flaskenv > .env. Otherwise, you'd never be able to override something in `.flaskenv` with your private `.env` or custom file. 🤔

davidism:
Closing in favor of fixing the precedence, see #5628

Issue #5537, opened 2024-07-31
`pyright` type checking seems to be unused. `mypy` still being used.


pygeek:
We added `pyright` in 3.0.3 (see: https://github.com/pallets/flask/pull/5457), but `mypy` is still being used for type checking https://github.com/pallets/flask/blob/main/tox.ini#L29-L32, and `pyright` is not.

If this was an intentional change of type checkers, then would expect that `mypy` be completely removed, and `pyright` be the command run in `tox typing`.

It should be noted, however, that running `pyright` yields 90 errors (71 tests/19 src). Is there an upgrade path or was this an oversight?

Environment:

- Python version:
- 3.12
- Flask version:
- 3.0.3

davidism:
fixed in #5620

PR #5552, opened 2024-08-10
git add tox.ini requirements/typing.txt

ComputerScienceGirl12:
git commit -m "Replace mypy with pyright in tox.ini for consistent type checking"

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,8 +27,8 @@ skip_install = true
 commands = pre-commit run --all-files
 
 [testenv:typing]
-deps = -r requirements/typing.txt
-commands = mypy
+deps = pyright
+commands = pyright
 
 [testenv:docs]
 deps = -r requirements/docs.txt


ComputerScienceGirl12:
This PR addresses [Issue #5537](https://github.com/pallets/flask/issues/5537).

Changes:
- Updated `tox.ini` to replace `mypy` with `pyright` for type checking.
- Modified `requirements/typing.txt` to include `pyright`.
- Ran `pyright`, which yielded no errors.

This change ensures consistency in type checking tools within the Flask project, aligning with the addition of `pyright` in version 3.0.3.

davidism:
This isn't correct. We want to run both tools. We want to install them from the lock file, not directly. I think we actually had it set up before but then it got removed for some reason. You can see the other Pallets repos for how it's set up there. However, just enabling it isn't sufficient either, you need to actually get the tests passing.

PR #5601, opened 2024-09-09
Updated deploy.rst for better clarity

najasnake12:
I changed deploy.rst for better clarity specifically this part:
Before:
$ python -c 'import secrets; print(secrets.token_hex())'

    '192b9bdd22ab9ed4d12e236c78afcb9a393ec15f71bbf5dc987d54727823bcbf'

After:
$ python -c 'import secrets; print(secrets.token_hex())'

    Example secret key:
    '192b9bdd22ab9ed4d12e236c78afcb9a393ec15f71bbf5dc987d54727823bcbf'

### DIFF
--- a/docs/tutorial/deploy.rst
+++ b/docs/tutorial/deploy.rst
@@ -62,6 +62,7 @@ You can use the following command to output a random secret key:
 
     $ python -c 'import secrets; print(secrets.token_hex())'
 
+    Example secret key:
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I wonder whether we shouldn't simply blacklist that particular not-so-secret key, since I guess there's always a chance that people copy it instead of generating their own, even with that line message...

Also I don't think that adding this message is a good idea - this code block shows a shell command and its output, and that message you added is not part of the output.

PR #5610, opened 2024-09-29
helpers.py

Kartikeya-Kotkar:
The type: ignore[return-value] was used to suppress a return type mismatch that likely no longer exists. Since mypy is now flagging it as unused, the directive can be safely removed.


davidism:
Ended up fixing this directly while updating dependencies.

PR #5611, opened 2024-10-10
Update README.md

Vrusharth:


ThiefMaster:
This is spam. Are you trying to get Hacktoberfest PRs? If yes, stop creating such spam PRs immediately as it will just get you banned from that event. IF you're just learning Git, then do it on your own repos and not those of other people.

PR #5613, opened 2024-10-16
Address a mypy failure in CI

kurtmckee:
mypy 1.11.1 is reporting that [`return-value` is an unused ignore in CI](https://github.com/pallets/flask/actions/runs/10655465924/job/29533008516#step:6:25), and when running `tox -e typing` locally.

As this is a code comment for mypy's benefit, I'm anticipating that this doesn't need to be added to the CHANGELOG, but please let me know if that's incorrect. Thanks!


davidism:
Ended up fixing this directly while updating dependencies.

Issue #5628, opened 2024-11-06
dotenv file arg should take precedence over default dotenv files


davidism:
As pointed out in #5532, `-e file` is loaded after the default `.env` and `.flaskenv` files, which means it won't override env vars that are set by the defaults. It will be a little tricky to track during CLI startup, but I think it should be possible to fix this so that precedence is: `os.environ` > `-e path` > `.env` > `.flaskenv`.

jalaj711:
Hey, can I work on this issue?

PR #5629, opened 2024-11-07
give priority to '-e/--env-file' option over default files (#5628)

jalaj711:
This PR fixes issue #5628 . The issue arose from the fact that `FlaskGroup().make_context()` was loading the default env files before loading the arguments (by calling `super().make_context()`. Simply reversing the order of these actions fixes the issue. I have added the relevant tests to make sure it works.

- [x] Add tests that demonstrate the correct behavior of the change. Tests
  should fail without the change.
- [x] Add an entry in CHANGES.rst summarizing the change and linking to the issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@ Unreleased
     about resource limits to the security page. :issue:`5625`
 -   Add support for the ``Partitioned`` cookie attribute (CHIPS), with the
     ``SESSION_COOKIE_PARTITIONED`` config. :issue`5472`
+-   Give priority to file specified through the -e/--env-file option
+    over default .env & .flaskenv files in CLI.
 
 
 Version 3.0.3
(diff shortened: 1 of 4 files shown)


jalaj711:
The failing test seems to be failing in other actions before this PR as well: https://github.com/pallets/flask/actions/runs/11709066697/job/32612272783

davidism:
I don't think this is going to work. The env vars need to be loaded before creating the context, as the processing pipeline needs to be able to see the env vars to use their values for the parameters.

davidism:
`override` is also not correct, as `os.environ` should still take precedence.

jalaj711:
@davidism Thank you for your feedback, I have changed that approach. The tests are still passing as expected. Can you please take a look at this now?

davidism:
I'm also already working on this using a different approach, sorry for the confusion.

jalaj711:
Okay, I am kind of stuck at this point. Can you provide any ideas about how this might be implemented?

davidism:
Closing in favor of #5630

PR #5643, opened 2024-11-17
Use SPDX license identifier

cdce8p:
[PEP 639](https://peps.python.org/pep-0639/) recommends the use of SPDX license identifiers.

At the moment the `license` key is ignored by `flit`, see https://github.com/pypa/flit/issues/525. However, it's still possible to replace it now. Once the support in flit is added, the value will be used automatically.

_At some point, the table will need be be changed to just a string for full PEP 639 compliance. That requires support for it first though. The change is also pretty simple._
```diff
-license = {text = "BSD-3-Clause"}
+license = "BSD-3-Clause"
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ include = [
     "tests/",
     "CHANGES.rst",
     "CONTRIBUTING.rst",
+    "LICENSE.txt",
     "tox.ini",
 ]
 exclude = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "Flask"
 version = "3.1.0"
 description = "A simple framework for building complex web applications."
 readme = "README.md"
-license = {file = "LICENSE.txt"}
+license = {text = "BSD-3-Clause"}
 maintainers = [{name = "Pallets", email = "contact@palletsprojects.com"}]
 classifiers = [
     "Development Status :: 5 - Production/Stable",


cdce8p:
Working on adding support in flit now. Marking it as draft until then.

davidism:
I moved to `license = {file = "LICENSE.txt"}` because when I re-read the current spec months ago, it seemed like that was "more correct" until SPDX was supported. `license = {text = "BSD-3-Clause"}` isn't technically valid (it specifies the full text of the license). I'm fairly sure we need to hold off until we can specify both `license = "BSD-3-Clause` _and_ `license_files = ["LICENSE.txt]" so that both the SPDX and file are present.

cdce8p:
> I moved to `license = {file = "LICENSE.txt"}` because when I re-read the current spec months ago, it seemed like that was "more correct" until SPDX was supported.

Yeah, that's the current spelling in the docs. I'd just argue that it's probably outdated. There is an open issue to rework that section of the packaging docs. https://github.com/pypa/packaging.python.org/issues/1245
From my experience, it's actually better to use to existing `license` field to in a way "back-populate" the license expression until that is fully supported

> `license = {text = "BSD-3-Clause"}` isn't technically valid (it specifies the full text of the license).

Just from a metadata perspective, it's fully valid. At least setuptools, poetry, and hatch support it.

> I'm fairly sure we need to hold off until we can specify both `license = "BSD-3-Clause` _and_ `license_files = ["LICENSE.txt]" so that both the SPDX and file are present.

While working on a PR for flit, I noticed that all `project.license.file` really does atm is include the license file in the sdist. For the wheel a glob `('COPYING*', 'LICENSE*')` is used to put in `.dist-info`. Besides that it should be `.dist-info/licenses`, the current behavior can also be archived by adding `LICENSE.txt` to `[tool.flit.sdist.include]`.

davidism:
I'm going to close this for now. Once Flit supports `license = "string"` and `license_files = [...]`, you're welcome to create a PR for that.

cdce8p:
> Once Flit supports `license = "string"` and `license_files = [...]`, you're welcome to create a PR for that.

Sound good! I've been working on adding the support for it in `flit`. Hopefully it gets reviewed soon.
- https://github.com/pypa/flit/pull/705
- https://github.com/pypa/flit/pull/707

PR #5664, opened 2025-01-01
Update LICENSE.txt, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright 2010 Pallets
+Copyright 2025 Pallets
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are


davidism:
> yearly

Looks like you didn't stop to consider that the current text is not 2023, so this change _isn't_ yearly.

Copyright/license, if it contains a year at all, reflects the year something was _created_, not the last year it was worked on. Please don't make a bunch of PRs like this on different projects, it's just noise and not helpful.

webknjaz:
spam xref https://github.com/pytest-dev/pytest/pull/13097#issuecomment-2566991642 / https://github.com/community/maintainers/discussions/442#discussioncomment-11710168

davidism:
Yeah, I also noticed that and reported it to GitHub, forgot to update here.

webknjaz:
@davidism 

Oh, I didn't realize you weren't there. It's a new location of a private community by GH. Kinda.

It used to be possible to join via https://maintainers.github.com, but I'm not sure if there are any limitations now. In the past, it was invite-only where current members could nominate new people, but that was paused at some point.

PR #5665, opened 2025-01-02
Workflow

pgjones:

### DIFF
--- a/src/flask/sansio/sessions.py
+++ b/src/flask/sansio/sessions.py
@@ -1,0 +1,215 @@
+from __future__ import annotations
+
+import collections.abc as c
+import typing as t
+from abc import ABCMeta
+from collections.abc import MutableMapping
+from datetime import datetime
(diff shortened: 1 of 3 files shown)


davidism:
It doesn't seem to cut down on much boilerplate, there's still approximately the same amount of workflow code to keep synced across projects. Is there more than can be done?

pgjones:
Yea, I can also change the pre-commit and publish workflows to use shared actions. If we like this pattern?

pgjones:
Added pre-commit and build actions - will clean up and fix if we like this pattern.

src/flask/sansio/sessions.py
```diff
@@ -0,0 +1,215 @@
+from __future__ import annotations
```
pamelafox, line 1: This stuff isnt meant to be in the PR, right?

samuelpondugala (commented):
.

PR #5682, opened 2025-02-19
Use SPDX license identifier

cdce8p:
Followup to #5643

Flit 3.11 was released today with basic support for PEP 639.
https://flit.pypa.io/en/stable/history.html#version-3-11

_If accepted, I'm happy to redo the other PRs as well._

/CC: @davidism

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,14 +3,14 @@ name = "Flask"
 version = "3.2.0.dev"
 description = "A simple framework for building complex web applications."
 readme = "README.md"
-license = {file = "LICENSE.txt"}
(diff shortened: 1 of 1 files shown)


davidism:
I'm dealing with more sweeping changes than this to our build system, so I'll probably just do this when I have time for all that.

cdce8p:
> I'm dealing with more sweeping changes than this to our build system, so I'll probably just do this when I have time for all that.

Fine for me. Although with support added in `flit`, the new keys can be used with other build backends as well. _Setuptools will hopefully follow next week._

Should I close this PR then?

PR #5705, opened 2025-03-30
Merge app and request contexts into a single context (#5639)

### LINKED ISSUE #5639: merge app and request contexts into a single context
Right now we have two separate contexts managed separately, the app and request contexts. This makes the implementation pretty complicated, as we need to maintain two context var stacks, and do a bunch of checks in the request context to make sure we're managing the correct app context. It makes an already confusing topic more complicated to explain: app context is active for requests and cli commands, don't push an app context before making a request, etc.

I think merging the two contexts could be possible. The single context (`ExecutionContext`?) would have the `g`, `request`, and `session` attributes, but accessing `request` or `session` when not in a request would raise an error.

ascender1729:
# Merge app and request contexts into a single context (#5639)

This PR implements the proposal from #5639 to merge the app and request contexts into a single execution context. This simplifies the implementation by:

1. Creating a new `ExecutionContext` class that combines both app and request contexts
2. Using a single context variable `_cv_execution` instead of separate `_cv_app` and `_cv_request`
3. Maintaining backward compatibility by setting both old context vars when needed
4. Adding deprecation warnings for the old context methods
5. Updating all globals to use the new execution context

Key changes:

- Added new `ExecutionContext` class in `ctx.py` that handles both app and request contexts
- Updated `globals.py` to use the new execution context for all proxies
- Modified `app.py` to create execution contexts instead of separate app/request contexts
- Added deprecation warnings for `has_app_context()` and `has_request_context()`
- Simplified context management by removing the need to track multiple context stacks

The changes maintain backward compatibility while paving the way for a cleaner implementation in Flask 4.0. The new execution context:

- Has a single context variable instead of two
- Handles both app and request state in one place
- Maintains the same behavior for existing code
- Provides clearer error messages when contexts are missing

This change will make the codebase simpler and easier to understand, while also making it more efficient by reducing the number of context variables that need to be managed.

Closes #5639

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -28,10 +28,8 @@ from werkzeug.wsgi import get_host
 
 from . import cli
 from . import typing as ft
-from .ctx import AppContext
-from .ctx import RequestContext
-from .globals import _cv_app
-from .globals import _cv_request
+from .ctx import ExecutionContext
+from .globals import _cv_execution
 from .globals import current_app
 from .globals import g
 from .globals import request
@@ -1057,7 +1055,7 @@ class Flask(App):
         .. versionadded:: 2.2
             Moved from ``flask.url_for``, which calls this method.
         """
-        req_ctx = _cv_request.get(None)
+        req_ctx = _cv_execution.get(None)
 
         if req_ctx is not None:
             url_adapter = req_ctx.url_adapter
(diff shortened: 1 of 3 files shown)


davidism:
Sorry if this was unclear, but the issue was assigned to me because I did no want others working on it. It's extremely complex and will require a lot of thought on compatibility and deprecation. At a glance, this PR has numerous issues, and it's not something I want to shepherd through review. It also seems like it was written at least in part by LLM tools, contributing to the quality issues.

PR #5732, opened 2025-05-14
Dev0

Wiczka03:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,1 @@
+Dodano tę linię z terminala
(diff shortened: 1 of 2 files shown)


ThiefMaster:
We received 2 such weird PRs within a few minutes. Is this part of some class? If yes, please forward the message to your teacher/professor that it is not OK to direct such noise towards Open Source maintainers.

It's great to get people into Open Source and Git/GitHub. But testing should always be done on dedicated repos, not forks of real ones (where it's very easy to send PRs upstream).

PR #5741, opened 2025-05-21
Docs: Fix escaping in HTML escaping example

Badhreesh:
The code in the [Quickstart's HTML Escaping section](https://flask.palletsprojects.com/en/stable/quickstart/#html-escaping) does not render name=`<script>alert("bad")</script>` to text because the slash in `</script>` gets interpreted as a trailing slash, leading to a 404.

Adding the `path` type converter to the `name` variable allows:
1.  the script tag to be rendered as text if HTML escaping is done
2.  the alert in the script tag to be executed if HTML escaping is not done.

Since its a simple one line change for code in the documentation, I wasn't sure if I needed to raise an issue regarding this first. I'd be happy to do so if its required.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,7 +19,6 @@ jobs:
           - {python: '3.12'}
           - {python: '3.11'}
           - {python: '3.10'}
-          - {python: '3.9'}
           - {name: PyPy, python: 'pypy-3.11', tox: pypy3.11}
           - {name: Minimum Versions, python: '3.13', tox: tests-min}
           - {name: Development Versions, python: '3.10', tox: tests-dev}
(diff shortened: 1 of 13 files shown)


davidism:
Please rebase the branch to stable, then retarget the PR to stable.

Badhreesh:
I realized that I made the PR on the wrong branch, when I should've done it for the `stable` branch. I'll close this PR and make a new one pointing to the correct branch. Apologies.

ThiefMaster:
FYI you can change the target branch in a PR via the edit button on top. You'd of course need to use the git client to rebase your branch as well in that case.

Badhreesh:
Thanks for the info. I tried rebasing but it introduced way more changes than my simple one line change into the PR. I've not done a rebase before and so I thought I must have done something wrong. I've made a [new PR](https://github.com/pallets/flask/pull/5742) pointing to the stable branch now.

PR #5742, opened 2025-05-21
Docs: Fix escaping in HTML escaping example

Badhreesh:
The code in the [Quickstart's HTML Escaping section](https://flask.palletsprojects.com/en/stable/quickstart/#html-escaping) does not render name=<script>alert("bad")</script> to text because the slash in </script> gets interpreted as a trailing slash, leading to a 404.

Adding the path type converter to the name variable allows:
1. the script tag to be rendered as text if HTML escaping is done
2. the alert in the script tag to be executed if HTML escaping is not done.

Since its a simple one line change for code in the documentation, I wasn't sure if I needed to raise an issue regarding this first. I'd be happy to do so if its required.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -141,7 +141,7 @@ how you're using untrusted data.
 
     from markupsafe import escape
 
-    @app.route("/<name>")
+    @app.route("/<path:name>")
     def hello(name):
         return f"Hello, {escape(name)}!"
 


Badhreesh:
@davidism, I realized that for docs changes I should've made the PR from the `stable` branch. I'll closed the previous PR and make a new one pointing to the correct branch.

Badhreesh:
> I wonder if there's a way to demonstrate the issue using the existing url, rather than changing to the `path` type, which doesn't really make sense in this context. Either using something like `` that doesn't require a closing tag, or taking the argument from `request.args`?

I agree. Since the emphasis on escaping is to prevent the execution of scripts, I went ahead with using `request.args` to ensure correct demonstration of the issue.

Badhreesh:
@davidism, is there anything else I need to do to have the PR reviewed?

PR #5761, opened 2025-06-17
fix: template_filter function ( function parameter was overridden by 'name')

p-r-a-v-i-n:
with ref #5729 
This issue was happening bcs in template_filter decorator "name" argument was overriding default "function" argument by which decorator never receives the actual function lets.
bcs this the function also not being set in filter too.

I have added the function argument in decorator now .

Now no need to called with parenthesis it work as standard decorator now.


davidism:
Why did you open this despite the last comment on the linked issue and the multiple existing open PRs?

PR #5777, opened 2025-07-16
use IO[bytes] instead of BinaryIO for wider compatibility

tvuotila:
`flask.helpers.send_file` was the only place where `t.BinaryIO` was used. I changed it to be more loose type `t.IO[bytes]`.

fixes #5776


davidism:
Since this is a bug fix, please rebase and target it at stable rather than main.

tvuotila:
> Since this is a bug fix, please rebase and target it at stable rather than main.

Done.

PR #5779, opened 2025-07-23
Add devcontainer with Python 3.13 and uv support

hola-kundan:
This PR adds a development container setup under `.devcontainer/` using Microsoft's Python 3.13 image, with uv support for dependency management.

Changes include:
- Custom Dockerfile using `mcr.microsoft.com/devcontainers/python:3.13`
- Switching to uv for editable installs and dev dependencies
- Config updates in `devcontainer.json` and `on-create-command.sh`

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,9 @@
+# Base image: Match with .readthedocs.yaml and pyproject.toml
+FROM mcr.microsoft.com/devcontainers/python:3.13
+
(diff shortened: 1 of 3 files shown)


hola-kundan:
Hi @davidism  — noticed the PR was closed without merge. Just wanted to check if I missed anything or if there’s a different direction preferred for devcontainers.

Since Flask seems to be adopting uv, I thought aligning the devcontainer with that would be helpful. Happy to make changes if needed.

PR #5790, opened 2025-08-01
docs: Fix grammar and article usage in documentation

geekalaa:
- Fix verb agreement in lifecycle.rst: 'creates and send'
   -> 'creates and sends'
  - Fix article usage: 'a HTTP' -> 'an HTTP' in quickstart.rst and errorhandling.rst


davidism:
Thanks, however "an HTTP" vs "a HTTP" is arguable either way, and we typically don't accept random minor typo fixes.

PR #5964, opened 2026-03-27
Handle missing secret key during session save

pefia:
I noticed a real crash path in session saving: `SecureCookieSessionInterface.save_session` called `dumps()` on `get_signing_serializer(app)` without checking whether the serializer existed.

I validated the behavior locally with a minimal reproduction where a session object was present, then `app.secret_key` became falsy before save. In that flow, Flask raised an internal `AttributeError` (`'NoneType' object has no attribute 'dumps'`) instead of the normal session configuration error.

Then I fixed the code in a streamlined way:

- I added a guard in `save_session` to handle a missing serializer explicitly.
- I raised the same user-facing `RuntimeError` message Flask already uses for missing secret key session failures.
- I removed the `type: ignore[union-attr]` usage by handling the union properly.
- I deduplicated the error message by using a shared constant for both `NullSession` and `save_session`.

After that, I added a focused regression test that reproduces the exact scenario:

- Start with a valid secret key.
- Create a modified session.
- Clear the secret key.
- Call `save_session` and assert Flask raises the expected runtime error message.

I also made sure the patch stayed minimal and clean:

- Only two files are changed:
  - `src/flask/sessions.py`
  - `tests/test_session_interface.py`
- No scratch files or unrelated edits remain.

Validation I ran:

- Direct repro before/after behavior check (crash replaced with expected runtime error).
- Focused test run:
  - `pytest tests/test_session_interface.py -q`
  - Result: `2 passed`

### DIFF
--- a/src/flask/sessions.py
+++ b/src/flask/sessions.py
@@ -13,6 +13,12 @@ from werkzeug.datastructures import CallbackDict
 
 from .json.tag import TaggedJSONSerializer
 
+_missing_secret_key_err_msg = (
+    "The session is unavailable because no secret "
+    "key was set.  Set the secret_key on the "
+    "application to something unique and secret."
+)
+
 if t.TYPE_CHECKING:  # pragma: no cover
     import typing_extensions as te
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
This feels AI-generated...

`app.secret_key` becoming falsy is not something that makes any sense and would be a bug in the application doing it. Causing an error in Flask code in such a case doesn't feel like a problem to me.

PR #5992, opened 2026-04-10
<spam>

AhsanSheraz:
Invalid AI security slop


davidism:
This is an example project, the requirements are just there to have something reproducibly installable.

PR #6003, opened 2026-04-19
AI junk

sahilmalhotra1987:
## Summary

- Adds a `--json` flag to `flask routes` that outputs the route table as a JSON array instead of the text table
- Each entry has `endpoint`, `methods` (sorted list), and `rule`; subdomain apps also emit `subdomain`, host-matching apps emit `host`
- `HEAD` and `OPTIONS` are filtered unless `--all-methods` is also passed; empty apps produce `[]`
- The `--sort` flag is honoured in JSON mode (`match` preserves natural `iter_rules()` order)
- Fully backwards-compatible — all existing `flask routes` invocations are unaffected

## Test plan

- [ ] `test_json_output` — basic shape: keys `endpoint`, `methods`, `rule` present; methods is a list
- [ ] `test_json_no_routes` — empty app returns `[]`
- [ ] `test_json_excludes_head_options_by_default` — HEAD/OPTIONS absent without `--all-methods`
- [ ] `test_json_all_methods` — HEAD/OPTIONS present with `--all-methods`
- [ ] `test_json_with_subdomain` — `subdomain` key present and correct value
- [ ] `test_json_with_host_matching` — `host` key present and correct value
- [ ] `test_json_sort` (parametrized) — `endpoint` and `rule` sort keys produce sorted output
- [ ] `test_json_sort_match_preserves_iter_rules_order` — `--sort match` preserves natural route order

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@ Version 3.2.0
 
 Unreleased
 
+-   The ``flask routes`` command has a ``--json`` flag to output the route
+    table as a JSON array, suitable for scripting and tooling.
 -   Drop support for Python 3.9. :pr:`5730`
 -   Remove previously deprecated code: ``__version__``. :pr:`5648`
 -   ``RequestContext`` has merged with ``AppContext``. ``RequestContext`` is now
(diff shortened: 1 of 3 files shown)


ThiefMaster:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

No thanks. If we wanted AI-generated code, we could just use these tools ourselves.

PR #6010, opened 2026-04-25
AI junk

Devansh-66:


ThiefMaster:
Considering that you added some strange shell script to run tests and not just the actual test, I assume you used an LLM to generate this PR. We are not interested in such contributions.

PR #6017, opened 2026-05-08
Improve Git Bash virtual environment activation docsImprove Git Bash virtual environment activation docs

HarsshA2660:
Adds a note for Git Bash users on Windows when activating
a virtual environment.

This helps avoid activation command issues during setup.


ThiefMaster:
I think it's clear that bash-like environments use the linux version...

PR #6018, opened 2026-05-10
AI junk

zinc-builds:
## Summary

Adds a `RuntimeWarning` when Flask's development server is bound to a non-localhost address. The Werkzeug debugger allows arbitrary code execution when exposed to the network, yet many tutorials and quick-start guides instruct users to bind to `0.0.0.0` without warning them of the risks.

## Security Impact

Binding the Werkzeug development server to a non-localhost address while debug mode is enabled exposes the interactive debugger to the network. The debugger's PIN-protected console can execute arbitrary Python code. This change warns users when:

1. The host is set to a non-local address (anything other than `127.0.0.1`, `localhost`, or `::1`)
2. Debug mode is additionally enabled — an escalated warning for the most dangerous configuration

## Implementation

- Uses `warnings.warn()` with `RuntimeWarning` so the warning is visible but doesn't prevent the server from starting
- Uses `stacklevel=2` so the warning points to the user's `app.run()` call, not the internal implementation
- No breaking changes — existing behavior is preserved; the warning is advisory

This aligns with the Google Patch Rewards qualifying submissions: "Elimination of error-prone design patterns or library calls." Binding a debug-capable server to 0.0.0.0 without warnings is a well-known footgun that has led to real-world compromises.

`kumquat`

## Checklist

- [x] No breaking changes to existing behavior
- [x] Uses standard library `warnings` module (no new dependencies)
- [x] Warning points to correct caller location via `stacklevel`
- [x] Compatible with all Python versions Flask supports

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -14,6 +14,7 @@ from types import TracebackType
 from urllib.parse import quote as _url_quote
 
 import click
+import warnings
 from werkzeug.datastructures import Headers
 from werkzeug.datastructures import ImmutableDict
 from werkzeug.exceptions import BadRequestKeyError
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Not needed IMHO, and I have a strong feeling that this PR is LLM-generated.

davidism:
Have you ever run the dev server before? It shows a warning always, no need for another one. Do not submit LLM junk to projects.

PR #6034, opened 2026-05-24
AI junk

JosefVacha:
## Summary
Replaces print() with logger for better logging.

## Verification
- GPG signed commit


ThiefMaster:
I'm not sure if this is AI junk or human-written junk, but it's junk... Did you even LOOK at the changes "you" made?!

PR #6036, opened 2026-05-25
edited 'a' to 'an' before 'SQL'

jaketanis:
Since 'SQL' makes a vowel sound, it should be 'an' instead of 'a'.


jaketanis:
I was reading through the docs and this stood out to me, but I realize that SQL also has the "sequel" pronunciation, which is very common. So this pr didn't really do anything valuable for the docs. I'll be more mindful in the future before submitting a pr.

@davidism, Is there anything else I should keep in mind for helpful documentation pull requests that I may have missed?

PR #6037, opened 2026-05-26
AI junk

Architect8989:
### Problem
`from flask.cli import CertParamType` raises `NameError: name 'ssl' is not defined` because `ssl` is only imported under `TYPE_CHECKING`, but referenced in the class signature:
```python
class CertParamType(click.ParamType[str | os.PathLike[str] | ssl.SSLContext]):
```

### Root Cause
Even with `from __future__ import annotations`, `click.ParamType` is not subscriptable. The subscript causes eager evaluation of the type arguments at class definition time.

### Fix
1. Move `import ssl` out of the `if t.TYPE_CHECKING:` block to the top-level imports
2. Remove subscript from `click.ParamType[...]` since ParamType is not generic

### Verification
- All 491 tests pass
- `from flask import Flask` imports without error

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -24,9 +24,9 @@ from .globals import current_app
 from .helpers import get_debug_flag
 from .helpers import get_load_dotenv
 
-if t.TYPE_CHECKING:
-    import ssl
+import ssl
 
+if t.TYPE_CHECKING:
     from _typeshed.wsgi import StartResponse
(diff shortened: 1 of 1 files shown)


jwag956:
Hi David... I am actually getting the exact same error running Flask-Security tests against latest main Flask...
Using python 3.13...

davidism:
Ok, please report an issue and if you want open a human written pr.

PR #6039, opened 2026-05-31
AI spam

jiyujie2006:
## Summary

Avoid subscripted `click.ParamType[...]` at runtime for the `flask run --cert` option type, while preserving the generic base for static type checking.

## Root cause

`CertParamType` inherited from `click.ParamType[str | os.PathLike[str] | ssl.SSLContext]`. That expression is evaluated when `flask.cli` is imported. With Flask's minimum Click version, `ParamType` is not subscriptable at runtime; with newer Click, the expression also requires `ssl` at runtime. Both cases can break `import flask` before any CLI command runs.

## Checks

- `uv run python -c "import flask"`
- `uv run pytest tests/test_cli.py -q`
- `uv run ruff check src/flask/cli.py`
- `uv run --locked --no-default-groups --group dev tox run -e tests-min -- tests/test_cli.py -q`
- `uv run --locked --no-default-groups --group dev tox run -e typing`

### DIFF
--- a/src/flask/cli.py
+++ b/src/flask/cli.py
@@ -7,6 +7,7 @@ import inspect
 import os
 import platform
 import re
+import ssl
 import sys
 import traceback
 import typing as t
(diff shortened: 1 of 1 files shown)


jiyujie2006:
CI note: the remaining `Development Versions` failure appears unrelated to this PR. It is failing during test collection because Werkzeug main emits deprecation warnings for `parse_cache_control_header` and `parse_set_header`, while Flask treats warnings as errors. The relevant checks for this change are passing now, including normal Python/OS jobs, `Minimum Versions`, `typing`, pre-commit, docs, and the local targeted `tests-min` run.

PR #6062, opened 2026-06-22
Add request id response header

AdviSriv:


ThiefMaster:
You can do that in your own code, this does not belong in the flask core.

(also, general advice, don't open incomplete PRs)

PR #6064, opened 2026-06-23
Add query() route shortcut and MethodView support for HTTP QUERY

### LINKED ISSUE #6065: Add a query() route shortcut and MethodView support for HTTP QUERY (RFC 10008)
The HTTP `QUERY` method was standardized in
[RFC 10008](https://www.rfc-editor.org/info/rfc10008/) (June 2026): a safe,
idempotent request method that carries content in the request body (the
standardized replacement for the "use POST for a read-only query with a large
body" workaround in GraphQL / REST search APIs).

I'd like Flask to treat `QUERY` as a first-class method like the others:

- a `query()` route shortcut on `Scaffold` (inherited by `Flask` and
  `Blueprint`), mirroring `get`/`post`/`put`/`delete`/`patch`;
- `MethodView` dispatch of `QUERY` to a `query` handler (add `"query"` to
  `views.http_method_funcs`).

---

Is this solvable without changes to Flask? Partly — `@app.route("/x",
methods=["QUERY"])` already works today, since Flask doesn't restrict method
names and Werkzeug parses the body. What's missing is the ergonomic shortcut and
`MethodView` dispatch that every other common method already has, so this is a
consistency/ergonomics gap rather than new capability — the same rationale behind
the existing `get`/`post`/... shortcuts. Use case: safe, idempotent search
endpoints that need a request body too large or complex for the URL.

aliunsal:
## Description

[RFC 10008](https://www.rfc-editor.org/info/rfc10008/) (June 2026) standardizes
the `QUERY` HTTP method: a safe, idempotent request method that carries content
in the request body (standardizing the "POST for a read-only query with a large
body" workaround used by GraphQL and REST search endpoints).

Flask already routes and dispatches arbitrary method names, so this adds the
idiomatic ergonomics that mirror the existing
`get`/`post`/`put`/`delete`/`patch` shortcuts:

- `Scaffold.query()` registers a route with `methods=["QUERY"]`, inherited by
  both `Flask` and `Blueprint`.
- `"query"` is added to `views.http_method_funcs`, so `MethodView` dispatches a
  `QUERY` request to a `query` handler.

Higher-level RFC 10008 features (the `Accept-Query` header, server-assigned
query-result URIs, caching semantics) are application/Werkzeug concerns and are
intentionally out of scope.

## Related

fixes #6065

## Checklist

- [x] Tests added that fail without the change (`test_query_method_route`,
      `test_query_method_view`).
- [x] Code docs updated (`query()` docstring with `.. versionadded:: 3.2`); the
      method is surfaced via the `Flask`/`Blueprint` autoclass, so no manual
      `docs/` change is needed.
- [x] `CHANGES.rst` entry added, linking to the issue.
- [x] `.. versionadded::` added (new method, so `versionadded` rather than
      `versionchanged`).

---

Note: tests use `client.open(method="QUERY")` rather than `client.query()` so
they pass against the released Werkzeug. A `Client.query()` shortcut is a
separate, independent Werkzeug change (pallets/werkzeug#3194).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ Unreleased
     it's disabled in config. Previously, only disabling worked. :issue:`5916`
 -   ``Flask.select_jinja_autoescape`` uses case-insensitive comparison instead
     of only lower case file extensions. :pr:`6012`
+-   Add ``query`` method route shortcut for the ``QUERY`` HTTP method
+    (:rfc:`10008`). ``MethodView`` dispatches ``QUERY`` requests to a ``query``
+    handler. :issue:`XXXX`
 
 
 Version 3.1.3
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ Unreleased
     of only lower case file extensions. :pr:`6012`
 -   Add ``query`` method route shortcut for the ``QUERY`` HTTP method
     (:rfc:`10008`). ``MethodView`` dispatches ``QUERY`` requests to a ``query``
-    handler. :issue:`XXXX`
+    handler. :issue:`6065`
 
 
 Version 3.1.3
--- a/src/flask/sansio/scaffold.py
+++ b/src/flask/sansio/scaffold.py
@@ -333,6 +333,17 @@ class Scaffold:
         return self._method_route("PATCH", rule, options)
 
     @setupmethod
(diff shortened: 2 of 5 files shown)


aliunsal:
Superseded by #6066, which opens the same change from a topic branch (`feat/http-query-method`) instead of `main`.

PR #6069, opened 2026-07-01
ignore jinja suffix when selecting autoescape

arib06:
strip a trailing .jinja/.jinja2/.j2 suffix in select_jinja_autoescape before matching the extension, otherwise a template using the common Jinja naming convention like page.html.jinja or icon.svg.j2 renders without HTML escaping while a plain notes.txt.j2 stays unescaped.


arib06:
Was this closed on purpose? The extension check still misses templates using the standard Jinja suffix (page.html.jinja, icon.svg.j2, data.xml.jinja2), so those render with autoescaping off while the equivalent .html/.svg name gets escaped. Seemed like a natural follow-up to the case-insensitive change in #6012. Happy to rework it if the concern is the approach rather than the fix itself.

PR #6089, opened 2026-07-09
AI junk

M0nd0R:
## Summary

`SECRET_KEY_FALLBACKS` is documented and implemented as a list of old signing keys. If an app accidentally configures it as a scalar string or bytes value, the session interface currently calls `keys.extend(fallbacks)`, which expands the value into one-character or one-byte fallback keys before constructing the `itsdangerous` serializer.

This rejects scalar `str` and `bytes` fallback values before building the key list, while preserving the existing list-based key rotation behavior.

## Security impact

For an affected app using signed cookie sessions, a misconfigured value such as `SECRET_KEY_FALLBACKS = "old-secret"` could make Flask accept cookies signed with a single-character fallback key such as `"o"`. I verified the pre-fix behavior with a cookie signed using `"o"`; the session loaded successfully when `SECRET_KEY_FALLBACKS` was a string. After this change, the same request raises a configuration `TypeError` instead of accepting the forged session.

This can also happen through environment configuration if a value is supplied as a plain string instead of JSON/list syntax.

## Validation

- `python3 -m py_compile src/flask/sessions.py tests/test_basic.py`
- `/tmp/flask-verify-venv/bin/ruff check src/flask/sessions.py tests/test_basic.py`
- `/tmp/flask-verify-venv/bin/ruff format --check src/flask/sessions.py tests/test_basic.py`
- `/tmp/flask-verify-venv/bin/python -m pytest tests/test_basic.py::test_session_secret_key_fallbacks tests/test_basic.py::test_session_secret_key_fallbacks_rejects_string_or_bytes -q`
- local full suite before cleanup: `493 passed in 6.36s`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,8 @@ Unreleased
     it's disabled in config. Previously, only disabling worked. :issue:`5916`
 -   ``Flask.select_jinja_autoescape`` uses case-insensitive comparison instead
     of only lower case file extensions. :pr:`6012`
+-   Reject scalar string or bytes values for ``SECRET_KEY_FALLBACKS`` instead
+    of treating each character as a fallback signing key.
 
 
 Version 3.1.3
--- a/src/flask/sessions.py
+++ b/src/flask/sessions.py
@@ -307,6 +307,12 @@ class SecureCookieSessionInterface(SessionInterface):
         keys: list[str | bytes] = []
 
         if fallbacks := app.config["SECRET_KEY_FALLBACKS"]:
+            if isinstance(fallbacks, (str, bytes)):
+                raise TypeError(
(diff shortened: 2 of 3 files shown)


M0nd0R:
I checked the two failing CI jobs. They appear unrelated to this patch:

- `Development Versions` fails during collection because current Werkzeug dev emits deprecation warnings for existing test imports of `parse_cache_control_header` and `parse_set_header`.
- `Minimum Versions` fails during Flask import before tests run because existing `CertParamType(click.ParamType[t.Any])` is not compatible with the pinned minimum Click 8.1.3.

The jobs that exercise the supported Python/OS matrix, style, docs, and typing passed. I kept this PR scoped to the session fallback validation rather than mixing in dependency/test-suite maintenance changes.

PR #6095, opened 2026-07-14
Replace the use of private `monkeypatch` fixture API

### LINKED ISSUE #6071: Tests fail with pytest 9.1: _pytest.monkeypatch.notset removed
Flask's test suite imports the private notset sentinel from _pytest.monkeypatch in both tests/conftest.py and tests/test_cli.py. pytest 9.1 removed this attribute, moving it to _pytest.compat.NOTSET. This causes a collection-time crash — no tests can run at all.

### Reproducing
```
pip install -e . pytest==9.1.1
pytest tests/
tests/test_cli.py:14: in <module>
    from _pytest.monkeypatch import notset
E   ImportError: cannot import name 'notset' from '_pytest.monkeypatch'
```
Even after fixing test_cli.py, every test fails because tests/conftest.py accesses monkeypatch.notset as a module attribute in the _standard_os_environ session fixture:

`tests/conftest.py:19: AttributeError: module '_pytest.monkeypatch' has no attribute 'notset'`
### Expected Behavior
The test suite should work with pytest 9.1+.

### Environment:

Python version: 3.15.0b2
Flask version: 3.1.3

mgorny:
Replace the use of private `monkeypatch` API with the standard invocations to fix compatibility with pytest 9.1, as well as to make the tests more reliable in the future.

The previous code operated on `monkeypatch` fixture internals to trigger clearing environment variables expected by tests at the beginning of tests, and reverting any changes made by the tests.  However, this does not seem strictly valid, and it actually lead to tests making wrong assumptions.  For example, `test_disable_dotenv_from_env` relied on a previous test unsetting `FOO` in the environment, and failed if `FOO` was set while it was run alone.

The new logic uses public API only: it cleans up the standard set of environment variables for every test, and additionally cleans up variables specifically used by dotenv tests.  With this approach, it is entirely possible for a test to leave the environment "dirty"; however, that's fine since other tests need to clean up the environment anyway to account for user-set environment variables.

Fixes #6071

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,43 +2,26 @@ import os
 import sys
 
 import pytest
-from _pytest import monkeypatch
 
 from flask import Flask
 from flask.globals import request_ctx
 
 
-@pytest.fixture(scope="session", autouse=True)
-def _standard_os_environ():
-    """Set up ``os.environ`` at the start of the test session to have
-    standard values. Returns a list of operations that is used by
-    :func:`._reset_os_environ` after each test.
-    """
-    mp = monkeypatch.MonkeyPatch()
-    out = (
-        (os.environ, "FLASK_ENV_FILE", monkeypatch.notset),
-        (os.environ, "FLASK_APP", monkeypatch.notset),
-        (os.environ, "FLASK_DEBUG", monkeypatch.notset),
-        (os.environ, "FLASK_RUN_FROM_CLI", monkeypatch.notset),
-        (os.environ, "WERKZEUG_RUN_MAIN", monkeypatch.notset),
-    )
-
-    for _, key, value in out:
-        if value is monkeypatch.notset:
-            mp.delenv(key, False)
-        else:
-            mp.setenv(key, value)
-
-    yield out
-    mp.undo()
-
-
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -2,43 +2,26 @@
...
 from flask.globals import app_ctx as _app_ctx
 
 
-@pytest.fixture(scope="session", autouse=True)
-def _standard_os_environ():
-    """Set up ``os.environ`` at the start of the test session to have
-    standard values. Returns a list of operations that is used by
-    :func:`._reset_os_environ` after each test.
-    """
-    mp = monkeypatch.MonkeyPatch()
```
RonnyPfannschmidt, line 17: actualy keep this fixture and use 

with pytest.MonkeyPatch.context() as mp:
   ...
   yield ....

mgorny: Good idea!

mgorny: Sorry; actually, does that change anything? I mean, either way we need a function scope fixture, and my understanding is that it will undo the changes just the same.

mgorny:
Thank you!

PR #6096, opened 2026-07-14
Fix `.partition(":")` usage on potential IPv6 addresses

### LINKED ISSUE #6093: IPv6 addresses parsed incorrectly because of `.partition(":")`?
Hi, I checked the code and saw that two places use `.partition(":")` on addresses, and IPv6 addresses contain ":".

This breaks 1. `session_transaction` value setting, and 2. `Flask.run()` if `SERVER_NAME` is set. I've already prepared two tests that fail, you can use them to reproduce:

1. In `test_testing.py`
```
    client = app.test_client()

    @app.get("/")
    def index():

        sess["value"] = 42

```

2. Add one case to `test_run_from_config`
```
(None, None, "[::1]:8080", "::1", 8080),
```

Is this known and intended? I could send the patches for these two places.

lkk7:
Fix two usages of `.partition(":")` on potential IPv6 addresses.
Replace them with alternatives.

Fixes https://github.com/pallets/flask/issues/6093

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -12,6 +12,7 @@ from inspect import iscoroutinefunction
 from itertools import chain
 from types import TracebackType
 from urllib.parse import quote as _url_quote
+from urllib.parse import urlsplit
 
 import click
 from werkzeug.datastructures import Headers
@@ -721,7 +722,9 @@ class Flask(App):
         sn_host = sn_port = None
 
         if server_name:
-            sn_host, _, sn_port = server_name.partition(":")
+            server_url = urlsplit(f"//{server_name}")
+            sn_host = server_url.hostname
+            sn_port = server_url.port
 
         if not host:
             if sn_host:
@@ -731,8 +734,8 @@ class Flask(App):
 
         if port or port == 0:
             port = int(port)
-        elif sn_port:
-            port = int(sn_port)
+        elif sn_port is not None:
+            port = sn_port
         else:
             port = 5000
 
@@ -745,7 +748,7 @@ class Flask(App):
         from werkzeug.serving import run_simple
 
         try:
-            run_simple(t.cast(str, host), port, self, **options)
+            run_simple(host, port, self, **options)
         finally:
             # reset the first request information if the development server
             # reset normally.  This makes it possible to restart the server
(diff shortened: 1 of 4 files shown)


src/flask/app.py
```diff
@@ -645,8 +648,8 @@ def run(
 
         if port or port == 0:
             port = int(port)
-        elif sn_port:
-            port = int(sn_port)
+        elif sn_port is not None:
```
lkk7, line 737: `sn_port` changed to int/none, that's the reason for this change

src/flask/app.py
```diff
@@ -659,7 +662,7 @@ def run(
         from werkzeug.serving import run_simple
 
         try:
-            run_simple(t.cast(str, host), port, self, **options)
+            run_simple(host, port, self, **options)
```
lkk7, line 751: Mypy: Redundant cast to "str"

src/flask/testing.py
```diff
@@ -177,7 +177,7 @@ def session_transaction(
             app.session_interface.save_session(app, sess, resp)
 
         self._update_cookies_from_response(
-            ctx.request.host.partition(":")[0],
+            urlsplit(ctx.request.host_url).hostname or "localhost",
```
lkk7, line 180: That's the part I was unsure about. 
Hostname can be `None`, so I used the behavior from the werkzeug change you mentioned:

`hostname = url.hostname or "localhost"`

src/flask/app.py
```diff
@@ -635,7 +636,9 @@ def run(
         sn_host = sn_port = None
 
         if server_name:
-            sn_host, _, sn_port = server_name.partition(":")
+            server_url = urlsplit(f"//{server_name}")
```
lkk7, line 725: I added "//" to make urlsplit work in this instance. I assume that this won't break anything, but you definitely know more about that.

lkk7:
Not sure about the reason of "Development Versions" fail...

davidism:
Dev tests are not updated for Werkzeug main yet, they're expected to fail until next Werkzeug release.

Samielakkad (commented):
Retracted.

davidism:
All looks good, thanks. I'll target this at main since I don't think there will be another bug fix release, and this isn't anything critical.

PR #6110, opened 2026-07-30
AI junk

Tarunjit45:
Fixes #5200.

### Root Cause Hypothesis
The issue is caused by the lack of support for passing positional arguments to class-based views in Flask.

### Approach
To fix this issue, we need to modify the `dispatch_request` method in the `View` and `MethodView` classes to accept positional arguments. We can do this by changing the method signature to `def dispatch_request(self, *args, **kwargs)` and then passing these arguments to the view function. This will allow class-based views to accept positional arguments.

### Explanation
The `dispatch_request` method in the `View` and `MethodView` classes has been modified to accept positional arguments. The `as_view` method in the `View` class has also been updated to pass these arguments to the view function.

Automated pull request submitted by Autonomous SWE Agent.

### DIFF
--- a/src/flask/views.py
+++ b/src/flask/views.py
@@ -12,7 +12,6 @@ http_method_funcs = frozenset(
     ["get", "post", "head", "options", "delete", "put", "trace", "patch"]
 )
 
-
 class View:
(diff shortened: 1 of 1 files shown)


davidism:

PR #6115, opened 2026-08-01
AI junk

DijieDeng:
## Fix IPv6 address parsing in `session_transaction`

Fixes part of #6093.

### Problem

Two places in Flask use `.partition(":")` to parse host:port strings, but IPv6 addresses like `[::1]:8000` contain multiple colons, so `.partition(":")` splits at the wrong position:

1. **`session_transaction` in `src/flask/testing.py`** — `ctx.request.host.partition(":")[0]` returns `"["` instead of `"[::1]"` for IPv6 hosts. This breaks cookie domain extraction during session transactions with IPv6 base URLs.
2. **`Flask.run()` in `src/flask/app.py`** — `server_name.partition(":")` incorrectly parses `SERVER_NAME` when it contains an IPv6 address (e.g., `[::1]:8080`).

### Fix applied in this PR (testing.py)

Replaced `ctx.request.host.partition(":")[0]` with `urlsplit(ctx.request.host_url).hostname or "localhost"`. The `host_url` property includes the scheme (e.g., `http://[::1]:8000`), so `urlsplit` can correctly parse the hostname from IPv6 addresses. The `urlsplit` import already exists in `testing.py`.

### Fix still needed (app.py)

The `run()` method in `src/flask/app.py` also needs the same fix. The change would be:

```python
# In imports, add:
from urllib.parse import urlsplit

# In run() method, replace:
sn_host, _, sn_port = server_name.partition(":")
# With:
sn_url = urlsplit(f"//{server_name}")
sn_host = sn_url.hostname
sn_port = sn_url.port

# And change:
port = int(sn_port)
# To:
port = sn_port  # sn_port is already int | None from urlsplit
```

This follows the same approach used in the [Werkzeug fix](https://github.com/pallets/werkzeug) for the same issue.

### Test added

Added `test_session_transaction_ipv6` to `tests/test_testing.py` which verifies that session transactions work correctly with IPv6 base URLs:

```python
    client = app.test_client()

    @app.get("/")
    def index():

        sess["value"] = 42

```

This test fails without the fix (raises an error or returns incorrect results) and passes with the fix.

### DIFF
--- a/src/flask/testing.py
+++ b/src/flask/testing.py
@@ -109,8 +109,8 @@ def _get_werkzeug_version() -> str:
 class FlaskClient(Client):
     """Works like a regular Werkzeug test client, with additional behavior for
     Flask. Can defer the cleanup of the request context until the end of a
-    ``with`` block. For general information about how to use this class refer to
-    :class:`werkzeug.test.Client`.
+    ``with`` block. For general information about how to use this class refer
+    to :class:`werkzeug.test.Client`.
 
     .. versionchanged:: 0.12
        `app.test_client()` includes preset default environment, which can be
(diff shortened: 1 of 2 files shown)


ThiefMaster:
This appears LLM-generated.

PR #6116, opened 2026-08-01
Add allow_nan attribute to DefaultJSONProvider

### LINKED ISSUE #6114: jsonify emits NaN and Infinity, which are not valid JSON
`jsonify` serialises `float("nan")` and `float("inf")` as the bare literals `NaN` and `Infinity`. Those are not valid JSON (RFC 8259 allows only decimal numbers), so the response is served as `application/json` but cannot be parsed by a strict parser, including the browser's `JSON.parse`.

This inherits from `json.dumps`, whose `allow_nan` defaults to `True`. `DefaultJSONProvider` sets `default`, `ensure_ascii` and `sort_keys`, but never touches `allow_nan`, so the permissive default applies.

The path to it is ordinary: any average, ratio or division that can meet an empty input produces `nan` without anything looking wrong on the server.

Replication:

```python
from flask import Flask, jsonify

app = Flask(__name__)

@app.route("/stats")
def stats():
    values = []
    mean = sum(values) / len(values) if values else float("nan")
    return jsonify({"count": len(values), "mean": mean})

with app.test_client() as c:
    r = c.get("/stats")
    print(r.headers["Content-Type"])      # application/json
    print(r.get_data(as_text=True))       # {"count":0,"mean":NaN}
```

The response advertises `application/json` with a 200, but the body is not JSON:

```python
import json
json.loads('{"count":0,"mean":NaN}')      # accepted, Python is permissive by default
json.loads('{"count":0,"mean":NaN}',
           parse_constant=lambda x: (_ for _ in ()).throw(ValueError(x)))
# ValueError: NaN     <- what a strict parser, and JSON.parse in a browser, does
```

So the failure surfaces in the client rather than the server, which makes it awkward to trace back.

Expected: either the same behaviour with it documented, or serialisation raises so the problem shows up server-side, where `ValueError: Out of range float values are not JSON compliant: nan` names the cause immediately.

There is currently no class attribute for this, although the plumbing already works, since `dumps` forwards `**kwargs` to `json.dumps`:

```python
app.json.dumps({"a": float("nan")}, allow_nan=False)
# ValueError: Out of range float values are not JSON compliant: nan
```

`DefaultJSONProvider` exposes `sort_keys`, `ensure_ascii` and `compact` as class attributes, so an `allow_nan` attribute alongside them would fit the existing pattern and let applications choose, without changing the default for anyone. I could not find `NaN` mentioned in the `DefaultJSONProvider` or `flask.json.dumps` docstrings, so at minimum a documentation note seems worthwhile.

Happy to send a PR for whichever you prefer, an `allow_nan` provider attribute or just the docs.

Environment:

- Python 3.14
- Flask 3.1.3
- Werkzeug 3.1.8

Omacrow:
DefaultJSONProvider now exposes an allow_nan attribute (default True,
matching the existing json.dumps default), so applications can opt
into strict RFC 8259 output with app.json.allow_nan = False.

fixes #6114

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ Unreleased
     it's disabled in config. Previously, only disabling worked. :issue:`5916`
 -   ``Flask.select_jinja_autoescape`` uses case-insensitive comparison instead
     of only lower case file extensions. :pr:`6012`
+-   Add ``allow_nan`` attribute to ``DefaultJSONProvider`` to control whether
+    ``NaN`` and ``Infinity`` values are allowed when serializing to JSON.
+    Defaults to ``True`` to preserve existing behavior. :issue:`6114`
 
 
 Version 3.1.3
--- a/src/flask/json/provider.py
+++ b/src/flask/json/provider.py
@@ -160,6 +160,15 @@ class DefaultJSONProvider(JSONProvider):
     or ``None`` in debug mode, it will use a non-compact representation.
     """
 
+    allow_nan = True
+    """Allow ``NaN``, ``Infinity``, and ``-Infinity`` to be serialized as
+    the bare literals ``NaN``, ``Infinity``, and ``-Infinity``. These are
+    accepted by Python's :mod:`json` but are not valid JSON per
+    :rfc:`8259`, so strict parsers such as a browser's ``JSON.parse``
+    will reject them. Set this to ``False`` to raise a ``ValueError``
+    instead when such a value is serialized.
+    """
+
     mimetype = "application/json"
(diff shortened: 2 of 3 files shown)


Omacrow:
The failing "Development Versions" check looks unrelated to this change 
it's failing on a DeprecationWarning from Werkzeug's `main` branch
(`parse_cache_control_header` / `parse_set_header`) inside test_blueprints.py
and test_views.py, neither of which this PR touches.

davidism:
Duplicate of https://github.com/pallets/flask/pull/2832

PR #6118, opened 2026-08-02
AI junk

fyrsta7:
## Summary

This adds a direct fast path for `Request.blueprints` when the endpoint name belongs to a single blueprint and does not contain a dot.

## Why this can improve performance

The cached `_split_blueprint_path` helper is still needed for nested blueprint names. For the common single-name case, returning `[name]` directly avoids the cached recursive helper call while preserving the existing nested behavior.

## Validation

- `git diff --check`
- `python3 -m compileall -q src/flask/wrappers.py`
- Behavior check for both single and nested blueprint endpoint names

## Benchmark

Current machine, CPython microbenchmark, 9 samples, median:

| Benchmark | Base | This change | Delta |
|---|---:|---:|---:|
| `Request.blueprints` for single blueprint names | 199,323.717 ns/op | 197,332.378 ns/op | 1.00% faster |

### DIFF
--- a/src/flask/wrappers.py
+++ b/src/flask/wrappers.py
@@ -192,6 +192,9 @@ class Request(RequestBase):
         if name is None:
             return []
 
+        if "." not in name:
+            return [name]
+
         return _split_blueprint_path(name)
 
     def _load_form_data(self) -> None:


davidism:

PR #6124, opened 2026-08-08
AI junk

### LINKED ISSUE #6123: `stream_with_context`: abandoned generator leaves the app context current on the worker thread; later requests skip `teardown_appcontext`
`stream_with_context` pushes the request's contexts onto whichever thread iterates the response body, and pops them only if the generator completes or is closed on that same thread. When a client disconnects mid-stream, the server abandons the body iterator and GC finalizes it later — often on a different thread. That cross-context pop fails (`LookupError`/`ValueError` from `ContextVar.reset`, swallowed as "Exception ignored in ..."), and the app context stays current on the worker thread.

Because `RequestContext.push` reuses an ambient app context of the same app, every later request on that pooled thread runs under the stuck context and `teardown_appcontext` never fires for it. Anything keyed to the app context then leaks once per thread for the life of the process — with Flask-SQLAlchemy that is one connection held idle-in-transaction per poisoned thread, with nothing in the logs.

Related: #5774 fixed the same cross-context token failure for async views; #5804 covers teardown ordering in the 3.1.2 rewrite. Neither covers the abandonment case. The docs mention that headers/session need care with `stream_with_context`, but not that a disconnect can leave the context stuck on the thread — a warning there would be helpful even if the behavior itself is hard to change.

## Reproduction

One long-lived "worker" thread serves two requests; the abandoned body iterator is GC-finalized on the main thread in between (in real deployments the reference is dropped by server machinery outside the worker — servers may never call `close()` on the iterator after a disconnect):

```python
import gc
import queue
import threading

from flask import Flask, Response, stream_with_context
from flask.globals import _cv_app

app = Flask(__name__)
teardowns = []

@app.teardown_appcontext
def note_teardown(exc):
    teardowns.append(threading.current_thread().name)

@app.route("/stream")
def stream():
    def gen():
        for i in range(100):
            yield f"chunk{i}\n"
    return Response(stream_with_context(gen()))

@app.route("/plain")
def plain():
    return "ok"

abandoned = queue.Queue()   # holds the dropped iterator, like server machinery
step2 = threading.Event()

def worker():
    # --- request A: streaming response, client disconnects after one chunk ---
    with app.test_request_context("/stream"):
        rv = app.full_dispatch_request()
    body = iter(rv.response)
    next(body)   # first chunk: stream_with_context pushes app+request ctx HERE
    abandoned.put((body, rv))   # server-side reference outlives the request
    del body, rv
    print("A: ambient ctx after abandonment:", _cv_app.get(None))

    step2.wait()  # main thread GC-finalizes the generator meanwhile

    # --- request B: any later request on this pooled worker thread ---
    print("B: ambient ctx at request start:", _cv_app.get(None))
    teardowns.clear()
    with app.test_request_context("/plain"):    # reuses the ambient ctx of the
        app.full_dispatch_request()             # same app -> no fresh app ctx
    print("B: teardown_appcontext ran for B:", bool(teardowns))
    print("B: ambient ctx after B finished:", _cv_app.get(None))

t = threading.Thread(target=worker, name="wsgi-worker-0")
t.start()

abandoned.get()          # take the reference, then drop it on the MAIN thread
gc.collect()             # generator finalized here -> cross-context pop fails
step2.set()
t.join()
```

Output (Flask 3.1.3, Python 3.13):

```
Exception ignored in: <generator object stream_with_context.<locals>.generator at 0x...>
  File ".../flask/helpers.py", line 130, in generator
    with app_ctx, req_ctx:
  File ".../flask/ctx.py", line 284, in __exit__
    self.pop(exc_value)
  File ".../flask/ctx.py", line 264, in pop
    ctx = _cv_app.get()
LookupError: <ContextVar name='flask.app_ctx' at 0x...>
A: ambient ctx after abandonment: <flask.ctx.AppContext object at 0x10b49c830>
B: ambient ctx at request start: <flask.ctx.AppContext object at 0x10b49c830>
B: teardown_appcontext ran for B: False
B: ambient ctx after B finished: <flask.ctx.AppContext object at 0x10b49c830>
```

## Expected behavior

Request B runs under its own app context and `teardown_appcontext` fires when it finishes. Instead it silently runs under request A's leaked context, teardown never runs, and the thread stays poisoned indefinitely.

Environment:

- Python version: 3.13 (also reproduced on 3.11)
- Flask version: 3.1.3. The `main` branch (3.2-dev, post-#5639 merged context) looks affected as well: `stream_with_context` there enters the single context via `with ctx:` around the priming `yield None`, so the context is pushed at creation and only exited when the generator completes or is closed.

rowkav09:
## What changed

`stream_with_context` now pushes the captured context only while advancing or closing the wrapped iterator. This prevents an abandoned stream finalized on another thread from leaving the worker thread with a stale application context.

Added a regression test covering a cross-thread stream finalization and a subsequent request on the same worker.

## Testing

- `uv run --project C:/oss-flask pytest C:/oss-flask/tests -q` (492 passed)
- Ruff check and format check passed

Fixes #6123

### DIFF
--- a/src/flask/helpers.py
+++ b/src/flask/helpers.py
@@ -130,19 +130,27 @@ def stream_with_context(
                 " context is active, such as in a view function."
             )
 
-        with ctx:
-            yield None  # type: ignore[misc]
+        # Push the context only while advancing the wrapped iterator. Keeping
+        # it pushed across a yield means that generator finalization on a
+        # different thread cannot reset the context variable.
+        yield None  # type: ignore[misc]
 
-            try:
-                yield from gen
-            finally:
-                # Clean up in case the user wrapped a WSGI iterator.
-                if hasattr(gen, "close"):
+        try:
+            while True:
+                with ctx:
+                    try:
+                        item = next(gen)
+                    except StopIteration:
+                        return
+                yield item
+        finally:
+            # Clean up in case the user wrapped a WSGI iterator.
+            if hasattr(gen, "close"):
+                with ctx:
                     gen.close()
 
-    # Execute the generator to the sentinel value. This captures the current
-    # context and pushes it to preserve it. Further iteration will yield from
-    # the original iterator.
+    # Execute the generator to the sentinel value to capture the current
+    # context. Further iteration will advance the original iterator under it.
     wrapped_g = generator()
     next(wrapped_g)
     return wrapped_g
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -1,5 +1,8 @@
+import gc
 import io
 import os
+import queue
+import threading
 
 import pytest
 import werkzeug.exceptions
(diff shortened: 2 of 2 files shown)


davidism:
Besides being AI generated, which is against our policy, this introduces extreme inefficiency due to pushing and popping the context on every loop.

PR #6127, opened 2026-08-09
AI junk

### LINKED ISSUE #6065: Add a query() route shortcut and MethodView support for HTTP QUERY (RFC 10008)

aryansk:
fixes #6065

Add the `query()` route shortcut and recognize `query` methods when deriving `MethodView.methods` and dispatching HTTP `QUERY` requests. The existing `route(..., methods=["QUERY"])` behavior remains unchanged.

The change includes focused view tests, code documentation for the new shortcut and dispatch behavior, and an entry in `CHANGES.rst`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,8 @@ Unreleased
     it's disabled in config. Previously, only disabling worked. :issue:`5916`
 -   ``Flask.select_jinja_autoescape`` uses case-insensitive comparison instead
     of only lower case file extensions. :pr:`6012`
+-   Add ``query`` route shortcut and ``MethodView`` dispatch support for the
+    HTTP ``QUERY`` method. :issue:`6065`
 
 
 Version 3.1.3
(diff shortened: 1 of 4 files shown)


davidism:

PR #6131, opened 2026-08-10
AI junk

### LINKED ISSUE #6093: IPv6 addresses parsed incorrectly because of `.partition(":")`?

ameerhmz:
# What does this PR do?

Fixes #6093 where IPv6 addresses (e.g. `[::1]:8080` or `http://[::1]:8000/`) were parsed incorrectly using `.partition(":")`, causing string splits on colons inside IPv6 host brackets.

### Root Cause
1. In `src/flask/app.py`: `server_name.partition(":")` split `"[::1]:8080"` into `sn_host = "["`, `sn_port = ":1]:8080"`.
2. In `src/flask/testing.py`: `ctx.request.host.partition(":")[0]` returned `"["` instead of normalized IPv6 host `"::1"`.

### Solution
1. Updated `src/flask/app.py` `SERVER_NAME` parsing to check for `"]:"` or `rpartition(":")` to extract IPv6 host and port correctly.
2. Updated `src/flask/testing.py` `session_transaction` cookie host parsing to handle IPv6 bracketed hosts.
3. Added test case in `tests/test_basic.py` `test_run_from_config` for `[::1]:8080`.
4. Added `test_session_transaction_ipv6` in `tests/test_testing.py`.

- [x] This PR fixes #6093.
- [x] Included unit tests.

### DIFF
--- a/src/flask/app.py
+++ b/src/flask/app.py
@@ -721,7 +721,15 @@ class Flask(App):
         sn_host = sn_port = None
 
         if server_name:
-            sn_host, _, sn_port = server_name.partition(":")
+            if "]:" in server_name:
+                sn_host, _, sn_port = server_name.rpartition("]:")
+                sn_host += "]"
+            elif server_name.startswith("[") and server_name.endswith("]"):
+                sn_host, sn_port = server_name, None
+            elif ":" in server_name:
+                sn_host, _, sn_port = server_name.rpartition(":")
+            else:
+                sn_host, sn_port = server_name, None
 
         if not host:
             if sn_host:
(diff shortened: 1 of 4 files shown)


davidism:

PR #6138, opened 2026-08-24
[invalid] Add support for custom JSON serialization with `__json__`

sebbASF:
Allow classes to provide a JSON serialiser which is higher priority than the html method.

This allows classes to support both JSON and HTML with appropriate output.

### DIFF
--- a/src/flask/json/provider.py
+++ b/src/flask/json/provider.py
@@ -115,6 +115,9 @@ def _default(o: t.Any) -> t.Any:
     if dataclasses and dataclasses.is_dataclass(o):
         return dataclasses.asdict(o)  # type: ignore[arg-type]
 
+    if hasattr(o, "__json__"):
+        return str(o.__json__())
+
     if hasattr(o, "__html__"):
         return str(o.__html__())
 


davidism:

tisonkun:
@davidism This should be a false positive. Would you do a human review?

davidism:
This was unasked for, has little explanation and doesn't follow the template or contributing guide, and the account is private. Nowadays, that all points to a user asking AI tools to output stuff that looks like a contribution. It looks like you're invloved with ASF, and the username suggests association with ASF too, so I guess I'll reverse the AI marking. However, it also looks like you're heavily using AI tools, so please be aware of our AI policy as well.

tisonkun:
@davidism Thanks for your reply!

I think then if @sebbASF wants to continue this PR, he should start with an issue to describe why it is needed?

davidism:
We have an extensive contributing guide.

Regardless, I don't want this feature. I've responded as such for other proposed extensions to our JSON types, a search of previous issues would reveal. Use a serialization library like Marshmallow, cattrs, Pydantic, etc to get data prepared for JSON serialization.

davidism:
I suggest @sebbASF make their account activity public and read contributing guides to projects before contributing. Not doing either of those are huge red flags.

PR #6142, opened 2026-08-28
AI junk

YeshwanthRajSelvaraj:
- Handle `errno.ENOTDIR` in `Config.from_file` when `silent=True`, matching the behavior of `Config.from_pyfile`.
- Expand `Config.from_mapping` type annotations to include `t.Iterable[tuple[str, t.Any]]` (such as `dict.items()` or sequence of tuples), which is supported by `dict.update`.
- Add test in `tests/test_config.py`.


davidism:

PR #6144, opened 2026-08-29
AI junk

### LINKED ISSUE #6143: `add_url_rule` leaves a route registered even after an endpoint conflict
## Description
‘add_url_rule’ correctly raises an error when an endpoint already exists and you try to register a different view function with the same endpoint.

However, after you catch the error and continue running the code, the URL you tried to add (/b, for example) still exists. When a user requests /b, it is routed to the view function that was already registered for that endpoint, rather than the new view function whose registration failed.

## Example

```python
from flask import Flask

app = Flask(__name__)

def first():
    return "first"

def second():
    return "second"

app.add_url_rule("/a", endpoint="same", view_func=first)

try:
    app.add_url_rule("/b", endpoint="same", view_func=second)
except AssertionError as e:
    print(e)

client = app.test_client()

print(client.get("/a").status_code, client.get("/a").data)
print(client.get("/b").status_code, client.get("/b").data)
```

### Expected Result

```text
View function mapping is overwriting an existing endpoint function: same
200 b'first'
404 ...
```

### Actual Result

```text
View function mapping is overwriting an existing endpoint function: same
200 b'first'
200 b'first'
```

I think Flask correctly rejects the view-function overwrite, but the failed registration has already added the URL rule.

Environment:

- Python version: 3.11.2
- Flask version: 3.1.3

amtbsl:
Fixes #6143.

When URL rule registration detects an endpoint conflict, roll back the newly inserted rule so the application state remains consistent. This also centralizes removal logic and adds a regression test.

Tests: `pytest tests/test_basic.py` (135 passed).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,8 @@ Unreleased
     of only lower case file extensions. :pr:`6012`
 -   Fix parsing IPv6 with port in ``run`` and the test client. :pr:`6096`
 -   Add ``app.query`` route decorator for the HTTP QUERY method.
+-   Do not leave a URL rule registered when ``add_url_rule`` rejects a
+    conflicting endpoint function. :issue:`6143`
 
 
 Version 3.1.3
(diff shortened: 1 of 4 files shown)


davidism:

